PR #28, opened 2020-03-30
reduce file sizes of images with FileOptimizer

wehttvM:
Reduce file size without sacrificing any visual loss.


willmcgugan:
Thanks.

Issue #90, opened 2020-05-24
Deadlock when exiting Progress context


eulertour:
I can consistently reproduce this by letting the following script run. Eventually the progress bars will hang when they're nearly full, like so:

![deadlock](https://user-images.githubusercontent.com/43117506/82743832-85102500-9d25-11ea-8a92-8b8472f12dde.png)

```python
from rich.progress import track
from rich.progress import Progress
import random
import time

while True:
    with Progress() as progress:
        task1 = progress.add_task(f"[red]Downloading...", total=10)
        task2 = progress.add_task("[green]Processing...", total=10)
        task3 = progress.add_task("[cyan]Cooking...", total=10)

        while not progress.finished:
            progress.update(task1, advance=random.random())
            progress.update(task2, advance=random.random())
            progress.update(task3, advance=random.random())
            time.sleep(0.1)
```

I believe the problematic sequence of events is as follows:
1. The main thread exits from a `Progress` context and acquires `Progress._lock` in `Progress.__exit__()`
https://github.com/willmcgugan/rich/blob/55988c6e721cc03ab4f6dc6aeca3089e7c5c13ae/rich/progress.py#L447-L451

2. The refresh thread finishes waiting on `_RefreshThread.done` and calls `Progress.refresh()`, at which point it starts waiting for `Progress._lock`
https://github.com/willmcgugan/rich/blob/55988c6e721cc03ab4f6dc6aeca3089e7c5c13ae/rich/progress.py#L346-L348
https://github.com/willmcgugan/rich/blob/55988c6e721cc03ab4f6dc6aeca3089e7c5c13ae/rich/progress.py#L574-L582

3. The main thread calls `Progress._refresh_thread.stop()` in `Progress.stop()`, at which point it waits for the refresh thread to finish with `_RefreshThread.join()`
https://github.com/willmcgugan/rich/blob/55988c6e721cc03ab4f6dc6aeca3089e7c5c13ae/rich/progress.py#L422-L436
https://github.com/willmcgugan/rich/blob/55988c6e721cc03ab4f6dc6aeca3089e7c5c13ae/rich/progress.py#L342-L344

At this point the main thread is holding `Progress._lock` and waiting on the refresh thread to finish while the refresh thread is waiting to acquire `Progress._lock` so it can finish.

I don't think it's safe to call `_RefreshThread.join()` while holding `Progress._lock`, because there's always a chance that the refresh thread started waiting on `Progress._lock` the nanosecond after it was acquired.

[...]

PR #91, opened 2020-05-24
Attempt at solving progress bar deadlock

eulertour:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This is my attempt at solving #90 by controlling access to `Progress._tasks` and `Progress._refresh_count` with a separate lock. I haven't been able to reproduce a deadlock after applying this change, but it's hard to tell if there are any errors without more knowledge of which data races the original lock is meant to prevent (I noticed for instance that the `Progress.tasks` property didn't acquire the original lock, which leads me to believe read races on that data are permissible).

### DIFF
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -374,6 +374,8 @@ class Progress:
             TextColumn("[progress.percentage]{task.percentage:>3.0f}%"),
             TimeRemainingColumn(),
         )
+        self._lock = RLock()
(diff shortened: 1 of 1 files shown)


willmcgugan:
Thanks. I'll get back to you after I've looked in to the deadlock in more detail.

PR #107, opened 2020-06-08
enable emojis for rich logging handler

smacke:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [X] Tests
- [X] Other

## Checklist

I didn't do this since it's a small change.

:)

## Description

Run logged strings through _emoji_replace so we can use the emoji shorthand in conjunction with rich's logging handler :). Add basic test.

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -589,7 +589,7 @@ class Console:
             rich_text.overflow = overflow
         else:
             rich_text = Text(
-                _emoji_replace(text) if emoji_enabled else text,
+                self.emoji_replace(text, emoji_enabled),
                 justify=justify,
                 overflow=overflow,
                 style=style,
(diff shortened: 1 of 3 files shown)


willmcgugan:
Thanks for the PR, but I'm afraid I'm not keen on enabling emoji code in logging. The emoji isn't going to be replaced in all handlers, and the function to do the replace is fast but not free in terms of cpu.

If you want emoji in your logs my suggestion would be to either copy RichHandler to your own project and customize it, or define a logging Formatter that does the emoji replace. The later is probably better because you could get emoji an all handlers.

smacke:
Ah that makes sense; I should have looked for a way to do this that didn't involve modifying code first. Thanks for the response and for your work on this great library!

PR #130, opened 2020-06-29
Add markup support to log handler

furunkel:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [ ] I've run the latest [black](https://github.com/ambv/black) with default args on new code.

## Description

### DIFF
--- a/rich/logging.py
+++ b/rich/logging.py
@@ -62,7 +62,12 @@ class RichHandler(Handler):
 
         level = Text()
         level.append(record.levelname, log_style)
-        message_text = Text(message)
+
+        if getattr(record, "markup", False):
+            message_text = Text.from_markup(message)
+        else:
+            message_text = Text(message)
+
         if self.highlighter:
             message_text = self.highlighter(message_text)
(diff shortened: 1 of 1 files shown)


willmcgugan:
Not convinced this is a good idea. Granted is off by default, but as convenient as console markup is, you risk hiding messages containing square brackets. Often log messages come from external libraries that will be out of your control.

Would the `highlighter` option not serve your needs? You can highlight something based on a regex.

furunkel:
Using a `highlighter` would be quite painful. Good point though. It should be possible to use [setLogRecordFactory](https://docs.python.org/3/library/logging.html#logging.setLogRecordFactory) to install a custom factory function; this way the `markup` parameter would move to the actual log calls. Would that do?

willmcgugan:
That sounds like a good idea.

furunkel:
Passing it inside the `extra` dict does the trick. PR updated.

willmcgugan:
Good solution 👍

PR #160, opened 2020-07-15
add test for rstrip_end() in Text class

hedythedev:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

been looking at rich/text.py and found that tests for rstrip_end was missing, not sure why the coverage was 100% for text.py before then, now obviously is still 100% for the file. 

Changed file: tests/test_text.py

Also, 
1. I tried running `black . --check` but found that several other files needs reformatting, haven't pushed the reformatting changes yet, need to know your thoughts on that first:

image

2. Looks like you forgot to add the return type for rstrip_end() ?
https://github.com/willmcgugan/rich/blob/e9e72000c50f56654c89e119ff882e322e51ecf3/rich/text.py#L415
https://github.com/willmcgugan/rich/blob/e9e72000c50f56654c89e119ff882e322e51ecf3/rich/text.py#L411


hedythedev:
by the way just a suggestion, if you would like all the files to be black formatted, you should integrate `black . --check` in the github actions

willmcgugan:
Thanks @hedythedev 

I have black setup with vscode, but for some reason it leaves empty lines at the end of the file. Weird. Putting black in GH actions is a good idea. I'll sort the return type of rstrip_end.

Issue #72, opened 2020-05-14
Console(log_path=False, log_time=False) doesn't work with logging module


keelung-yang:
![image](https://user-images.githubusercontent.com/20385450/81889337-ddcd1880-95d5-11ea-8b27-aff3a3c0ebed.png)

```python3
import sys
import logging

from rich.console import Console
from rich.logging import RichHandler

log_fmt = '%(asctime)s.%(msecs)03d %(levelname)s:\t%(message)s'
date_fmt='%H:%M:%S'
level = logging.INFO

console = Console(
    tab_size=4,
    #record=False,
    log_path=False,
    log_time=False,
    #log_time_format=date_fmt,
    #file=sys.stderr,
)

handle = RichHandler(
    level=level,
    console=console,
)

logging.basicConfig(
    level=level,
    format=log_fmt,
    datefmt=date_fmt,
    handlers=[handle],    # comment this line to disable rich
)

logging.info("Hello, World!")
```
And BTW, `msecs` is not supported in rich now? Thanks!

willmcgugan:
Those parameters affect the `Console.log` method, the logging handler works differently, it takes the date format from the logging formatter and there is currently no option to not show the path.

If you want to configure logging more than what is offered, I'd suggest copying the [handler code](https://github.com/willmcgugan/rich/blob/master/rich/logging.py). It's a very simple class and you can modify it to your requirements.

PR #182, opened 2020-07-26
Update RichHandler to include all flags of LogRender

nathanrpage97:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

I saw a reference to this issue in #72. 

Update the Console and RichHandler class to include show_time and show_level.

### DIFF
--- a/rich/logging.py
+++ b/rich/logging.py
@@ -58,7 +58,9 @@ class RichHandler(Handler):
         super().__init__(level=level)
         self.console = console or get_console()
         self.highlighter = highlighter or self.HIGHLIGHTER_CLASS()
-        self._log_render = LogRender(show_time=show_time, show_level=show_level, show_path=show_path)
+        self._log_render = LogRender(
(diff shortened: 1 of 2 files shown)


rich/logging.py
```diff
@@ -56,7 +58,9 @@ def __init__(
         super().__init__(level=level)
         self.console = console or get_console()
         self.highlighter = highlighter or self.HIGHLIGHTER_CLASS()
-        self._log_render = LogRender(show_level=True, show_path=show_path)
+        self._log_render = LogRender(
```
willmcgugan, line 61: The `LogRender` class is also use in Console when you call `console.log` -- which doesn't have a `level` argument.

nathanrpage97: Ok, I've removed the log level from the creation of a console.

willmcgugan:
Thanks

ewels:
Nice, thanks both! I was about to come and request this feature 😅

PR #187, opened 2020-07-29
Replace all imports from rich with relative imports.

clintonc:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [X] Other

## Checklist

## Description

Replace all absolute imports within the library with relative ones.  The name of the library is itself an implementation detail, and the circumstances of its installation is the business of the system's owner.

### DIFF
--- a/rich/__main__.py
+++ b/rich/__main__.py
@@ -1,16 +1,16 @@
 import io
 from time import process_time
 
-from rich.console import Console, ConsoleOptions, RenderGroup, RenderResult
-from rich.markdown import Markdown
-from rich.measure import Measurement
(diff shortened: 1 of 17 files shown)


willmcgugan:
Reliable builds wouldn’t work if you install things under a different name. Besides, the core lib is all relative. The absolute imports are all in the main sections.

If you need this, and a test that increases coverage and I’ll consider merging.

clintonc:
There is an instance of non-relative imports outside of __main__ sections, in logging.py.  Just those could be fixed if you prefer.

Issue #37, opened 2020-04-26
Improve test coverage


willmcgugan:
Test coverage is currently siting at 89%. I'd love to get that up to high nineties before a 1.0.0 release.

If you can, contribute a test. Even if it only adds a single line to coverage, it would be appreciated!

edraobdu:
Hi Will, I just forked and ran the tests with coverage, and it shows a 91%, I know the higher the better, but isn't that an acceptable coverage rate to (at least) release the 1.0.0 version?

_Note: I used ```coverage run -m --source=. pytest```, see attached picture_

In the mean time, I realized some required packages were not listed in the requirement.txt:
**colorama==0.4.3
commonmark==0.9.1
typing-extensions==3.7.4.2**

![Screenshot from 2020-04-26 14-15-44](https://user-images.githubusercontent.com/31775663/80317315-bcdc8900-87c8-11ea-9805-ff7531c1a667.png)

willmcgugan:
Well its not bad... But I would like to be closer to 100%.

You can run the tests with `make test` which will also show the coverage.

I'll update the requirements. They're only needed for development, pyproject.toml stores the requirements for release.

willmcgugan:
Coverage is currently 94%

willmcgugan:
Test coverage is currently 97% which is not bad at all. I'll integrate coveralls at some point so I can brag about it on the readme.

willmcgugan:
Coverage is at 98%. Missing coverage is mostly in rich.progress, which is a little challenging to test.

Still plenty of easy pickings if anyone wants to contribute.

willmcgugan:
Test coverage is now at 99% which is pretty good!

Thanks to @pim-hoeven for picking up the slack there.

There are still gaps in the progress bar, some of which will be tricky due to the dependency on threads. And plenty of one liners to pick up if anyone is interested.

Edward-Knight:
We now have 99.57% coverage, which isn't _truely_ 100% coverage, but is now close enough to be rounded to 100%:

[![codecov](https://codecov.io/gh/willmcgugan/rich/branch/master/graph/badge.svg)](https://codecov.io/gh/willmcgugan/rich)

willmcgugan:
Closing this. Coverage in 99th percentile. 🍻 Will try to keep it there.

PR #228, opened 2020-08-10
Tests on traceback to 100% coverage

pim-hoeven:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

- [x] I've run the latest [black](https://github.com/ambv/black) with default args on new code. **But** there was a file (`tests/_card_render.py`) about which black complained, but this one isn't included in the changes.

## Description

Another small contribution to #37 . I am not so sure about the last two tests, because I couldn't exactly pinpoint what the traceback is supposed to do in those cases. 

Also: I added a `no cover` statement. I noticed that when I add a print statement before the `break`, both the `break` and the print are reached according to coverage (which makes sense). When I remove the print, the `break` isn't reached anymore. Not sure what happens there, might be a bug in ~~pytest~~ _coveragepy_?

Happy to make some changes / additions to this PR. Let me know if you need or want it. Thanks!

**Edit:** formatting and small change in wording

### DIFF
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -208,7 +208,9 @@ class Traceback:
                 exc_value = cause
                 traceback = cause.__traceback__
                 continue
-            break
+            # No cover, code is reached but coverage doesn't recognize it.
+            break  # pragma: no cover
+
         trace = Trace(stacks=stacks)
         return trace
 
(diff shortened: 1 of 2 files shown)


pim-hoeven:
Someone already filed an [issue](https://github.com/nedbat/coveragepy/issues/772) related to `break` not being recognized in coverage, so I guess there really wasn't an alternative to adding a `no cover`.

willmcgugan:
This is great, thanks @pim-hoeven ! I've noticed that issue with break myself.

PR #260, opened 2020-08-31
Add the ability to force terminal with environment variable

nathanrpage97:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Add the ability to set Console force_terminal from `FORCE_TERMINAL` environment variable. Useful for places such as CI tools.

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -352,7 +352,9 @@ class Console:
         )
 
         self._color_system: Optional[ColorSystem]
-        self._force_terminal = force_terminal if self._environ.get("FORCE_TERMINAL") is None else True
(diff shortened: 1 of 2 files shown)


willmcgugan:
I don't think this a good idea. I wouldn't want to presume how the application developer would like to configure their environment. It would also effect all Console instances, which might not be desirable.

If you do need it, it's a straightforward thing to implement as required.

nathanrpage97:
Makes sense. I was thinking this would standardize the behavior with `TERM` environment variable, but this does have much more of an effect on output.

PR #276, opened 2020-09-11
docs: Fix simple typo, garanteed -> guaranteed

timgates42:
There is a small typo in rich/_ratio.py.

Should read `guaranteed` rather than `garanteed`.


willmcgugan:
Thanks

PR #288, opened 2020-09-17
Remove align argument from table Column example

paw-lu:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

The example of using `rich.table.Column` in the docs seemed to use an older version of the api with an `align` argument that was no longer supported. I have removed the argument form the example and replaced it with `width` and `style` arguments.

Writing this example made me realize I actually do not know what the mandatory `index` argument does... 😅.

---

Been using your library for the past month on a personal project and it's been a joy to use. Thanks!

### DIFF
--- a/docs/source/tables.rst
+++ b/docs/source/tables.rst
@@ -59,13 +59,13 @@ You may also add columns by specifying them in the positional arguments of the :
 
     table = Table("Released", "Title", "Box Office", title="Star Wars Movies") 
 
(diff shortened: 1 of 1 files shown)


willmcgugan:
Good spot. I think there is an API wart there, in that the `index` value in Column should be public. I've fixed that in another branch as well as the docs. Thanks for drawing my attention to it.

PR #294, opened 2020-09-19
Fix broken HTML encoding

ssbarnea:
## Type of changes

- [x] Bug fix

## Checklist

## Description

Missing to specify encoding can mess some clients which may break the file encoding or render it incorrectly.

Example: file that loads ok locally, after upload to S3 gets broken if the encoding is not mentioned in file.

Use of UTF-8 is a very good practice, but we still need to mention it.

```
aws s3 cp report.html s3://bucket/report.html
```

The table drawing characters are those that are getting broken when not using UTF-8 but I can suspect that the same could happen with other Unicode characters too. [example](https://sbarnea.com/gri/ssbarnea.html) renders ok in Chrome but bad in Firefox, but once the charset is mentioned the issue is sorted.

Note: I don't think I deserve a mention in CONTRIBUTORS.md ;)


willmcgugan:
Thanks

PR #297, opened 2020-09-20
Added tests and fixed documentation

taliraj:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -920,7 +920,7 @@ class Console:
         """Prints a rich render of the last exception and traceback.
 
         Args:
-            code_width (Optional[int], optional): Number of characters used to render code. Defaults to 88.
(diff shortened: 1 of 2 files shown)


ssbarnea (approved):
Looks good by I am worried that the test may be too large and may break when we touch other stuff. Wouldn't be possible to make the test bit smaller just to test that wrapping happens as expected?

taliraj:
The `test_inspect_text` uses `Text` from `rich.text,` and indeed changes to `Text` may break it.
I will change it!
In the others tests I am testing the rendering of builtins like int, dict, print; unless I am missing something they shouldn't 
break by modifications in others parts of rich.

willmcgugan:
Looks good @taliraj Thanks. Probably is a good idea to use a builtin for that one test.

Issue #298, opened 2020-09-21
[REQUEST] Adding rich tracebacks support to RichHandler


taliraj:
Hi, I would like to contribute to rich with the following feature.
I am opening this issue in order to discuss the details of the implementation with the maintainers.

I want to modify RichHandler in order to use `print_exception` in case of exceptions.
It means that calls to `logger.exception()` would display rich tracebacks, and uncaught exceptions in django would also display rich tracebacks.
This is how it would look with django:

<pre>System check identified no issues (0 silenced).
September 21, 2020 - 18:09:32
Django version 3.1.1, using settings &apos;dj.settings&apos;
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
<font color="#116171">[18:09:48] </font><font color="#AA0000">WARNING</font>  Not Found: <font color="#AA22AA">/</font>                                                                                                                                                                                                                                     log.py:224
<font color="#116171">           </font><font color="#AA0000">WARNING</font>  <font color="#44AA44">&quot;</font><font color="#DED82C"><b>GET</b></font><font color="#44AA44"> / HTTP/1.1&quot;</font> <font color="#295FCC"><b>404</b></font> <font color="#295FCC"><b>2023</b></font>                                                                                                                                                                                                                   basehttp.py:157
<font color="#116171">[18:09:55] </font><font color="#AA0000">WARNING</font>  Not Found: <font color="#AA22AA">/</font>                                                                                                                                                                                                                                     log.py:224
<font color="#116171">           </font><font color="#AA0000">WARNING</font>  <font color="#44AA44">&quot;</font><font color="#DED82C"><b>GET</b></font><font color="#44AA44"> / HTTP/1.1&quot;</font> <font color="#295FCC"><b>404</b></font> <font color="#295FCC"><b>2023</b></font>                                                                                                                                                                                                                   basehttp.py:157
<font color="#FFFFFF"><b>Traceback</b></font> (most recent call last):
<font color="#0039AA">╭──────────────────────────────────────────────────────────────────────────────────────╮</font>
<font color="#0039AA">│ File </font><font color="#44AA44">&quot;/home/landi/Documents/dj_apps/django_traceback/env/lib/python3.8/site-packages/</font><font color="#0039AA">│</font>
[...]

PR #303, opened 2020-09-22
handle exceptions with rich's logging handler

taliraj:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Handling exceptions with rich's logging handler #298

### DIFF
--- a/rich/logging.py
+++ b/rich/logging.py
@@ -9,6 +9,7 @@ from ._log_render import LogRender
 from .console import Console
 from .highlighter import Highlighter, ReprHighlighter
 from .text import Text
+from .traceback import Traceback
 
 
 class RichHandler(Handler):
@@ -91,6 +92,16 @@ class RichHandler(Handler):
         level = Text()
         level.append(record.levelname, log_style)
 
+        traceback = None
+        if self.handle_tracebacks and record.exc_info:
+            traceback = Traceback(
+                width=self.tracebacks_width,
+                extra_lines=self.tracebacks_extra_lines,
+                theme=self.tracebacks_theme,
+                word_wrap=self.tracebacks_word_wrap,
+            )
+            message = record.getMessage()
+
         use_markup = (
             getattr(record, "markup") if hasattr(record, "markup") else self.markup
         )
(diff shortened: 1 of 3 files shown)


taliraj:
Well this is strange, `test_track_thread` of test_progress.py  failing on macos. I don't think I touched it...

willmcgugan:
There is a test involving a thread that fails randomly. Need to look at that. Rerunning tests tends to fix it

rich/logging.py
```diff
@@ -89,18 +104,26 @@ def emit(self, record: LogRecord) -> None:
...
-            self._log_render(
-                self.console,
-                [message_text],
-                log_time=log_time,
-                time_format=time_format,
-                level=level,
-                path=path,
-                line_no=record.lineno,
-                link_path=record.pathname if self.enable_link_path else None,
+        if self.handle_tracebacks and record.exc_info:
```
willmcgugan, line 107: This will render the exception without the other configured columns (time, severity, filename). Logged exceptions also have a message which I don't think we're displaying here.

You can capture the current exception with `rich.exception.Exception` then add that to the log renderer. i.e. if there is an exception, call `_log_render` with `[message_text, my_exception]`

willmcgugan (changes requested):
This is a nice addition. Added a comment re columns.

taliraj:
The `exc_value `of the traceback does not get printed.


Capture_rich

I tried to find out why and I think it disappears in the print function, at this point:

https://github.com/willmcgugan/rich/blob/a16a234c2906665279f5fc01fd138ff42e4c7dbb/rich/console.py#L958-L959

Thou I don't know why...

rich/logging.py
```diff
@@ -76,6 +92,16 @@ def emit(self, record: LogRecord) -> None:
         level = Text()
         level.append(record.levelname, log_style)
 
+        traceback = None
+        if self.handle_tracebacks and record.exc_info:
```
willmcgugan, line 96: The Traceback will need to be constructed with the exception in `record.exc_info`, since it's possible to log an exception that wasn't necessarily the most recent one. See [Traceback.from_exception](https://rich.readthedocs.io/en/latest/reference/traceback.html#rich.traceback.Traceback.from_exception)

willmcgugan:
@taliraj That is strange. Any change you can print `traceback.stacks`. If you see the exc_value there then something must have gone wrong in rendering.

taliraj:
> 
> 
> @taliraj That is strange. Any change you can print `traceback.stacks`. If you see the exc_value there then something must have gone wrong in rendering.

@willmcgugan  I added a `print(traceback.trace.stacks)`, this is the result:

Capture_rich_stack

Indeed the `exc_value` is there.

willmcgugan:
Still not sure what is happening to that exc_value, but it does look to be on the render side. I'll merge this and fix. Thanks!

PR #311, opened 2020-09-30
Add reference to rich usage case 'norfair' in README

joaqo:
As proposed in https://github.com/willmcgugan/rich/issues/309


willmcgugan:
Thanks.

Issue #323, opened 2020-10-02
[Suggestion] Use binary prefixes instead of decimal ones


Rubo3:
I was tinkering with progress bars and noticed the file size unit is expressed with [decimal prefixes](https://github.com/willmcgugan/rich/blob/33a171d72013f470c0edb5f61e226ca80729cbf6/rich/progress.py#L301) (i.e. KB, MB, GB, etc.) while the actual number is expressed as binary. I suggest you to consider switching to binary prefixes (i.e. KiB, MiB, GiB, etc.) in order to have consistency with the number and its unit of measure and to remove the ambiguity regarding in which base the prefix is expressed: KiB, MiB, etc. are used exclusively with powers of 2, while I've seen KB, MB, etc. used with powers of 10 (as they should be) and powers of 2 (as in this case, but generally in casual conversations).
For instance, I tried with a file with a size of 6'988'379 bytes. Expressed in MiB it's 6,7 and this is the number printed by the progress bar, while in MB it's 6,9. As you can notice, while dealing with MiB vs MB the difference is ~5%, but when the size is bigger it can be ~7% (GiB vs GB), ~10% (TiB vs TB) and so on, so please consider using only one format.

Edit: grammar

willmcgugan:
PRs welcome

willmcgugan:
I think download size should be expressed in decimal prefixes for consistency with file managers, as it is what the user would expect. KiB MiB etc. may be unambiguous, but they are just not in common usage.

So the fix would be changing 1024 to 1000.

Rubo3:
It depends. Dolphin, for examples, expresses it with binary prefixes.

devratapuri:
hi i am new to this can  anyone tell me from where to begin

willmcgugan:
@Rubo3 will be fixed in next version. Thanks to @amartya-dev

Rubo3:
Thank you. What about giving the user the possibility to switch between decimal and binary?

willmcgugan:
Good idea

amartya-dev:
We can give a command-line argument for that flag?

willmcgugan:
@amartya-dev Rich doesn't handle command line switches, except for a few built-in commands, but the application could chose to expose that via a CLI

amartya-dev:
@willmcgugan I guess the way to incorporate this is using a parameter in `render()` function of `DownloadColumn` ?

willmcgugan:
It would be a parameter in the constructor.

[...]

PR #326, opened 2020-10-03
Change to binary prefixes in progress.py

amartya-dev:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

Incorporates #323 

I have just changed the prefixes to binary as suggested in the issue

Edit: Checked off a few more boxes and added the description at the correct place

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [8.0.1] - unreleased
+
+### Changed
+
+- Change the render prefix to correspond to the decimal units in progress
+
 ## [8.0.0] - 2020-10-03
 
 ### Added
@@ -19,7 +25,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed Style.empty to Style.null to better reflect what it does
 - Optimized combining styles involving a null style
 - Change error messages in Style.parse to read better
-- Change the render prefix to correspond to the decimal units in progress
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed Style.empty to Style.null to better reflect what it does
 - Optimized combining styles involving a null style
 - Change error messages in Style.parse to read better
+- Change the render prefix to correspond to the decimal units in progress
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


rich/progress.py
```diff
@@ -299,7 +299,9 @@ def render(self, task: "Task") -> Text:
         completed = int(task.completed)
         total = int(task.total)
         unit, suffix = filesize.pick_unit_and_suffix(
-            total, ["bytes", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"], 1024
+            total,
```
willmcgugan, line 302: I think filesize should be in decimal units, to be consistent with file managers. i.e. same units, just with the third parameter as 1000.

willmcgugan:
Great. Please add a test for this and an entry in the README.

rich/progress.py
amartya-dev: Sure let me change that

willmcgugan (changes requested):

tests/test_progress.py
```diff
@@ -254,6 +254,40 @@ def test_columns() -> None:
     assert result == expected
 
 
+def test_download_progress_uses_decimal_units() -> None:
```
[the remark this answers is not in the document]
amartya-dev: Ok, I will look into it...

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed Style.empty to Style.null to better reflect what it does
 - Optimized combining styles involving a null style
 - Change error messages in Style.parse to read better
+- Change the render prefix to correspond to the decimal units in progress
```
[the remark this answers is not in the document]
amartya-dev: Ok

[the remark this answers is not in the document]
amartya-dev: What date should I use there?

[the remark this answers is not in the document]
willmcgugan: Just say "unreleased" for now. https://semver.org/spec/v2.0.0 is a worthwhile read.

amartya-dev:
@willmcgugan could you please check and let me know if any further changes are required?

tests/test_progress.py
```diff
@@ -79,6 +79,28 @@ class FakeTask(Task):
     assert str(text) == "0:01:00"
 
 
+def test_download_progress_uses_decimal_units() -> None:
```
willmcgugan, line 82: Better, but too complex for a test.

I see you're trying to be thorough (which is good!) but a test should be no more complex that is needed to catch a bug.

Have a look at `test_time_remaining_column`, you'll notice that there are no loops, just assignments and method calls, but that's enough to execute every line of the changed code. I think you can do something similar here...

willmcgugan (changes requested):

tests/test_progress.py
amartya-dev: Ok, makes sense, I thought I will add test all the symbols. Should I just create a task say half completed and just assert the rendered string?

willmcgugan: Sounds reasonable 👍

amartya-dev:
@willmcgugan are there any further changes required?

willmcgugan:
LGTM Thanks.

PR #334, opened 2020-10-04
add tests for syntax

phase06:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

#37 Added more test for syntax.py. Please review.


tests/test_syntax.py
```diff
@@ -46,6 +46,42 @@ def test_python_render():
...
+        lexer_name="python",
+        line_numbers=True,
+        line_range=(2, 10),
+        theme="foo",
+        code_width=60,
+        word_wrap=True,
+        background_color="red",
+    )
+    text = syntax.highlight(CODE)
+    assert syntax.highlight(CODE) == text
```
willmcgugan, line 82: The test is guaranteed to pass.

willmcgugan, line 82: This test is guaranteed to pass. So not much point in having it.

tests/test_syntax.py
```diff
@@ -46,6 +46,42 @@ def test_python_render():
...
+    )
+    from pygments.style import Style as PygmentsStyle
+
+    style = PygmentsStyle()
+    assert syntax.get_theme(style).get_background_style().bgcolor.name == "#ffffff"
+
+
+def test_pygments_syntax_theme():
+    style = PygmentsSyntaxTheme("default")
+    assert style.get_style_for_token("abc") == Style.parse("none")
```
willmcgugan, line 67: Adds a line of coverage and the test is meaningful :-) 

There are a few more lines in `get_style_for_token` that you could check here.

tests/test_syntax.py
```diff
@@ -46,6 +46,42 @@ def test_python_render():
...
+        line_numbers=True,
+        line_range=(2, 10),
+        theme="foo",
+        code_width=60,
+        word_wrap=True,
+    )
+    from pygments.style import Style as PygmentsStyle
+
+    style = PygmentsStyle()
+    assert syntax.get_theme(style).get_background_style().bgcolor.name == "#ffffff"
```
willmcgugan, line 62: `get_theme` is a class method, and doesn't need an instance.

willmcgugan (changes requested):
I'll need a bit more consideration re what is being tested please. The assert should be meaningful, i.e. it should fail if there is a bug.

If you could get syntax.py to 100% it would be 🥇

tests/test_syntax.py
phase06: yeah, will update the test

tests/test_syntax.py
phase06: got it

tests/test_syntax.py
phase06: will update the test

phase06:
@willmcgugan , Got syntax.py and pretty.py to 100%. Please review 🥇

willmcgugan:
Fantastic. Thanks!

PR #335, opened 2020-10-04
Added An Example to show the use of Rainbow text with an Array output

NicholasBlackburn1:
Added An Sample to show users how to use rainbow text with an array and a for a loop.


NicholasBlackburn1:
Finished one Sample use of the library

willmcgugan:
What feature of Rich does this demonstrate?

NicholasBlackburn1:
this demonstrates on how to use the rainbow text in an supplementary way to the orginal rainbow text example.

NicholasBlackburn1 (commented):
Fixed Misleading file name to rainbowArrays to Comply with python standards.

willmcgugan:
Sorry. I don't think this adds any value. Examples should demonstrate a feature not already covered.

PR #347, opened 2020-10-06
Make soft_wrap preference configurable per Console

ssbarnea:
Allows user to define default soft_wrap behavior when creating Console instances. This allows us to control wrapping without having to modify each print() statements, something that may even
be impossible if called from within rich library itself.

## Type of changes

- [x] New feature

## Checklist

## Description

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Change the render prefix to correspond to the decimal units in progress
+- Add ability to mention soft_wrap value when initializing Console
 
 ## [8.0.0] - 2020-10-03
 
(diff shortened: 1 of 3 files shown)


willmcgugan:
I wouldn't accept this. It would break third party code if soft_wrap is enabled by default, forcing everyone to add `soft_wrap` to their print statements to be sure it would output what they expect.

If you need it, I'd suggest writing a function that calls print with soft_wrap=False.

ssbarnea:
> I wouldn't accept this. It would break third party code if soft_wrap is enabled by default, forcing everyone to add `soft_wrap` to their print statements to be sure it would output what they expect.
> 
> If you need it, I'd suggest writing a function that calls print with soft_wrap=False.

This change should not change behavior of existing code in any way. Only those that will start adding `Console(soft_wrap=False)` would be able to do it.

I will have another look at the code, as my main goal was to not change behavior.

willmcgugan:
If you print a third party renderable to a console with soft wrapping disabled, it may product broken output. It doesn't really fix anything, just save a little typing, at the expense of potentially breaking thing. Sorry, can't merge this.

Issue #295, opened 2020-09-19
Add a tox.ini file to ease local testing


ssbarnea:
As tox.ini still seems to be the ubiquitous for managing local testing for any python project, I would highly appreciate if I can make use of it to run tests.

I would be happy to propose of change for that but I considered better to ask first, as I encountered at least one case were the owner disliked it.

ssbarnea:
It would be useful to disable the codecov comments on PRs, it is spammy (both email and notificationsl) and does not add any real value. We still have the action that can fail if coverage fails. I did this on several projects, is one line like https://github.com/pycontribs/jira/blob/master/codecov.yml#L2

willmcgugan:
I'd accept a PR to add tox. Although I'd prefer a solution to run GH actions locally.

I like the codecov reports. I wouldn't want to enforce coverage at this point.

reillysiemens:
@willmcgugan: I have a little experience working with tox and I'd be happy to add a configuration if you like. I'm interested to know more about what you mean by
> I'd prefer a solution to run GH actions locally.

though. Do you mean you'd like to run whatever the GH actions run locally or do you mean you'd like to trigger a GH action from a local testing environment? :thinking:

willmcgugan:
This [tool](https://github.com/nektos/act) can run GH Actions locally. I'm not sure how practical that is, but it's worth investigating. Otherwise, tox is a reasonable solution.

reillysiemens:
@willmcgugan: I'll give [`act`](https://github.com/nektos/act) a look and report back here then. For comparison, in my own project I'm [running tox through Poetry in a GH action](https://github.com/reillysiemens/ipython-style-gruvbox/blob/77518384a12e30187aeca7f50910ed154eca7c20/.github/workflows/test.yml#L22-L25) quite nicely.
```yaml
- name: Test
  run: |
    source $HOME/.poetry/env
    poetry run tox -q --parallel=auto
```

reillysiemens:
@willmcgugan: So, I went ahead and installed/ran [`act`](https://github.com/nektos/act) locally and the results are, to put it mildly, a little confusing...
```
$ act pull_request
...
[Test Rich module/build-6] ⭐  Run Set up Python ${{ matrix.python-version }}
[Test Rich module/build-6]   ☁  git clone 'https://github.com/actions/setup-python' # ref=v1
[...]

PR #348, opened 2020-10-06
Add tox configurations

ssbarnea:
## Type of changes

- [x] Other

## Checklist

## Description

Fixes: #295

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,45 @@
+[tox]
+minversion = 3.9.0
+envlist =
+    lint
+    docs
+    py{36,37,38,39}
+isolated_build = True
+
+[testenv]
+description = Run unit-testing
+# develop temporary disabled as project packaging does not work with it yet:
+# https://github.com/willmcgugan/rich/issues/345
+usedevelop = False
+deps =
+    -r requirements-dev.txt
+# do not put * in passenv as it may break builds due to reduced isolation
+passenv =
+    CI
+    GITHUB_*
+    HOME
+    PYTEST_*
+    SSH_AUTH_SOCK
+    TERM
+setenv =
+    PYTHONDONTWRITEBYTECODE=1
+    PYTHONUNBUFFERED=1
+commands =
+    # failsafe as older pip may install incompatible dependencies
+    pip check
+    pytest --cov-report term-missing --cov=rich tests/ {posargs}
+
+[testenv:lint]
+description = Runs all linting tasks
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -0,0 +1,40 @@
+[tox]
+minversion = 3.9.0
+envlist =
+    lint
+    py{36,37,38}
+
+# do not enable skip missing to avoid CI false positives
+skip_missing_interpreters = False
```
willmcgugan, line 8: Won't this mean that it won't work for users without all three versions of Python installed?

tox.ini
```diff
@@ -0,0 +1,40 @@
...
+isolated_build = True
+
+[testenv]
+# develop temporary disabled as project packaging does not work with it yet:
+# https://github.com/willmcgugan/rich/issues/345
+usedevelop = False
+# do not put * in passenv as it may break builds do to reduced isolation
+deps =
+    -rrequirements-dev.txt
```
willmcgugan, line 17: Should there not be a space after `-r` ?

ssbarnea: Not needed, pip works fine with both. I used that because on openstack projects is seems to be the practice, but I am happy to add the space if you want.

tox.ini
ssbarnea: There is a long story about this, I will find some links tomorrow. All openstack projects have this added http://codesearch.openstack.org/?q=skip_missing_interpreters&i=nope&files=tox.ini&repos= -- with extra comment to avoid accidental regressions.

Shortly we decided to enforce it to this value because it happened multiple times to get some faile positive that nobody notices for long time.

You can blame tox for this but if you set this to true and run `tox -e py38` on a machine that does not have python3.8 already installed, you will get a success exit code!

As you can expect, the downside is that running just "tox" may no longer return 0 if you do not have all supported versions of python installed. I do not remember seeing complaints, especially after they understood the dangers.

willmcgugan: The purpose of using tox was to ease local testing, not necessarily for CI. To that end, I think we should allow `tox` to run without the `-e` switch, even if the user doesn't have three version of Python.

ssbarnea:
Apparently codecove is facing some outages, so do not trust its reported failure.

ssbarnea:
@willmcgugan Any change to look at this one? I think I addressed all your original questions.

willmcgugan:
Thanks.

PR #349, opened 2020-10-06
Update pythonpackage.yml

willmcgugan:
Python3.9 doesn't seem to be available by default on GH Actions.


ssbarnea:
Give them few days, 3.9.0 is not even present on pyenv and they are known to be from the first to do it (previous betas and dev are). GitHub Actions team moves at decent speed, I suspect less than a month to get it.

If you are in a hurry you can use the beta version mentioned on https://github.com/actions/setup-python

Issue #346, opened 2020-10-06
[BUG]


ssbarnea:
**Describe the bug**
When trying to install rich in editable mode user may get into error below. This is likely caused by current packaging configuration.

**To Reproduce**
```
$ pip install -e .
Obtaining file:///Users/ssbarnea/c/os/rich
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Requirement already satisfied: typing-extensions<4.0.0,>=3.7.4 in /Users/ssbarnea/.pyenv/versions/3.8.2/lib/python3.8/site-packages (from rich==7.0.0) (3.7.4.2)
Requirement already satisfied: pygments<3.0.0,>=2.6.0 in /Users/ssbarnea/.pyenv/versions/3.8.2/lib/python3.8/site-packages (from rich==7.0.0) (2.6.1)
Requirement already satisfied: colorama<0.5.0,>=0.4.0 in /Users/ssbarnea/.pyenv/versions/3.8.2/lib/python3.8/site-packages (from rich==7.0.0) (0.4.3)
Requirement already satisfied: commonmark<0.10.0,>=0.9.0 in /Users/ssbarnea/.pyenv/versions/3.8.2/lib/python3.8/site-packages (from rich==7.0.0) (0.9.1)
Installing collected packages: rich
  Running setup.py develop for rich
    ERROR: Command errored out with exit status 1:
     command: /Users/ssbarnea/.pyenv/versions/3.8.2/bin/python3.8 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/Users/ssbarnea/c/os/rich/setup.py'"'"'; __file__='"'"'/Users/ssbarnea/c/os/rich/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps
         cwd: /Users/ssbarnea/c/os/rich/
    Complete output (3 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
    ModuleNotFoundError: No module named 'setuptools'
    ----------------------------------------
[...]

PR #354, opened 2020-10-07
Fixed pip editable install

ssbarnea:
Fixes: #346

## Type of changes

- [x] Bug fix

## Description

This should bring make editable installs with pip working again. It also includes a minimal CI testing of editable installs, just to avoid future regression.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -31,6 +31,9 @@ jobs:
         run: |
           pip install .
           python -m pytest tests -v --cov=./rich --cov-report=xml:./coverage.xml --cov-report term-missing
+      - name: Check if pip editable installs still work
+        run: |
+          pip install -e .
 
       - name: Upload code coverage
         uses: codecov/codecov-action@v1.0.10
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,5 +37,8 @@ jupyter = ["ipywidgets"]
 [tool.poetry.dev-dependencies]
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
+requires = [
+    "poetry-core>=1.0.0",
+    "setuptools",  # keep it here or "pip install -e" would fail
+    ]
 build-backend = "poetry.core.masonry.api"


willmcgugan:
`poetry install` does an editable install.

Tested with a fresh venv, and it works fine. rich.__file__ points to the local dir, not to site packages.

ssbarnea:
I would have said the same yesterday when it worked like you said. Still, this morning I started to wonder why it no longer works.

I should mention that, by default, I do not have a virtualenv/venv enabled.

I tried creating a venv and did the same test and it worked. This may explain why it also worked for you. Try to run it when you do not have any venv active.

willmcgugan:
It wouldn't work for me anyway. Py27 is still the default in MacOS

Regardless, installing in your system Python is not a good idea. And I don't want to muddy the waters with two ways of installing it for development.

If you have identified a problem with installs, you might want to post an issue on poetry repos.

ssbarnea:
This PR has nothing to do with poetry itself, it does fix a packaging bug in this project, one that is very easy to spot: "setuptools" is a build time dependency that was missed to be added to `build-system` section.

https://www.python.org/dev/peps/pep-0517/ should explain it better, [poetry-core](https://www.python.org/dev/peps/pep-0517/) is relying on it.

Still, poetry-core itself does not directly depend on setuptools, but setuptools is needed for performing editable installs, that is a known limitation. Keep in mind that this requirement is a build-time, adding it there does not install it in destination environment. The build system has uses a temporary virtualenv solely for building.

**Why do you prefer to force a developer to install poetry?**


I made another test using native python3 from macos:
```
$ /usr/bin/python3 -m poetry install
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: rich (8.0.0)

$ /usr/bin/python3 -m pip freeze|grep rich
FAIL: 1

# test that rich package was installed (avoid accidental import of rich from cwd!)
$ cd / && /usr/bin/python3 -c "import rich"                                                                               Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'rich'
```

I tested even on CentOS 8 using default system python, some issue! 

Poetry does not install the package to current python environment unless that one is a virtualenv.

It does create a temp virtualenv but clearly that is something very different than installing a package.

The only reason why you may be able to succeed running "python -c 'import rich'" is due to the fact there is a rich folder at the root level of repo project, but that is not an installed package. I seen few others strongly recommending keeping packages in a subdirectory like src/ or lib/ in order to avoid accidental spills but I only started to do it this myself recently)

willmcgugan:
Like I said, it's a bad idea to do an editable install in system python.

Use a venv, it works regardless of current working directory.

Please read https://python-poetry.org/docs/basic-usage/ Particularly the section on virtual envs.

ssbarnea:
Just to be clear: these is "user level" install, not root level. Newer versions of pip finally made the switch of avoiding to install on system python by default, installed packages will go by default to `~/.local/lib/python3.x/`, which also takes precedence over system packages. -- that user stuff is nothing new, probably is a decade old.

PR #356, opened 2020-10-07
Avoid codecov spam

ssbarnea:
Current configuration of codecov does SPAM not only github user interface but also our inboxes.

This disables the codecov comment, but not the tool itself. The tool still runs and reports using check API, being visible like any other github action.

See: https://docs.codecov.io/docs/pull-request-comments

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,1 @@
+comment: false


willmcgugan:
I think I explained previously that I find the codecov comment useful.

PR #361, opened 2020-10-08
Add block bar

megaserg:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This implements a solid block bar, as requested in https://github.com/willmcgugan/rich/issues/293.
Try with: `python -m rich.block_bar`

**Implemented:**
- Block elements support 
  - left-aligned bars with 1/8-of-a-cell granularity on the right side
  - right-aligned bars with 1/2-of-a-cell granularity on the left side
- Render nothing when value is min

**Not implemented:**
- Render nothing when value is max
- Color gradient blends between colors

**Note:** `BlockBar` might not be the best name. Let me know!


willmcgugan:
Bravo for attempting this.

It might be a mistake to duplicate some of the arguments from the progress bar. I can see this being a more general purpose thing, which would have different uses beyond progress.

I'm thinking it should have a width (in cells), plus a start and end argument as a float (which could be internally converted to 1/8ths). Plus a foreground and background color.

You're absolutely correct that there are only left aligned block characters. But flip the foreground and background colors, and you have the equivalent of right aligned. :-)

rich/bar.py
```diff
@@ -195,6 +195,118 @@ def __rich_measure__(self, console: Console, max_width: int) -> Measurement:
...
+
+    def __init__(
+        self,
+        total: float = 100,
+        completed: float = 0,
+        width: int = None,
+        style: StyleType = "bar.back",
+        complete_style: StyleType = "bar.complete",
+        finished_style: StyleType = "bar.finished",
+        blend_colors: bool = False,
```
willmcgugan, line 223: I think that we probably don't want to implement blending here. This class could be a primitive that you could extend to add the blend functionality.

It should probably use colors, rather than styles. Since much of the attributes (italic / bold) etc don't make sense for a solid bar.

Here's what I'm thinking for the constructor:

```
width: Optional[float],  # Number of cells to use, or None for full width
size: float, # Value for end of the bar
start: float,  # Start point (in 0 -> size units)
end: float,  # End point
color: Union[Color, str] = "default",
background: Union[Color, str] = "default", 
```
A union for the colors, so you can pass a Color or a string to call Color.parse with.

What do you think?

megaserg:
> You're absolutely correct that there are only left aligned block characters. But flip the foreground and background colors, and you have the equivalent of right aligned. :-)

That's a nice trick @willmcgugan! I got it working. But in my iTerm, if I set some non-zero window transparency, the background color from the flipped Style looks faded compared to the foreground color from the non-flipped Style, unless I set checkbox "Keep background colors opaque". Is there anything I can do? 🤔 

Also, how can I flip colors in case they are both `"default"`?

willmcgugan:
>  But in my iTerm, if I set some non-zero window transparency, the background color from the flipped Style looks faded compared to the foreground color from the non-flipped Style, unless I set checkbox "Keep background colors opaque". Is there anything I can do? 🤔

Ah that's a shame. 

There is an "right half block" character (▐) . You could have the 1/8 granularity for the right hand side of the bar, and the 1/2 block granularity for the left.

> Also, how can I flip colors in case they are both "default"?

Good point. If we don't use the flipping colors trick and use the right half block, then this won't be an issue.

Alternatively we could disallow default colors for the bar, and document the transparency issue.

I'll leave it up to you.

BTW would love to see screenshots!

megaserg:
Sure, this is a checkbox I'm talking about:
checkbox
I've made a bar with `color = "bright_green"` and `bgcolor = "bright_red"`.

With checkbox unchecked, notice how on the light background the rightmost red (where it's the background color) is less saturated than the leftmost red (where it's the foreground color after flip).
unchecked_light
Similarly, on the dark background a bit of green on the left (where it's the background color after flip) is less saturated than the the rest of the green (where it's the foreground color).
unchecked_dark

With checkbox checked, it's looking good on light and dark backgrounds:
checked_light
checked_dark

willmcgugan:
Just had a thought. Rather than swap the foreground / background manually, what if you were to set `inverse=True` on the style? I think that may work better. And it should also solve the "default on default" issue.

willmcgugan:
Would you mind also adding an example in `examples/` ?

Something very simple, a few lines that show how you might use it in practice.

megaserg:
Do you mean `reverse=True`? Yes, this solves the "default on default" issue, but the problem of non-opaque background color is still there.

willmcgugan:
Thanks @megaserg This looks great. I'll merge this soon.

willmcgugan:
Thanks @megaserg !

PR #362, opened 2020-10-08
Fix typos, comments, unused imports

megaserg:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Just some semi-automatically found typos.

### DIFF
--- a/rich/style.py
+++ b/rich/style.py
@@ -345,7 +345,7 @@ class Style:
         return self._link
 
     @property
-    def transaprent_background(self) -> bool:
+    def transparent_background(self) -> bool:
         """Check if the style specified a transparent background."""
         return self.bgcolor is None or self.bgcolor.is_default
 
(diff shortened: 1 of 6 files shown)


rich/style.py
```diff
@@ -345,7 +345,7 @@ def link(self) -> Optional[str]:
         return self._link
 
     @property
-    def transaprent_background(self) -> bool:
+    def transparent_background(self) -> bool:
```
willmcgugan, line 348: This one is technically a breaking change. Would you mind documenting this in  CHANGELOG.md

willmcgugan (commented):
Looks good. If you wouldn't mind documenting the silly typo with background_style in the changelog.

rich/style.py
ssbarnea: Indeed breaking but I doubt anyone used it and failed to make a PR or at least raise a bug. Still, updating changelog seems reasonable.

ssbarnea (approved):

rich/style.py
megaserg: Done! FWIW, here's a full GitHub search for that token: https://github.com/search?q=transaprent_background&type=code 😅

willmcgugan:
Thanks @megaserg I'll take that as a lesson to be careful about cutting and pasting typos!

PR #365, opened 2020-10-08
Additional tests for `measure.py`, `table.py`, `padding.py` and `containers.py`, as well as some possible code fixes

MoltenMuffins:
## Type of changes

- [x] Bug fix (See description under `table.py` and `containers.py`)
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Some additional tests in response to #37 for the following files:
* `measure.py`
    * `test_null_get`: Checks for cases where `Measurement.get()` should return `Measurement(0, 0)`
    * `test_measure_renderables`: Tests that `measure_renderables` returns the correct `Measurement` object 
* `table.py`
    * `test_append_column`: Checks whether headers are correctly added to `self.columns` in `Table`. I'm not too sure if this test is meaningful enough though.
    * `test_rich_measure`: Checks whether the `__rich_measure__` method of `Table` returns the correct `Measurement` object for several cases
    * Tried to add a test for lines 496-497, however I can't engineer a case where the code in `try` would raise a `ValueError`:
        ```python
        try:
            second_max_column = max(
                width if allow_wrap and width != max_column else 0
                for width, allow_wrap in zip(widths, wrapable)
            )
        except ValueError:
            second_max_column = 0
        ```
        Afaik, a `ValueError` would only be thrown by `max()` in a situation where an empty sequence is passed to it. But for us to get to the `try/except`, both `widths` and `wrapable` would have to be valid, which would in turn generate valid arguments for the `max()` function. Because of this, I'm not convinced a `try/except` for a `ValueError` is necessary in the function and hence removed it from the method..
* `padding.py`
    * `test_indent`: Tests the `indent` method of `Padding`. Also not sure if this test is meaningful enough.
    * `test_rich_console`: Tests the output of the `__rich_console__` method
* `containers.py`
    * `test_lines_justify` Tests the outputs of the `justify` method for `Lines`
    * It seems the `if next_word is None:` condition in line 156 of `conatiners.py` will never be triggered as `next_word` will only be `None` if `index == len(spaces)`, however the preceeding `if index < len(spaces):`  skips the execution of the `if next_word is None:` condition.
    ```python
    if index < len(spaces): 
            if next_word is None: # Only occurs if index==len(spaces)?
                space_style = Style.null()
            else:
                style = word.get_style_at_offset(console, -1)
                next_style = next_word.get_style_at_offset(console, 0)
                space_style = style if style == next_style else line.style
    ```
    Propose removing the if/else like so and have made a corresponding change in the PR:
    ```python
    if index < len(spaces): 
            style = word.get_style_at_offset(console, -1)
            next_style = next_word.get_style_at_offset(console, 0)
            space_style = style if style == next_style else line.style
    ```
Looking forward to your feedback!

### DIFF
--- a/rich/containers.py
+++ b/rich/containers.py
@@ -148,7 +148,6 @@ class Lines:
                         num_spaces += 1
                         index = (index + 1) % len(spaces)
                 tokens: List[Text] = []
-                index = 0
                 for index, (word, next_word) in enumerate(
                     zip_longest(words, words[1:])
                 ):
@@ -161,5 +160,4 @@ class Lines:
                             next_style = next_word.get_style_at_offset(console, 0)
                             space_style = style if style == next_style else line.style
                         tokens.append(Text(" " * spaces[index], style=space_style))
-                    index += 1
                 self[line_index] = Text("").join(tokens)
--- a/rich/containers.py
+++ b/rich/containers.py
@@ -153,11 +153,8 @@ class Lines:
                 ):
                     tokens.append(word)
                     if index < len(spaces):
-                        if next_word is None:
-                            space_style = Style.null()
-                        else:
-                            style = word.get_style_at_offset(console, -1)
-                            next_style = next_word.get_style_at_offset(console, 0)
-                            space_style = style if style == next_style else line.style
+                        style = word.get_style_at_offset(console, -1)
+                        next_style = next_word.get_style_at_offset(console, 0)
+                        space_style = style if style == next_style else line.style
                         tokens.append(Text(" " * spaces[index], style=space_style))
                 self[line_index] = Text("").join(tokens)
--- a/rich/table.py
+++ b/rich/table.py
@@ -488,13 +488,10 @@ class Table(JupyterMixin):
                 max_column = max(
                     width for width, allow_wrap in zip(widths, wrapable) if allow_wrap
                 )
-                try:
-                    second_max_column = max(
-                        width if allow_wrap and width != max_column else 0
-                        for width, allow_wrap in zip(widths, wrapable)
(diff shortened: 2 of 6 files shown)


willmcgugan:
@MoltenMuffins This is looking good.

You may be correct about that ValueError.

I don't understand why codecov is saying the coverage decreased!

MoltenMuffins:
@willmcgugan Hi, thanks for looking at my WIP! I was trying to figure out where the coverage decrease was coming from, but it seems it isn't from any of the files I made changes to (provided I'm intepreting the codecov report correctly). I have also added tests for `containers.py` to the PR and have marked it ready for review.

I also realize that codecov ignores several files that I have added tests for, should these be removed?

MoltenMuffins:
Looking at the [codecov report](https://codecov.io/gh/willmcgugan/rich/pull/365?src=pr&el=continue)
It seems the following files are being ignored?
* ‘tests/test_containers.py’
* ‘tests/test_table.py’
* ‘tests/test_measure.py’
* ‘tests/test_padding.py’
* ‘tests/table.py’

Screenshot:
1A90012B-BA13-4D3C-B42F-085965128EF1

willmcgugan:
That is strange. The tests/ directory is excluded from coverage, but I have no idea why it would ignore table.py.

Nothing in the configuration I could see that could cause this, and I'm sure it was working previously.

Safe to say, something has gone wrong with codecov. This PR has definitely increased coverage.

willmcgugan:
Thanks.

MoltenMuffins:
Glad I could help! :+1:

Issue #355, opened 2020-10-07
[REQUEST] Make use of pre-commit for running linters


ssbarnea:
I would like to suggest using [pre-commit](https://pre-commit.com/) tool (not hook!) for orchestrating linters during development. Currently rich has two linters: black and mypy*.

**How would you improve Rich?**

It would be much easier to run linting, much easier to enable additional checks (like newline consistency) and make is extremely easy to bump linters.

Projects like black, tox, pytest are already using it for a good number of years.

**What problem does it solved for you?**

It would make easier to test code locally before making a PR. 

If you do not have anything against it, I can easily make a PR to demonstrate it (probably less than 5min to do it).

willmcgugan:
I'm willing to give it a try. I think @hedythedev found some issues with pre-commit, but as long as its optional I don't see a problem.

ssbarnea:
I am quite curious which issues @hedythedev identified.  I even made a [pre-commit presentation](https://sbarnea.com/slides/pre-commit/#/) back in March around how it works.

hedythedev:
I had trouble with installing the hooks, but will try again when I have the time. However I think it's a good idea overall

ssbarnea:
While I used pre-commit myself for probable more than 3 years, I almost never use its hook feature. I prefer to be able to bypass it and I got a habbit of running it manually.

I even have `alias pc='pre-commit run -a'` in my shell. The real value is not the hook, but how easy it makes to manage the lifecycle of the linters (add new, upgrade, run, avoid taking too much disk space, isolate/contain). If curious about a more heavy use of it, take a look at https://github.com/ansible-community/molecule/blob/master/.pre-commit-config.yaml

Edward-Knight:
It might also be useful to add some of the built-in pre-commit checks as well (like `end-of-file-fixer` or `trailing-whitespace`)

ssbarnea:
I did not want to pollute the original PR with new checks but my plan was to do it in follow-ups, so we can keep them small and easy to review.

henryiii:
This might be helpful; in Scikit-HEP we are using pre-commit everywhere. https://scikit-hep.org/developer/style. Also you can use https://pre-commit.ci if you want.

PR #366, opened 2020-10-08
Adopts pre-commit for running already existing linters

ssbarnea:
## Type of changes

- [x] Tests

## Description

- Adds "make lint" as generic command
- Linters are removed from requirements-dev.txt as they are now managed by pre-commit itself. This avoids potential conflicts between linters as pre-commit installs each linter on its own
  environment.
- To be reviewed after https://github.com/willmcgugan/rich/pull/348 is merged as it builds on top of it
- Please note that while this change does not bring any apparent new feature, it does enable us to add and manage extra linters (like flake8 or pylint) much easier.

Fixes: #355

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,17 @@
+# to bump linters just run: pre-commit autoupdate
+repos:
+  - repo: https://github.com/psf/black
+    rev: 20.8b1
+    hooks:
+      - id: black
+        language_version: python3
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.790
+    hooks:
+    - id: mypy
+      # empty args needed in order to match mypy cli behavior
+      pass_filenames: false
+      args: ["--ignore-missing-imports", "--warn-unreachable", "--html-report", "mypy_report"]
+      entry: mypy -p rich
+      additional_dependencies:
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+    hooks:
+      - id: black
+        language_version: python3
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.782
+    hooks:
+    - id: mypy
+      # empty args needed in order to match mypy cli behavior
+      pass_filenames: false
+      args: ["--ignore-missing-imports", "--warn-unreachable", "--html-report", "mypy_report"]
```
willmcgugan, line 14: Do we need the html report on commit?

It's quite slow to generate, and it's not always needed.

Makefile
```diff
@@ -1,13 +1,11 @@
...
-format-check:
-	black --check .
+lint:
+	pre-commit run -a
 format:
-	black .
+	pre-commit run black -a
 typecheck:
-	mypy -p rich --ignore-missing-imports --warn-unreachable
-typecheck-report:
```
willmcgugan, line 9: I'd like to be able to generate the mypy report as needed.

ssbarnea: I put it back.

Issue #364, opened 2020-10-08
[BUG] Typo in repr.ipv4 regex?


deppiedave64:
**Describe the bug**
In the `ReprHighlighter` class, there seems to be a typo in the regex for IPV4 addresses:
`r"(?P<ipv4>[0-9]{1,3}\.[0-9]{1,3}\.[0-gt9]{1,3}\.[0-9]{1,3})",`
The third octet will also accept the letters 'g' and 't', which causes strings like '192.168.ggt.1' to be highlighted.

**To Reproduce**
```python
>>> from rich.console import Console
>>> c = Console()
>>> c.print('192.168.1.1')
192.168.1.1
>>> c.print('192.168.t.1')
192.168.t.1
>>> c.print('192.168.gt.1')
192.168.gt.1
>>> c.print('192.168.ggt.1')
192.168.ggt.1
```

See also screenshot:
![Screenshot_20201008_094247](https://user-images.githubusercontent.com/14608138/95429723-0f5ecb80-094b-11eb-90a2-f7f523b667ac.png)

**Platform**
Konsole on Manjaro Linux

Would it be viable to add some tests with valid and invalid strings for each regex to catch problems like this in the future?

willmcgugan:
I think it would also accept A-Z and a-g and a few punctuation characters like :;<

Yes, that's clearly a bug. 

Tests would catch false negatives easy enough. But false positives would be impossible to cover, since there is an infinite number of things it shouldn't match.

Any chance of a PR?

deppiedave64:
Sure, I will create a PR later today. :+1:

deppiedave64:
Done. Wasn't at home, so it took a little longer.
Do you think tests for this would be useful? If so I could take a look at it tomorrow.

willmcgugan:
Tests are always welcome.

There probably isn't much point in testing the highlighter system, since that's well covered. But something that individually checks the regexes might be worthwhile.

PR #371, opened 2020-10-09
Add tests for regular expression

deppiedave64:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

I added tests for the regular expressions in the `ReprHighlighter` class, as discussed here: #364

Also, I made a slight change to the attribute regex to also allow values surrounded by single quotes, since I thought that might be desired behavior.

### DIFF
--- a/tests/test_highlighter.py
+++ b/tests/test_highlighter.py
@@ -14,6 +14,47 @@ def test_wrong_type():
 @pytest.mark.parametrize(
     "style_name, test_str",
     [
+        # Braces
+        ("repr.brace", "["),
+        ("repr.brace", "]"),
+        ("repr.brace", "("),
+        ("repr.brace", ")"),
+        ("repr.brace", "{"),
+        ("repr.brace", "}"),
+        # Tags
+        ("repr.tag_start", "<tag>"),
+        ("repr.tag_end", "<tag>"),
+        ("repr.tag_content", "</tag>"),  # closing tag
+        ("repr.tag_content", "<some-tag>"),  # hyphen
+        ("repr.tag_content", "<some.tag>"),  # dot
+        ("repr.tag_content", "<some:tag>"),  # colon
+        # Attributes
+        ("repr.attrib_name", "attrib="),  # no value
+        ("repr.attrib_value", "attrib=value"),  # some value
+        ("repr.attrib_value", 'attrib="value"'),  # value with double quotes
+        ("repr.attrib_value", "attrib='value'"),  # value with single quotes
+        # Boolean values
+        ("repr.bool_true", "True"),
+        ("repr.bool_false", "False"),
+        ("repr.none", "None"),
+        # Numbers
+        ("repr.number", "123456789"),  # positive
+        ("repr.number", "-123456789"),  # negative
+        ("repr.number", "123.456789"),  # decimal
+        ("repr.number", "0x0123af"),  # hexadecimal
+        # Paths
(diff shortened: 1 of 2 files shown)


tests/test_highlighter.py
```diff
@@ -24,11 +65,23 @@ def test_wrong_type():
...
+        ("repr.uuid", "ffffffff-ffff-ffff-ffff-ffffffffffff"),  # highest uuid
     ],
 )
 def test_highlight_regex(style_name: str, test_str: str):
     """Tests for the regular expressions used in ReprHighlighter."""
     text = Text(test_str)
     highlighter = ReprHighlighter()
     highlighter.highlight(text)
-    assert text._spans[-1] == Span(0, len(test_str), style_name)
+    assert style_name in repr(text)
```
Edward-Knight, line 87: I should have probably added a comment to this affect, but originally I was testing the prevailing style (i.e. the last one) was the expected style, and that it covered the span of the entire string.

Otherwise if multiple regexes match then the wrong style could be applied on top. As an example, if the MAC address regexes are listed higher than the number regexes in the `highlights` list in `ReprHighlighter`, the individual numbers will be highlighted instead of the MAC address.

willmcgugan: I'm thinking the test should probably check that there is an appropriate span covering the entire string. Something along these lines:

```python
assert Span(0, len(test_str), style_name) in text.spans
```

Edward-Knight: Perhaps we could add another test specifically for the more complex ones (like MAC address) to ensure it's highlighted as a MAC address rather than a sequence of numbers

deppiedave64: > I should have probably added a comment to this affect, but originally I was testing the prevailing style (i.e. the last one) was the expected style, and that it covered the span of the entire string.

I didn't think about it, but yes, that really is to be considered.
I think checking that there is a span covering the entire string won't work in every case, since there are matches that intentionally do not cover the whole string, like `<tag_start>`, `<tag_name>`, `<tag_contents>` and `<tag_end>`. So probably the best solution would really be to add a separate method for the more complex ones.

willmcgugan: @deppiedave64 good point.

deppiedave64: Maybe dividing it in one method for regexes that are supposed to match the whole string and one for those with multiple matches?

willmcgugan: @deppiedave64 Makes sense.

deppiedave64: All right, I will change that later. :+1:

willmcgugan: I've been looking in to the highlighter regexes in general, and I may have complicated matters. Hopefully for a good reason.

The regexes could overlap quite a bit, so something like an IP address would also be highlighted as two floating point numbers. There are a few other such issues such as the ipv6 regex also matching a valid EUI.

The output looks ok give the styles used, but it's inefficient as it creates pointless spans to process. In master I've rearranged the regexes, and grouped some of the regexes in to a single regex. Which should hopefully mean it doesn't create more spans than it needs to.

I may change it further after a bit of testing, so you might want to hold-off on this PR until I'm confident it's working correctly.

Thanks @deppiedave64 @Edward-Knight

willmcgugan:
This is solved in Master. I made some improvements to the highlighter system. Thanks @deppiedave64 for prompting me to look at this.

Issue #262, opened 2020-09-01
[REQUEST] Arrow Keys and cursor control?


scherrey:
I have a little language interpreter/repl that is begging for integration with rich. What I really need to be able to support, however, is a rapid input history access (via up/down arrows) and support cursor movements on a line of text. 

Is there a method for getting access to individual keystroke input with rich?

My target environment in linux but it would be nice if the feature was platform independent. I'm running Python 3.8. 

Just being able to use up arrow to repeat the last input command and (preferably) being able to edit it before executing would be a godsend in terms of usability improvements. Pretty much something like what bash presently offers.

willmcgugan:
Rich doesn't offer any interactivity yet. It might in the future.

In the meantime, you might want to look at [prompt-toolkit](https://github.com/prompt-toolkit/python-prompt-toolkit).

scherrey:
Thanks I will check it out! Will see how prompt-toolkit and rich can co-exist.

tinchit0:
For anyone interested, I managed to make it work by extending ```rich```'s console, integrating a ```PromptSession``` from ```prompt-toolkit```, and using it to handle calls to built-in ```input()``` function.

```python
from getpass import getpass

from prompt_toolkit import PromptSession
from prompt_toolkit.history import FileHistory
from rich.console import Console


class HistoryConsole(Console):
    def __init__(self, history="session_history", *args, **kwargs):
        self.history = FileHistory(history)
        self.session = PromptSession(history=self.history)
        return super().__init__(*args, **kwargs)

    def input(
        self, prompt="", markup=True, emoji=True, password=False, stream=None
    ) -> str:
        if prompt:
            self.print(prompt, markup=markup, emoji=emoji, end="")
        if password:
            result = getpass("", stream=stream)
        else:
            if stream:
                result = stream.readline()
            else:
                result = self.session.prompt("")
        return result
```

And that works fine:
```python
console = HistoryConsole()
response = Prompt.ask("[cyan]Tell me something[/cyan]", console=console)
```

This allows the prompt to accept up/down arrows for history navigation and even reverse-i-search.

That said, I still believe it would be great if ```rich``` included history support natively, without external dependencies.

PR #382, opened 2020-10-13
Add live updating for any render

nathanrpage97:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

I've been looking at ways to have live updating data table from `rich`. Whenever I used `LiveRender` it always faced any issues if something else was being logged.

So with that in mind I took what was there already from the `Progress` component and made a generalized Live Updating view. It allows for the component to be updated by the user with an `update()` function. It currently isn't threaded like Progress is, but could be expanded maybe with a LiveThreaded?

### Limitations

If the render component is too large, then it will start too not fully clear any of the Renderable above the console view. I have messed around with creating a `safe` flag that will have it print "the terminal is too small" instead when this is true. I didn't want to add it here yet before getting feedback.

### Showcase

rich_live_example

[Some gist examples](https://gist.github.com/nathanrpage97/6eb6b92cc55a7536ad344709bfabe4e9)


willmcgugan:
I think this is a good idea! A number of people have asked for it.

It probably should have an option for thread-based auto refreshing so you could modify your data without having to explicitly refresh it. I'd also need typing, docstrings and tests, but I think you're on the right path. Let me know if you need any pointers.

nathanrpage97:
Great, I'm going to add in the threading next and then clean it up a bit. I think afterwards I can consolidate some of the design decisions this feature contains, that way it can be discussed upon what needs to be done. Then once those get through I can add testing and documentation.

nathanrpage97:
@willmcgugan I'm currently facing an issue trying to check the output of these live Renders for testing in python 3.6. It seems to be related to handling of utf-8 encoding. Do you know of any workarounds?

nathanrpage97:
Ok, so for now I've disabled testing on python3.6 for all tests in test_live.py.

I now think it is at a point for review on how to proceed.
Still needs some detailed examples in `examples/` and cleaning up of examples in rich.live.

rich/live.py
```diff
@@ -0,0 +1,347 @@
...
+        *,
+        console: Console = None,
+        auto_refresh: bool = True,
+        refresh_per_second: float = 4,
+        transient: bool = False,
+        redirect_stdout: bool = True,
+        redirect_stderr: bool = True,
+        hide_overflow: bool = True,
+    ) -> None:
+        """Renders an auto-updating live display of any given renderable.
```
willmcgugan, line 44: For some reason Sphinx won't pick up the docstring from __init__. You need to move it up a level to the class docstring.

rich/live.py
```diff
@@ -0,0 +1,347 @@
...
+        if self.console.is_terminal:
+            if self._redirect_stdout:
+                self._restore_stdout = sys.stdout
+                sys.stdout = _FileProxy(self.console, sys.stdout)
+            if self._redirect_stderr:
+                self._restore_stderr = sys.stderr
+                sys.stdout = _FileProxy(self.console, sys.stdout)
+
+    @property
+    def item(self) -> RenderableType:
```
willmcgugan, line 135: `item` may not be the best name here. I would think an item as in a dictionary items. How about simply "renderable" ?

rich/live.py
```diff
@@ -0,0 +1,347 @@
...
+        """Get the renderable that is being displayed
+
+        Returns:
+            RenderableType: Displayed renderable.
+        """
+        with self._lock:
+            return self._live_render.renderable
+
+    @property
+    def is_started(self) -> bool:
```
willmcgugan, line 145: I suspect this doesn't need to be part of the public interface.

As a context manager, I can't see the developer needing to query if its been started.

rich/live.py
```diff
@@ -0,0 +1,347 @@
...
+                ]
+        elif (
+            not self.is_started and not self.transient
+        ):  # if it is finished render the final output for files or dumb_terminals
+            return [*renderables, self._live_render]
+
+        return renderables
+
+
+if __name__ == "__main__":
```
willmcgugan, line 243: You might want to add a `# pragma: no cover` here to exclude the block from coverage.

tests/test_live.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def check_output(output_file: str, output: str) -> None:
+    output = output.replace("\r", "")
+    if os.getenv("CAPTURE") is not None:  # adjust the correct output check
+        set_capture_text("live", output_file, output=output)
+
+    correct_output = get_capture_text("live", output_file).replace("\r", "")
+
+    if os.getenv("VERBOSE") is not None:
```
willmcgugan, line 44: Is this env some kind of standard? I generally leave print statements in tests, because pyTest will hide the output if the test passes anyway.

tests/test_live.py
```diff
@@ -0,0 +1,265 @@
+import io
```
willmcgugan, line 1: Kudos on being thorough with the tests.

I think its best to simplify these tests though. By that I mean get the coverage of live by executing as little other code as possible. Which probably means using a different renderable other than a table (which is the most complex renderable). A Text instance is probably enough.

I'd also prefer to avoid the text files and have the expected output embedded.

BTW there is a slight lower level way of capturing output, that I find works better for tests:

```python
console.begin_capture()
console.print("Hello")
assert console.end_capture() == "Hello\n"
```

tests/test_live.py
```diff
@@ -0,0 +1,265 @@
...
+        assert live.item == ""
+
+        assert live.is_started
+        live.stop()
+        assert not live.is_started
+
+    assert not live.is_started
+
+
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="requires python3.7 or higher")
```
willmcgugan, line 66: What's the reason for skipping 3.6? Rich supports 3.6.1 or above...

tests/test_live.py
```diff
@@ -0,0 +1,265 @@
...
+import pytest
+from rich.console import Console
+from rich.live import Live
+from rich.table import Table
+from rich.text import Text
+
+from tests.util import get_capture_text, set_capture_text
+
+
+def create_capture_console(*, width: int = 60, height: int = 80) -> Console:
```
willmcgugan, line 17: I'd prefer to avoid external helper functions for tests. I think its best to have the entire test located in a single test function (other than the code being tested obviously).

rich/live.py
```diff
@@ -0,0 +1,347 @@
...
+        self.ipy_widget: Optional[Any] = None
+        self.auto_refresh = auto_refresh
+        self._started: bool = False
+        self.transient = transient
+
+        self._refresh_thread: Optional[_RefreshThread] = None
+        self.refresh_per_second = refresh_per_second
+
+        self.hide_overflow = hide_overflow
+        self._hide_render = LiveRender("Terminal too small\n")
```
willmcgugan, line 75: This is problematic. I doubt the dev will want the output to disappear, and it may not even be possible for the user to enlarge the terminal so the output fits. It's also presumptuous to assume the user can read English.

I see the problem though. I think the only thing that we can do at this level is to crop the output, and document the behaviour.

willmcgugan (changes requested):
Great work. Sorry for the delay in reviewing.

All for this in principle, but I'm not keen on the "terminal too small" warning.

I've also left some notes about tests. I'm very much pro-tests but they require maintenance in the same was other code, and I like them to be as lean as possible.

tests/test_live.py
nathanrpage97: The issue is I could not get the tests to pass on 3.6, I believe it is how python 3.6 handles utf-8 encoding. Perhaps with simpler tests it may not be an issue.

rich/live.py
nathanrpage97: I see, so it should be an optional flag to `crop_overflow`? User may also want the whole renderable to be printed no matter what.

tests/test_live.py
willmcgugan: I'm not aware of any differences in utf-8 encoding. You might want to investigate what's different there.

rich/live.py
willmcgugan: What actually happens when the output is taller than the terminal? If there is a chance the developer may want that behaviour, then maybe a flag would be appropriate.

tests/test_live.py
nathanrpage97: Ok, I will reinvestigate once finished the other changes and making the tests use simpler renderables.

rich/live.py
nathanrpage97: Actually a `vertical_overflow` flag that supports crop, ellipsis, ignore, would make it similar to the typical `overflow` feature.

tests/test_live.py
nathanrpage97: Its not something standard. Just something I thought would be useful when testing locally (turns out it wasn't). Forgot to remove it.

rich/live.py
nathanrpage97: ttps://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autoclass_content Is the reason it only gets it from the class. 

I will move it to the class.

willmcgugan: TIL!

nathanrpage97:
Alright I think it is ready for another review.

Main changes:
- `vertical_overflow` over the _terminal too small_ warning
- Test procedure simplified to simple string display

nathanrpage97:
**Note** there are still issues with fluctuating sized displays. Will have to look into deeper.

willmcgugan:
no worries. Let me know if you need any help with debugging.  Will review soon (busy with the day job).

nathanrpage97:
Ok so with some additional locking in the LiveRender and RefreshThread, it is now stable (and not deleting console log statements non-deterministically)

yihong0618:
Can this pr solve #262 ?

nathanrpage97:
@yihong0618 No, this pr doesn't add any user input handling.

yihong0618:
> @yihong0618 No, this pr doesn't add any user input handling.

Thanks a lot.

willmcgugan:
Sorry for the delay @nathanrpage97! Will look at this soon.

docs/source/live.rst
```diff
@@ -0,0 +1,172 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To avoid breaking the live display visuals, Rich will redirect ``stdout`` and ``stderr`` so that you can use the builtin ``print`` statement.
+This feature is enabled by default, but you can disable by setting ``redirect_stdout`` or ``redirect_stderr`` to ``False``.
+
+
+Examples
+--------
+
+See `table_movie_live.py <https://github.com/willmcgugan/rich/blob/master/examples/table_movie_live.py>`_ and
```
willmcgugan, line 170: I may end up just replacing replacing table_movie.py with table_movie_live. Since it could serve as a demo of table features and the Live class.

rich/live.py
```diff
@@ -0,0 +1,364 @@
...
+            lines = console.render_lines(self.renderable, options, pad=False)
+
+            shape = Segment.get_shape(lines)
+            _, height = shape
+            if height > console.size.height:
+                if self._live.vertical_overflow == "crop":
+                    lines = lines[: console.size.height]
+                    shape = Segment.get_shape(lines)
+                elif self._live.vertical_overflow == "ellipsis":
+                    lines = lines[: (console.size.height - 1)] + [
```
willmcgugan, line 64: Would this put the ellipsis on the left hand side of the terminal? I suspect it would look better centered. Also, what would happen if the terminal is less than three characters in width (crazy I know, just covering all the bases).

willmcgugan:
Excellent work @nathanrpage97 Testing now...

willmcgugan:
Works great. I think my only request would be the ellipsis in the center.

You could render it like this, which would also solve the unlikely scenario of a terminal of 2 characters.

```
from rich.text import Text

yield from console.render(
    Text('...', overflow="crop", justify="center", end="")
)
```

rich/live.py
nathanrpage97: Alright I added in the console render suggestion to cover this.

docs/source/live.rst
nathanrpage97: Updated the table_movie_live.py -> table_movie.py

nathanrpage97:
Alright the changes have been added in for table_movie_live.py -> table_movie.py and centered overflow ellipsis.

willmcgugan:
@nathanrpage97 Do you have a Twitter handle? I'd like you give you a shoutout when I release the next version of Rich.

nathanrpage97:
Awesome, I don’t use Twitter, so a GitHub plug instead?

PR #387, opened 2020-10-15
Change render_group's "fit" kwarg default from False to True

Honno:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

I may be missing something, but I believe it would make sense for `@render_group()` to have the same `fit` kwarg default as it's `RenderGroup` counter-part (i.e. `True`, as opposed to `False`)—this PR simply changes just that.

My experience: I used `RenderGroup` before `@render_group()`, and had gotten used to `fit` defaulting to `True` (without knowing about the measurement mechanism at the time), so I was initially confused as to what was creating huge measurements when I was rendering a nested renderable-like object.


willmcgugan:
@Honno I think you're correct. good spot. Any chance of a test, and fill out the docstring for `render_group` ?

Honno:
I goofed the commit history haha, will fix in a bit.

Honno:
I messed up the docs! On it.

Honno:
I added a coverage test for `render_group` (implicitly also `RenderGroup`), and documented the `fit` kwarg for both `RenderGroup` and `render_group`. <sup>sorry if you got notification spammed!</sup>

I can't get my local docs to update with these changes, so I'll have to figure that out and verify the docs will work (i.e. `/reference/console.html` updates). If you wanna do that, or I'll try it again later.

I also black-formatted `_card_render.py` as that was failing CI. It does not seem to affect the tests passing.

willmcgugan:
LGTM. THanks.

PR #403, opened 2020-10-23
Allow level name of `RichHandler` to be customized

pwwang:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Allow level name of `RichHandler` to be customized

```python
import logging

logger = logging.getLogger('rich')
logger.addHanlder(RichHandler(show_level=lambda level: level[:1]))

logger.info('info')
logger.warning('warning')
```
```
[10/23/20 20:14:18] I info
                    W warning
```

`show_level` argument of `RichHandler` now can receive a function to format or transform the level name of the record.
This function should return a fixed width of all level names.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -7,3 +7,4 @@ The following people have contributed to the development of Rich:
 - [Oleksis Fraga](https://github.com/oleksis)
 - [Hedy Li](https://github.com/hedythedev)
 - [Will McGugan](https://github.com/willmcgugan)
+- [Panwen Wang](https://github.com/pwwang)
--- a/rich/_log_render.py
+++ b/rich/_log_render.py
@@ -13,7 +13,7 @@ class LogRender:
     def __init__(
         self,
         show_time: bool = True,
-        show_level: bool = False,
+        show_level: Union[int, bool] = False,
         show_path: bool = True,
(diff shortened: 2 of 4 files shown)


willmcgugan:
Sorry, I'm not keen on this. I prefer not to overload parameters like that, I find having an argument that may be two quite different types tends to make typing difficult and makes the code harder to reason about.

I understand why you would want this functionality, but trying to pack everyone's use case in to a single class leads to difficult to maintain code.

What I would accept is a change that allows you to specify level names via inheritance. Say by having a class var that stores the level names. So you could create a custom handler class along these lines:

```python
class MyHandler(RichHandler):
    levels = {
        10: "D Debug",
        20: "I Info",
    }
```

pwwang:
What I also care about is the width of the level name column, which is fixed in your code as 8. That makes it look awkward when I only have a couple of records with short names together (there is a big gap between the level name and message). 

If your proposal is acceptable, is it also possible/acceptable to adjust that width to the max length of all available level names (with the one redefined in the handler subclass)?

willmcgugan:
I'd suggest adding a level_width parameter to the LogRender class. You could calculate the maximum width required in the handler.

pwwang:
Make sense. Let me give it a try.

PR #407, opened 2020-10-25
Allow subclassing RichHandler

pwwang:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Based on the discussion in #403 

Now `RichHandler` can be subclassed with two variables `levels` and `level_width` optionally.

`levels` is a dictionary with level numbers as keys and level names as values. One can rename a level with an existing level number, or add a new level number and name.

`level_width` is to define the width in the level column. If not given, the maximum with of all levels (including the ones defined by `level`) will be used. If the level name is longer than `level_width`, it will be cut.

Example:
```python
import logging

class MyHandler(RichHandler):
    levels = { 
        logging.WARNING: 'WARN',
        logging.ERROR: 'ERR',
        logging.DEBUG: 'DBG',
        9: 'FOO'
    }
    # If this is not defined, 8 will be used, since CRITICAL is the longest name
    level_width = 4 

logger = logging.getLogger('rich_level_rename')
logger.addHandler(MyHandler(show_path=False))
logger.setLevel(logging.DEBUG)

logger.info('An info message')
logger.warning('A warning message')
logger.error('An error message')
logger.debug('A debug message')
logger._log(9, 'A custom message', ())
logger.critical('A critical message')
```

image

### DIFF
--- a/rich/_log_render.py
+++ b/rich/_log_render.py
@@ -16,11 +16,13 @@ class LogRender:
         show_level: bool = False,
         show_path: bool = True,
         time_format: str = "[%x %X]",
+        level_width: int = 8,
     ) -> None:
         self.show_time = show_time
         self.show_level = show_level
         self.show_path = show_path
         self.time_format = time_format
+        self.level_width = level_width
         self._last_time: Optional[str] = None
 
     def __call__(
@@ -42,7 +44,7 @@ class LogRender:
         if self.show_time:
             output.add_column(style="log.time")
         if self.show_level:
-            output.add_column(style="log.level", width=8)
+            output.add_column(style="log.level", width=self.level_width)
         output.add_column(ratio=1, style="log.message", overflow="fold")
         if self.show_path and path:
             output.add_column(style="log.path")
--- a/rich/_log_render.py
+++ b/rich/_log_render.py
@@ -10,6 +10,16 @@ if TYPE_CHECKING:
 
 
 class LogRender:
+    """A helper class to render log records.
+
+    Args:
+        show_time (bool): Show a column for the time. Defaults to True.
+        show_level (bool): Show a column for the level. Defaults to True.
+        show_path (bool): Show the path to the original log call. Defaults to True.
+        time_format (str): The time format for the log time. Defaults to "[%x %X]".
+        level_width (int): The width of the level column. Defaults to 8.
+    """
+
     def __init__(
         self,
         show_time: bool = True,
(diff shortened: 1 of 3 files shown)


rich/_log_render.py
```diff
@@ -16,11 +16,13 @@ def __init__(
         show_level: bool = False,
         show_path: bool = True,
         time_format: str = "[%x %X]",
+        level_width: int = 8,
```
willmcgugan, line 19: Would you mind filling in the docstring for this class. I missed that one.

willmcgugan (changes requested):

rich/logging.py
```diff
@@ -50,6 +59,10 @@ class RichHandler(Handler):
     ]
     HIGHLIGHTER_CLASS: ClassVar[Type[Highlighter]] = ReprHighlighter
 
+    levels: ClassVar[Dict[int, str]] = {}
+    level_style: ClassVar[Dict[int, str]] = {}
```
pwwang, line 63: One can also give styles for the names. By default, the handler will use the ones defined in `_level_styles`.

rich/logging.py
```diff
@@ -70,8 +83,19 @@ def __init__(
         super().__init__(level=level)
         self.console = console or get_console()
         self.highlighter = highlighter or self.HIGHLIGHTER_CLASS()
+
+        for level, level_name in self.__class__.levels.items():
+            logging.addLevelName(level, level_name)
+
+        level_width = self.__class__.level_width or max(
+            len(level_name) for level_name in logging._levelToName.values()
```
pwwang, line 91: The use of `logging._levelToName` might be still a problem based on your comment. But I don't find a better way to walk through all level names, including the default ones and the ones defined in `levels` by `logging.addLevelName`.

Looks like `_levelToName` is where the levels and their names stored: https://github.com/python/cpython/blob/c12afa92b0db6f017e479b41f95d75bac7b59850/Lib/logging/__init__.py#L142

Let me know if you have a better way. Or the other idea is that we just walk through the levels defined in the subclass, instead of all available levels. This means all longer names (than `level_width`) should be shortened/renamed in `levels` if one doesn't want them to be cut.

willmcgugan:
Sorry for the delay. I think this is a worthwhile addition and I appreciate the work you put in to this, but I think I'd like to tackle it in a different way. I was unaware that you can set level names in the logging library itself, and another solution may be required.

I'll have a change to master soon.

willmcgugan:
Closing in favour of https://github.com/willmcgugan/rich/pull/419

PR #419, opened 2020-10-31
rich handler refactor

willmcgugan:
Factored out method RichHandler.get_style_and_level to allow overriding in subclasses.


willmcgugan:
@pwwang Let me know if this solved your use-case re level column in the logging handler.

pwwang:
Thanks for implementing it. I think this makes it better and gives it a more flexible way to do it.
I should work on my case theoretically, but I need to wait for you to release it to try.
Haven't seen you add any tests for this? But it's fine if you are confident to merge.

PR #420, opened 2020-10-31
WIP: New spinner code

### LINKED ISSUE #401: add support for Spinner
Indicator for long running task's that don't have iterator. 

Something like https://pypi.org/project/yaspin/

juniorRubyist:
## Type of changes

- [ ] Bug fix
- [x] New feature **(WIP)**
- [ ] Documentation / docstrings **(WIP)**
- [ ] Tests
- [ ] Other

## Checklist

## Description

Closes #401. 

Implements new `Spinner` class for generating an indeterminate loading signal. It can be used manually with a `start()`/`stop()` syntax, or it can also be used as a context.

### Start/Stop Example

```python
spinner = Spinner()
spinner.start()
# do stuff
spinner.stop()
```

### Context Example

```python
from rich.spinner import Spinner
with Spinner(message="Loading...") as s:
    # long running task
    s.message = "Almost done..."
    # another long task
```

### TODO:
- [ ] Documentation
- [ ] Colors
- [ ] More Unicode spinners
- [ ] Property update methods
- [ ] Tests


willmcgugan:
This is a fine solution for a stand-alone spinner. But it doesn't really make use of Rich. There's no way to use Rich's [console markup](https://rich.readthedocs.io/en/latest/markup.html) for example.

I think it should be implemented as part of the Progress class, which gives you the automatic update, and a captures stdout so it won't break the visuals.

rich/spinner.py
```diff
@@ -0,0 +1,47 @@
...
+            sleep(self._delay)
+
+    def start(self):
+        self._run = True
+        self._thread = Thread(target=self.loop)
+        self._thread.start()
+
+    def stop(self):
+        self._run = False
+        while self._thread.is_alive():
```
willmcgugan, line 31: Because there is nothing blocking here, this loop will use up 100% cpu while waiting for the tread to close. You can use `thread.join()` which does the same thing but won't hog the processor.

rich/spinner.py
```diff
@@ -0,0 +1,47 @@
...
+from threading import Thread
+
+RECALL = "\033[2K\r"
+
+
+class Spinner:
+    SLASHES = "|/-\\"
+
+    def loop(self):
+        i = 0
```
willmcgugan, line 11: I always avoid single letter variables. Better to think of a descriptive name.

rich/spinner.py
```diff
@@ -0,0 +1,47 @@
...
+                    icon=self._icons[i % len(self._icons)],
+                    message=self.message,
+                ),
+                end="",
+            )
+            i += 1
+            sleep(self._delay)
+
+    def start(self):
+        self._run = True
```
willmcgugan, line 25: Better to use threading primitives as a way of communicating across threads. Have a look in the docs for the Event class.

willmcgugan (commented):
Threads are tricky.

frostming:
From a rough review, I don't think this implementation is thread-safe. People may want multiple spinners to run together, like `Progress`. Refreshing the output with `RECALL` will likely mess up the console output.

After all, I am glad to see this will be landed in `rich`.

PR #421, opened 2020-10-31
Make TimeRemainingColumn in progress.track optional 

Senpos:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

As a convenience function, `rich.progress.track` allows to create progress bars easily with great defaults, however, I often find myself working with tasks where we can't really estimate the time remaining, so I wish I could have an opportunity to disable `TimeRemainingColumn` to not confuse users.

**Yes**, I can always build a `rich.progress.Progress` object myself with all the columns I need, but IMO it is pretty common use-case so it is worth it to add an additional parameter like `show_remaining_time` to the `rich.progress.track` function. :)

As it should be backwards compatible, the default value is `True` (time remaining will be shown) and the parameter is added to the very end, because it is not kwargs-only one, so we don't want to break code where people pass parameters positionally.

Please, let me know if you are interested in this feature or if I need to do anything extra to make it useful.

I have added this parameter the docstring and checked that all the tests pass.

I also ran `make format-check`, it found a problem with `rich/tests/test_pretty.py`, but I did not even touch it, so it might be a glitch. 😅

If you need this functionality to have an extra test, please, let me know. I will also need some guidance on how to do that.
I believe, I might need to duplicate+modify `test_track` test and call it `test_track_without_time_remaining`.

P. S. Thanks for creating the library, it is so convenient to use. 🎉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added tracebacks_show_locals parameter to RichHandler
 - Applied dim=True to indent guide styles
 - Added max_string to Pretty
+- Added show_remaining_time parameter to progress.track
 
 ## [9.1.0] - 2020-10-23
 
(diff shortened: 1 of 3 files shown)


willmcgugan:
I'm not keen on this, tbh.

I can see why it's useful, but I suspect somebody will ask for a `show_description` and `show_bar` etc. And there are already a fair number of configuration options on that method.

I think its best I draw a line under the parameters on that function given that you can build a Progress instance with your requirements for columns (like you said).

I'll leave this open for now, but I'll listen if you have any counter-points!

PR #423, opened 2020-11-01
Strip ANSI escapes when producing HTML

ssbarnea:
In order to avoid seeing ANSI garbage produced by other tools, we now strip ANSI escapes inside html_report(), as this will mare the output more readable.

## Type of changes

- [x] Bug fix

## Checklist

## Description


willmcgugan:
This won't be necessary in 9.2.0. If you have a look at the FileProxy class in Progress, it now decodes ANSI styles.

PR #426, opened 2020-11-02
doc: Fix styles=True typo for pager

The-Compiler:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Description

According to the code (and reference), the argument is "styles=True" rather than "style=True".


The-Compiler:
CI fails because Black would reformat /home/runner/work/rich/rich/tests/test_pretty.py - I don't think that's related to my changes.

willmcgugan:
Thanks.

PR #455, opened 2020-11-29
Use the proper error message with caused exceptions

nneonneo:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

When an exception is raised using "from", the chained exception goes into
__cause__, which results in a different message using Python's native traceback
handler. This patch properly replicates that error message.


willmcgugan:
Thanks!

PR #461, opened 2020-12-02
Changes typing for logging level to accept both str and int

heavelock:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

- [x ] I accept that @willmcgugan may be pedantic in the code review.

## Description

The standard logging module accepts either str or int as a logging level. str is internally casted to int. Rich does not do anything with the logging level so there is no reason to limit the type here.

### DIFF
--- a/rich/logging.py
+++ b/rich/logging.py
@@ -21,7 +21,7 @@ class RichHandler(Handler):
         under your control. If a dependency writes messages containing square brackets, it may not produce the intended output.
 
     Args:
-        level (int, optional): Log level. Defaults to logging.NOTSET.
+        level (Union[int, str], optional): Log level. Defaults to logging.NOTSET.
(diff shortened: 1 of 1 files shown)


rich/logging.py
```diff
@@ -55,7 +55,7 @@ class RichHandler(Handler):
 
     def __init__(
         self,
-        level: int = logging.NOTSET,
+        level: Union[int, str] = logging.NOTSET,
```
willmcgugan, line 58: Can you update the docstring above please.

willmcgugan (changes requested):

heavelock:
Black complains in CI at file I did not touch and when I am trying to run it on my system it does not complain.

willmcgugan:
That's my fault. I'll fix that. Thanks for the PR.

PR #500, opened 2020-12-15
Add example notebooks

donno2048:
No change to any of the code, only Updated _examples_ folder

## Description

The exact same examples but in Jupyter's format


fomightez:
I think the most impressive one is missing? The code from `table_movie.py` works as is in the notebook, and it is nice to see the animation running.

Also, I'm not sure for some of the simpler ones if using `console.print()` and not simply the `print()` imported from rich is warranted? I'm not seeing the dynamic adjusting in the notebook context with `console.print()` in regards to the width as discussed at the end of [this section](https://github.com/willmcgugan/rich#tables) in the documentation. So maybe it is misleading?

willmcgugan:
> Also, I'm not sure for some of the simpler ones if using console.print() and not simply the print() imported from rich is warranted? I'm not seeing the dynamic adjusting in the notebook context with console.print() in regards to the width as discussed at the end of this section in the documentation. So maybe it is misleading?

`console.print` has additional features over `rich.print` as it doesn't need to conform to the builtin print.

Rich can't detect the width of the 'terminal' when running in Jupyter, so it's fixed.

I've been delaying merging this as it feels unnecessary when you can do it in the terminal. But I guess for some Jupyter is the primary interface to Python.

Issue #471, opened 2020-12-07
[REQUEST] pretty.install does not work with IPython InteractiveShellEmbed.


amal-khailtash:
Have you checked the issues for a similar suggestions?  I did not find any

**How would you improve Rich?**

pretty.install is great and modifies sys.displayhook, but it does not seem to affect IPython's interactive shell for it has its own displayhook (RichPromptDisplayHook).

**What problem does it solved for you?**

It provides a much nicer printout for data structures.

What problem do you have that this feature would solve? I may be able to suggest an existing way of solving it.

More readable IPython REPL.  The following example does not seem to change IPython displayhook.

```
from IPython.terminal.embed import InteractiveShellEmbed

ipshell = InteractiveShellEmbed(user_ns=globals())
pretty.install(indent_guides=True)
print(sys.displayhook)
print(ipshell.displayhook)
ipshell()
```

Output:

```
<function install.<locals>.display_hook at 0x7f9cbcdcaca0>
<IPython.terminal.prompts.RichPromptDisplayHook object at 0x7f9cbd6021c0>
```

canbooo:
Bump as this would also make Spyder users happy.

willmcgugan:
Works in 9.6.0 Thanks to @nathanrpage97

PR #856, opened 2020-12-28
Install pretty on ipython

nathanrpage97:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

#471  describes the issue of injecting the display hook into ipython. 

The issue is that the sys.displayhook cannot be changed inside of the ipython repl. Instead, I've added the display_hook into the [ipython plain text formatter](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.formatters.html).


rich/pretty.py
```diff
@@ -63,7 +63,7 @@ def install(
 
     console = console or get_console()
 
-    def display_hook(value: Any) -> None:
+    def display_hook(value: Any, *_args, **_kwargs) -> None:
```
willmcgugan, line 66: Is this for iPython's benefit? sys.displayhook doesn't take any additional argument according to docs.

rich/pretty.py
```diff
@@ -83,7 +83,16 @@ def display_hook(value: Any) -> None:
...
-    sys.displayhook = display_hook
+    try:  # pragma: no cover
+        ip = get_ipython()  # type: ignore
+        from IPython.core.formatters import BaseFormatter
+
+        # replace plain text formatter with rich formatter
+        formatter = BaseFormatter()
+        ip.display_formatter.formatters["text/plain"] = formatter
+        formatter.for_type(object, func=display_hook)
+    except:
```
willmcgugan, line 94: This should probably be `except Exception` so as not to catch any system exceptions.

willmcgugan (commented):
This is great. iPython support is a common request.

willmcgugan:
@nathanrpage97 There is an unfortunate glitch where iPython print "In [X]:" and Rich expects the cursor to be at the start of the line. This can break rendering of the first line. It's really obvious for something like Panel, but occurs in other places.

Screen Shot 2020-12-28 at 17 35 42

The default pretty printing in iPython will insert a new line when printing containers like dicts or lists. I'm thinking that Rich should do the same, and also print a new line for Rich renderables.

I think the logic should be something like the following:

```python
requires_newline = (ConsoleRenderable, abc.Mapping, abc.Sequence, abc.Set)
if isinstance(value, requires_newline):
    console.line()
```

What do you think?

rich/pretty.py
nathanrpage97: It was there to allow it to work with the ipython display hook, but is no longer needed

nathanrpage97:
Alright, I've added in the change you suggested for the new line. Still needs testing in a few environments including jupyter.

nathanrpage97:
Note there seems to be a recursion error issue when running it in jupyter.

nathanrpage97:
Alright, I've tweaked it to make sure rich will play well with terminal ipython and rich display jupyter.

willmcgugan:
Great, thanks.

PR #868, opened 2020-12-30
add progress component that works cleanly with Live

nathanrpage97:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Creation of a `ProgressComponent` that extends Progress to work nicely with `Live`. As detailed in #482.

Renaming the class may be useful as it feels a bit lengthy.

### DIFF
--- a/docs/source/progress.rst
+++ b/docs/source/progress.rst
@@ -52,6 +52,30 @@ Here's a simple example::
 The ``total`` value associated with a task is the number of steps that must be completed for the progress to reach 100%. A *step* in this context is whatever makes sense for your application; it could be number of bytes of a file read, or number of images processed, etc.
 
 
+Combining with Live
+~~~~~~~~~~~~~~~~~~~
+
+If you would like to use a progress bar with the :class:`~rich.live.Live` to create complex updating dashboards, you should instead use the :class:`~rich.progress.ProgressComponent`.
+
+
+Here's a simple example::
+
+    import time
+
+    from rich.console import RenderGroup
+    from rich.live import Live
+    from rich.progress import ProgressComponent
+    from rich.table import Table
+
+    table = Table("1", "2", "3")
+    progress = ProgressComponent()
+    display = RenderGroup(table, progress)
+
+    with Live(display) as live:
+        for step in progress.track(range(10), description="Updating"):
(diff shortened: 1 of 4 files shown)


nathanrpage97:
Ok I think this is now at a reasonable setup, where it is convenient to call by the user.

ksj22000 (approved):

agravier:
You may want to add a note in the Live documentation page as well, liking to the _Combining with Live_ section in the Progress page.

willmcgugan:
Hi @nathanrpage97 

I see the need for this, bit I'm not keen on this `RenderProgress` class with stubs for `start()` etc.

Could we factor out a `ProgressRenderable` that handles the state and rendering, and extend that with the refresh thread etc to create the Progress class.

In psuedo-code...

```python
class ProgressRenderable:
    def add_task(self, ...):
        ...
    def __rich_console__(self,...):
        ...

class Progress(ProgressRenderable):
    def start(self):
        ...
```

Essentially adding functionality to the renderable to create a Progress class, rather than subtracting from Progress to create a renderable -- if that makes sense!

nathanrpage97:
I was going to do it that way, the issue I had is that it would imply to a users IDE that the `Progress` component is `RichRenderable ` when it is not. Even if you were to add a `NotImplemented` exception to the `__rich_console__` of `Progress` in this case. Thoughts?

willmcgugan:
> I was going to do it that way, the issue I had is that it would imply to a users IDE that the Progress component is RichRenderable when it is not.

I think it would be Rich renderable. It wouldn't be the recommended way to use Progress, but I don't see any harm in being able to do `console.print(progress)`.

nathanrpage97:
Alright sounds good, I will refactor it tonight. I still want to keep the static `renderable()` method on `Progress` to keep it as 1 import that users will realistically  need from `rich.progress`.

willmcgugan:
@nathanrpage97 You might want to hold off. I'd like to refactor Progress so that it uses the functionality in Live. I'm also thinking about ways to supported nested Live / Progress.

nathanrpage97:
Alright all good. That makes a lot of sense. Should this pr be closed?

PR #887, opened 2021-01-07
make in the line 33 should be `make` .

Joe-Sin7h:
Just like other Module names make should also be in tick marks like this `make`.
It can be a little confusing for newbies.

## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [X] Other

## Checklist

## Description

Just like the other module names make also should be inside tick marks like this `make`


willmcgugan:
:+1:

PR #902, opened 2021-01-09
allow stdin on rich.markdown / rich.syntax

amancevice:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

If user provides no PATH or uses "-" as PATH, markdown is read from sys.stdin.

Examples:

```bash
python -m rich.markdown < README.md
cat README.md | python -m rich.markdown
```

### DIFF
--- a/rich/syntax.py
+++ b/rich/syntax.py
@@ -521,11 +521,17 @@ class Syntax(JupyterMixin):
 if __name__ == "__main__":  # pragma: no cover
 
     import argparse
+    import sys
 
     parser = argparse.ArgumentParser(
         description="Render syntax to the console with Rich"
     )
-    parser.add_argument("path", metavar="PATH", help="path to file")
+    parser.add_argument(
+        "path",
+        metavar="PATH",
+        nargs="?",
+        help="path to file",
+    )
     parser.add_argument(
         "-c",
         "--force-color",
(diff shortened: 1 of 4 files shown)


willmcgugan:
Good idea. Any chance you can also do this for rich.syntax ?

amancevice:
Happy to, will update PR later tonight

amancevice:
I had to add a lexer option since the script can't deduce the file type from the stream, but should work!

amancevice:
It looks like the CI/CD job got hung up on something. Could it be rerun?

rich/syntax.py
```diff
@@ -286,30 +288,34 @@ def from_path(
...
             background_color (str, optional): Optional background color, or None to use theme color. Defaults to None.
             indent_guides (bool, optional): Show indent guides. Defaults to False.
+            lexer_name (str, optional): Lexer name. Defaults to None.
 
         Returns:
             [Syntax]: A Syntax object that may be printed to the console
         """
-        with open(path, "rt", encoding=encoding) as code_file:
-            code = code_file.read()
+        if not path or path == "-":
```
willmcgugan, line 296: Not keen on this I'm afraid. The function is called `from_path` but this change means it might not use the path.

You can create a Syntax object from the regular constructor, which takes the `code` plus `lexer_name`.

willmcgugan (changes requested):
Just one request.

Don't worry about that GH Actions fail. For some reason the MacOS build fails randomly, I still need to look in to that.

rich/syntax.py
amancevice: you're right, I dunno what I was thinking there! 

Anyway, I just pushed a change that's far less invasive and should work just as well.

willmcgugan:
THanks

amancevice:
By the way, I submitted a Homebrew formula called `richmd` that takes advantage of this PR to allow people to execute `richmd` like `cat`. Thanks for creating such useful software!

PR #906, opened 2021-01-11
Use `cls` instead of class name in `Traceback.from_exception()`

jack1142:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

Nothing too special, just changes `Traceback.from_exception()` to not rely on the specific class name and rather refer to `cls`. I'm guessing this probably wasn't meant to be subclassed, but it doesn't hurt to change this.

Haven't added any tests, but I could add a test looking somewhat like this if you think it makes sense to do so:
```py
class SubclassTraceback(Traceback):
    pass

try:
    1 / 0
except Exception:
    assert isinstance(SubclassTraceback.from_exception(*sys.exc_info())
```

### DIFF
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -222,7 +222,7 @@ class Traceback:
         rich_traceback = cls.extract(
             exc_type, exc_value, traceback, show_locals=show_locals
         )
-        return Traceback(
+        return cls(
             rich_traceback,
             width=width,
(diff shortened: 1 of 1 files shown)


willmcgugan:
Seems reasonable 👍

Issue #913, opened 2021-01-13
[BUG] Tracebacks don't seem to work in JupyterLab


henryiii:
**Read the docs**

There doesn't seem to be an indicator that this wouldn't work.

**Describe the bug**

The traceback handler doesn't seem to do anything in IPython or xeus-python in Jupyter.

**To Reproduce**

```
import rich.traceback
rich.traceback.install(show_locals=True)

1 / 0
```

I feel like this might have worked in the recent past? At least I know I tried it and I _thought_ I was in Jupyter, since I like to play around there. But it's not working now, certainly. It works in the regular REPL.

**Platform**
macOS, JupyterLab 3, rich 9.8.0

**Diagnose**

```
╭────────────────────────────────────────────────────────────────────────── <class 'rich.console.Console'> ───────────────────────────────────────────────────────────────────────────╮
│ A high level console interface.                                                                                                                                                     │
│                                                                                                                                                                                     │
│ ╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮ │
│ │ <console width=183 ColorSystem.TRUECOLOR>                                                                                                                                       │ │
│ ╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                                                                                                     │
│     color_system = 'truecolor'                                                                                                                                                      │
│         encoding = 'utf-8'                                                                                                                                                          │
│             file = <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>                                                                                                    │
│ is_dumb_terminal = False                                                                                                                                                            │
│       is_jupyter = False                                                                                                                                                            │
│      is_terminal = True                                                                                                                                                             │
│   legacy_windows = False                                                                                                                                                            │
│         no_color = False                                                                                                                                                            │
│          options = ConsoleOptions(legacy_windows=False, min_width=1, max_width=183, is_terminal=True, encoding='utf-8', justify=None, overflow=None, no_wrap=False, highlight=None) │
│           record = False                                                                                                                                                            │
│         safe_box = True                                                                                                                                                             │
│             size = ConsoleDimensions(width=183, height=73)                                                                                                                          │
│        soft_wrap = False                                                                                                                                                            │
│           stderr = False                                                                                                                                                            │
│            style = None                                                                                                                                                             │
│         tab_size = 8                                                                                                                                                                │
[...]

PR #994, opened 2021-02-07
Add traceback feature for IPython

nathanrpage97:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

#913 describes adding traceback for ipython terminals


rich/traceback.py
```diff
@@ -88,9 +88,55 @@ def excepthook(
...
+        ip.showsyntaxerror = lambda *args, **kwargs: ipy_display_traceback(
+            *args, is_syntax=True, **kwargs
+        )
+
+    try:  # pragma: no cover
+        # if wihin ipython, use customized traceback
+        ip = get_ipython()  # type: ignore
+        ipy_excepthook_closure(ip)
+        return sys.excepthook
+    except Exception:
```
willmcgugan, line 135: Is there a more explicit exception?

rich/traceback.py
```diff
@@ -88,9 +88,55 @@ def excepthook(
...
+        ip._showtraceback = ipy_display_traceback
+        # add wrapper to capture tb_data
+        ip.showtraceback = ipy_show_traceback
+        ip.showsyntaxerror = lambda *args, **kwargs: ipy_display_traceback(
+            *args, is_syntax=True, **kwargs
+        )
+
+    try:  # pragma: no cover
+        # if wihin ipython, use customized traceback
+        ip = get_ipython()  # type: ignore
```
willmcgugan, line 132: I'm guessing Mypy complains because `get_ipython` is injected in to globals or something?

willmcgugan (approved):
Great. I've been asked a few times for this feature.

willmcgugan:
Does this work in Jupyter as well as IPython ?

rich/traceback.py
nathanrpage97: Yep

rich/traceback.py
nathanrpage97: It’s the same method on pretty install. I have it like this so that it will silently fail in case of some breaking issues with ipython.

nathanrpage97:
> Does this work in Jupyter as well as IPython ?

I did some basic checks and it looks the same.

willmcgugan:
Thanks.

PR #998, opened 2021-02-09
Allow timestamps to include milliseconds

finnhughes:
Displaying timestamps requires overriding Console or RichHandler properties:
- time_format
- time_msecs_format

## Type of changes

- [ ] Bug fix
- [*] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [*] I've run the latest [black](https://github.com/psf/black) with default args on new code.
- [*] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate.
- [*] I've added tests for new code.
- [*] I accept that @willmcgugan may be pedantic in the code review.

## Description

This isn't quite what you proposed in https://github.com/willmcgugan/rich/discussions/984 so I'm not necessarily expecting you to accept it ;)

Handle milliseconds in Rich but only able to display them through overriding Console or RichHandler to avoid additional config and to maintain current behaviour.


finnhughes:
I didn't touch live_render.py so I didn't let black touch it. Should a let black format it and update my PR?

willmcgugan:
The black errors are my fault. I think I pushed to master without formatting. Feel free to run black on those files.

finnhughes:
Updated to make black happy but can't see how to retrigger the tests?

finnhughes:
Updated the PR largely as per https://github.com/willmcgugan/rich/discussions/984 but I made the callable return `str` rather than `Text` this means _log_render.py still handles colouring the timestamp and eliminating duplicates.

I'm not sure why the test failure, it's not related to my code and type checks worked fine under python 3.9 on my system?

rich/_log_render.py
```diff
@@ -15,7 +15,7 @@ def __init__(
         show_time: bool = True,
         show_level: bool = False,
         show_path: bool = True,
-        time_format: str = "[%x %X]",
+        time_format: Union[str, Callable[[datetime], str]] = "[%x %X]",
```
willmcgugan, line 18: The return type could be `Text` here, for flexibility.

Also, would you mind creating a type at module level, something like this:

```python
TimeFormatterCallable = Callable[[datetime], Text]
```

and then this param may be type as `Union[str, TimeFormatterCallable]`. Doesn't change anything, but I prefer it for readability sake

rich/_log_render.py
```diff
@@ -51,7 +51,12 @@ def __call__(
         row: List["RenderableType"] = []
         if self.show_time:
             log_time = log_time or console.get_datetime()
-            log_time_display = log_time.strftime(time_format or self.time_format)
+            if time_format:
```
willmcgugan, line 54: The `time_format` parameter should probably also take a callable, for sake of consistency. Then the logic can be simplified a little.

```python
time_format = time_format or self.time_format
if isinstance(time_format, str):
    log_time_display = Text(log_time.strftime(time_format))
else:
    log_time_display = time_format(log_time)
```

tests/test_console.py
```diff
@@ -4,6 +4,7 @@
 import sys
 import tempfile
 from typing import Optional
+from attr import attributes
```
willmcgugan, line 7: Don't think this is needed...

tests/test_console.py
```diff
@@ -119,6 +120,18 @@ def test_log():
     assert result == expected
 
 
+def test_log_milliseconds():
```
willmcgugan, line 123: 👍

willmcgugan (changes requested):

willmcgugan:
I think this is the way forward. 👍 

The log render should be ok using Text rather than str. If not, it should be an easy fix.

You may have to merge master again, the typing issues were unrelated to this PR.

rich/_log_render.py
finnhughes: I can change the typing of the time_format argument to the __call__ function to allow a callable but since time_format is coming directly from the logging packages Formatter class and is typed as `datefmt: Optional[str]` then we'll never actually get a callable.

I agree the conditional is cleaner and wasn't a big fan of my three way conditional so do you want to proceed with changing the typing and then not worry that the theoretically callable type will never be used?

rich/_log_render.py
finnhughes: I like the use of the type. If I pass in a text I get an exception:
```
  File "/Users/finn/src/finn/rich/rich/logging.py", line 141, in emit
    log_renderable = self.render(
  File "/Users/finn/src/finn/rich/rich/logging.py", line 187, in render
    log_renderable = self._log_render(
  File "/Users/finn/src/finn/rich/rich/_log_render.py", line 64, in __call__
    row.append(Text(log_time_display))
  File "/Users/finn/src/finn/rich/rich/text.py", line 139, in __init__
    self._text = [strip_control_codes(text)]
  File "/Users/finn/src/finn/rich/rich/control.py", line 47, in strip_control_codes
    return text.translate(_translate_table)
AttributeError: 'Text' object has no attribute 'translate'
```

finnhughes:
Many thanks Will, that looks loads better 👍

Not sure why it's complaining about progress.py given I didn't touch it but at least everything else looks happy now.

rich/_log_render.py
```diff
@@ -8,29 +8,31 @@
...
-        self._last_time: Optional[str] = None
+        self._last_time: Optional[Text] = None
 
     def __call__(
         self,
         console: "Console",
         renderables: Iterable["ConsoleRenderable"],
         log_time: datetime = None,
-        time_format: str = None,
+        time_format: Union[str, TimeFormatterCallable] = None,
```
finnhughes, line 35: I changed the type but since `time_format` is coming directly from the `logging` package's `Formatter` class, and is typed as `datefmt: Optional[str]` then we'll never actually get a callable?

finnhughes:
I pulled in the bumps to sphinx and pygments but suspect that's unrelated to coverage now passing and that's just a flaky test?

rich/_log_render.py
```diff
@@ -8,29 +8,31 @@
     from .console import Console, ConsoleRenderable, RenderableType
     from .table import Table
 
+TimeFormatterCallable = Callable[[datetime], Text]
```
willmcgugan, line 11: Could we call it `FormatTimeCallable`? I like to start callables with a verb.

rich/_log_render.py
willmcgugan: It can also come from the Console.log method, which also uses LogRender

willmcgugan (approved):

willmcgugan:
Not sure about the test, but its unrelated to your code.

This looks good. I'm working on a feature branch atm, I'll merge this after and it should be in the next release.

Thanks!

rich/_log_render.py
finnhughes: Sounds good, I'll rename and push

PR #1065, opened 2021-02-26
Test prompt confirm with choices

neutrinoceros:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

- [x] I accept that @willmcgugan may be pedantic in the code review. (<3)

## Description

Currently `rich.prompt.Confirm.ask` prints `Please enter Y or N` even when choices are setup, resulting is this absurd behaviour
Screenshot 2021-02-26 at 09 54 56

My change adresses this, _but_ writing this I realised it doesn't solve the deeper issue : this method accepts a `choices` kwargs, which only makes sense in case where the value passes is a sensible mapping for `[True, False]`.

It could check that `len(choices)==2` for instance, but there's no reliable way to check that the order makes sense.
Because it's a just a convenience wrapper, I think the best option would instead be to remove `choices` from 
`rich.prompt.Confirm.__init__` signature. WDYT ?

### DIFF
--- a/rich/prompt.py
+++ b/rich/prompt.py
@@ -329,8 +329,30 @@ class Confirm(PromptBase[bool]):
     """
 
     response_type = bool
-    validate_error_message = "[prompt.invalid]Please enter Y or N"
     choices = ["y", "n"]
+
+    def __init__(
+        self,
+        prompt: TextType = "",
+        *,
+        console: Console = None,
+        password: bool = False,
+        choices: List[str] = None,
+        show_default: bool = True,
+        show_choices: bool = True,
+    ):
+        super().__init__(
+            prompt=prompt,
+            console=console,
+            password=password,
+            choices=choices,
+            show_default=show_default,
+            show_choices=show_choices,
+        )
+
+        self.validate_error_message = "[prompt.invalid]Please enter " + " or ".join(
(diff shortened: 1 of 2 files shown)


willmcgugan:
The Prompt classes were intended to be extended for things like this. For instance your YES/NO prompt could be implemented as follows:

```python
class VerboseConfirm(Confirm):
    choices = ["yes", "no"]
    validate_error_message = "[prompt.invalid]Please enter YES or NO"

VerboseConfirm.ask("continue?")
```

You have a reasonable point about including the choices in the error message, but for a confirmation prompt it would be rare to have to write it more than once.

neutrinoceros:
> You have a reasonable point about including the choices in the error message, but for a confirmation prompt it would be rare to have to write it more than once.

True. I'm happy to document this in docstrings and/or docfiles, and as add a size check for the choices kwarg to make it even clearer. Just give me your go !

neutrinoceros:
@willmcgugan you can close this if you don't think it fits rich's model, but if this needs adjustments I'm ready 😄

willmcgugan:
Thanks @neutrinoceros Closing this for now. I think I just need to expand on this use-case in the docs.

PR #1085, opened 2021-03-02
Fix google colab

davidt0x:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

`rich.console._is_juptyer` was returning false when running in Google Colab's jupyter notebook. It seems to have its own IPython Shell class or something. If you force juptyer things seem to work. I thought this should be the default when doing `from rich.juptyer import print`. Its a simple change that detects if we are running in colab specifically. Not sure the best way to write a test for this so I didn't add any new tests. Here is a working example:

[rich_test.ipynb - Colab Example](https://colab.research.google.com/drive/1Ks3mBg_1wFLYeXQSFBfbNuoj9v8NHDKj?usp=sharing)

Anyway, not sure if you want this to be the default behavior but if you do this should handle it. Also, thanks for the great library.

Dave

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [9.12.5] - 2021-03-02
+
+### Fixed
+
+- Fixed issue where juptyer was not being detected in Google Colab
+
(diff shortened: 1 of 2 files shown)


davidt0x:
I think fixed (ignored) the mypy error.

willmcgugan:
Seems reasonable. I'll include this in the next release.

I wish there was a better way for Jupyter-like implementations to advertise their capabilties.

willmcgugan:
Thanks. Closing this, but the fix is in another branch. New release today.

davidt0x:
Thanks for the great library.

PR #1107, opened 2021-03-11
Support extras in logging handler

traviscook21:
When a user passes extras into a log statement, those extras
don't show up in the rich logging handler.

```
log.info("Message", extra={"foo": "bar"})
```

Now, there's a conditional option `include_extra` which
will tell the `RichHandler` to include the extras on the
next lines of the message as a `Pretty` object

## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -4,9 +4,11 @@ The following people have contributed to the development of Rich:
 
 <!-- Add your name below, sort alphabetically by surname. Link to Github profile / your home page. -->
 
+- [Travis Cook](https://github.com/traviscook21)
 - [Oleksis Fraga](https://github.com/oleksis)
 - [Finn Hughes](https://github.com/finnhughes)
 - [Hedy Li](https://github.com/hedythedev)
 - [Alexander Mancevice](https://github.com/amancevice)
 - [Will McGugan](https://github.com/willmcgugan)
 - [Nathan Page](https://github.com/nathanrpage97)
+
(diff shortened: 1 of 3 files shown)


willmcgugan:
This is not something that the extra dict is typically used for, AFAIK. But I can see why you would want to pass in something to pretty print.

I'm not keen on the filtering out reserved words. What I would accept is a single key that accepts a Rich renderable. There's less change of a key overlapping with something in the futures. Perhaps something like this:

```python
log.debug("Result", extra={"rich.print": results_table})
```

traviscook21:
when building an application that writes to logstash, there's a couple of libraries out there that support logging to JSON and use the `extra` arg to store additional keys outside the scope of `message` that can be indexed in logstash.

this is one of the more popular ones: https://github.com/madzak/python-json-logger

when doing local development, reading JSON is really challenging, so it would be ideal to use rich's logging handler, but it's missing the valuable `extra` args.  I'm filtering out the reserved words because they're not actually a part of `extra`, but the `LogRecord` object doesn't differentiate between standard attributes and the extra one.  Presumably, if someone wanted to include some of the reserved attributes, they'd use the a log `fmt` that was more descriptive than just `%(message)s`.

If all of the above isn't convincing though, would you be open to a `include / exclude` list of `LogRecord` attributes to include?

willmcgugan:
Setting the entire `extra` to your JSON is a bad idea precisely because of this mushing of namespaces together. There's no way to know wether your keys were part of the JSON or intended as part of the log message. 

Logging is way to dynamic to simply keep a include / exclude list. You may not have any control about what other handlers are in use when your code is in production.

The only way to be confident your JSON keys don't trample on something else would be to pick one key and put your data in that. It's slightly more verbose, but hardly that bad.

```python
log.debug("Result", extras={"print", my_json})
```

PR #1108, opened 2021-03-13
add option to not omit identical time values in log renderer

efenka:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

I really like the RichHandler for logging, but I find it not so nice that it omits time values if the line above has the same time value. The Main reason is, that I often send snippets of log messages around and then the time is missing because everything happened within a second. My change is a `repeat_same_time=True` switch to RichHandler (and LogRender) to turn off this behavior. 

``` python3
import logging

logger = logging.getLogger()
logger.addHandler(RichHandler(repeat_same_time=True))

logger.error("A")
logger.error("B")
```

Before (or with `repeat_same_time=False`):
```
[03/13/21 13:35:55] ERROR    A
                    ERROR    B
```

After:
```
[03/13/21 13:36:08] ERROR    A
[03/13/21 13:36:08] ERROR    B

### DIFF
--- a/rich/_log_render.py
+++ b/rich/_log_render.py
@@ -18,12 +18,14 @@ class LogRender:
         show_level: bool = False,
         show_path: bool = True,
         time_format: Union[str, FormatTimeCallable] = "[%x %X]",
+        repeat_same_time: bool = False,
         level_width: Optional[int] = 8,
     ) -> None:
         self.show_time = show_time
         self.show_level = show_level
         self.show_path = show_path
         self.time_format = time_format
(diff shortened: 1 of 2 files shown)


willmcgugan:
Thanks. I think it fits better to reverse the bool, i.e. have the default as `omit_repeated_times=True.`

efenka:
Sure, makes sense. I updated the request in accordance.

willmcgugan:
Thanks.

PR #1113, opened 2021-03-17
Function Name in Logger by Default

sizhky:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Really wanted to have function name as part of file name and line number

### DIFF
--- a/rich/logging.py
+++ b/rich/logging.py
@@ -200,7 +200,7 @@ class RichHandler(Handler):
             time_format=time_format,
             level=level,
             path=path,
-            line_no=record.lineno,
+            line_no=f'{record.funcName}:{record.lineno}',
             link_path=record.pathname if self.enable_link_path else None,
         )
         return log_renderable


sizhky:
The edit is literally one line, and works by default, so I did not want to add tests/documentation to the code. 
But I believe the feature will be a good addition to the library nonetheless

rich/logging.py
```diff
@@ -200,7 +200,7 @@ def render(
             time_format=time_format,
             level=level,
             path=path,
-            line_no=record.lineno,
+            line_no=f'{record.funcName}:{record.lineno}',
```
willmcgugan, line 203: This may break typing since line_no is an integer.

willmcgugan:
I wouldn't want to do this by default, mainly for space issues--a long function name could leave little space for the rest of the log text.

It would be a nice addition, but it would have to be optional and off by default.

rich/logging.py
willmcgugan: Yep

Issue #1117, opened 2021-03-17
[BUG] rich.console.Console.print ignores `markup=False` when printing rich.table.Table objects


0xDEC0DE:
**Describe the bug**
Attempting to disable markup processing via `rich.console.Console.print` does not work as documented when printing `Table` objects

**To Reproduce**
```
from rich.console import Console
from rich.table import Table
console = Console()
table = Table()
table.add_column("This is [/] not actually marked up text")
console.print(table, markup=False)
```
****Expected Result****

A table.  The [docs](https://rich.readthedocs.io/en/stable/markup.html#rendering-markup) explicitly state that `print(markup=False)` should allow it.

****Actual Behavior****

A stack trace, ending with `rich.errors.MarkupError: closing tag '[/]' at position 8 has nothing to close`

****Errata****

Disabling markup for the entire `Console` works as expected:
```
from rich.console import Console
from rich.table import Table
console = Console(markup=False)
table = Table()
table.add_column("This is [/] not actually marked up text")
console.print(table)
```

`markup=False` also appears to be respected for plain strings:
```
from rich.console import Console
console = Console()
console.print("This is [/] not actually marked up text", markup=False)
```

**Platform**
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G8022

**Diagnose**

```
$ python -m rich.diagnose
╭─────────────────────────────────────────── <class 'rich.console.Console'> ───────────────────────────────────────────╮
│ A high level console interface.                                                                                      │
│                                                                                                                      │
│ ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮ │
│ │ <console width=120 ColorSystem.TRUECOLOR>                                                                        │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                                      │
│     color_system = 'truecolor'                                                                                       │
│         encoding = 'utf-8'                                                                                           │
│             file = <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>                                     │
│ is_dumb_terminal = False                                                                                             │
│       is_jupyter = False                                                                                             │
│      is_terminal = True                                                                                              │
│   legacy_windows = False                                                                                             │
│         no_color = False                                                                                             │
│          options = ConsoleOptions(legacy_windows=False, min_width=1, max_width=120, is_terminal=True,                │
[...]

PR #1118, opened 2021-03-17
Allow Console.print(markup=) to override Console defaults

0xDEC0DE:
## Type of changes

- [✅] Bug fix

## Checklist

- [✅] I've run the latest [black](https://github.com/psf/black) with default args on new code.
- [✅] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate.
- [✅] I've added tests for new code.
- [✅] I accept that @willmcgugan may be pedantic in the code review.

## Description

When `print(markup=)` is passed in, set the Console default value in the context manager so that renderables can see it.

Fixes: issue #1117

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -1402,6 +1402,9 @@ class Console:
             crop = False
 
         with self:
+            if markup is not None:
+                self._markup = markup
+
             renderables = self._collect_renderables(
                 objects,
                 sep,
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -152,6 +152,19 @@ def test_no_columns():
     assert output == "\n"
 
 
+def test_print_markup_false():
+    no_markup = "Do not [/] attempt to markup this text"
+    console = Console()
+    table = Table()
+    table.add_column("test_print_markup_false")
+    table.add_row(no_markup)
+    console.begin_capture()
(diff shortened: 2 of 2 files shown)


0xDEC0DE:
If readers are inclined to paint this proposed change as a greasy hack, I'm inclined to agree, so don't worry about harsh words hurting my feeling.  😄  It does appear to be an adequate dodge of the stated problem, however.

Attempting to passing around the `markup=` value everywhere seemed to be a REALLY invasive change to my inexperienced eyes, so I'm not convinced that's the correct approach, either.  If there's a better, and more obvious solution, you'll have my attention.

kevinbringard:
This seems to make sense to my also inexperienced eyes, and I can confirm it resolves the immediate problem I was seeing

willmcgugan:
Close, but no cigar.

This will disable markup for every subsequent print. The fix is probably to add `markup` to ConsoleOptions.

In the meantime, you can wrap your table cells in a `rich.text.Text` instance.

retr0h:
> [✅] I accept that @willmcgugan may be pedantic in the code review.

I'm stealing this for all of my PR checklists!

willmcgugan:
There will be a fix for this in the next version.

PR #1125, opened 2021-03-21
Disabled Progress should not start attached Live

JoshKarpel:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

If you make a `Progress` that is disabled, it will still display an initial version of itself because it is the renderable of its attached `Live`, and the `Live` is started unconditionally in `Progress.start`. Example:

```python

with Progress(disable=True) as progress:
    task = progress.add_task("foobar", total=100)
    progress.update(task, advance=1)

print("done")
```

Running this produces this output:
image

With this patch, it produces this:
image

---

I added some tests for this, but they're poking around the internals to do it - let me know if this isn't your preferred style.


willmcgugan:
Thanks for this. Would you mind also testing the output? i.e. confirm that when `disable=True` nothing gets written. Just to catch any regression.

willmcgugan:
Thanks.

PR #1131, opened 2021-03-24
typing: unify type hints for `total` keyword arguments in rich.progress

neutrinoceros:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [NA] I've added tests for new code.

## Description

I think this qualifies as a bug fix since the underlying object in all cases is `rich.progress.Task`, which is already documented as using a `float` total argument rather than an `int`.


neutrinoceros:
switching to draft until I manage to actually satisfy mypy

willmcgugan:
Thanks.

PR #1138, opened 2021-03-29
Status spinners should not reset animations on update

SandmanDP:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description
Updating a rich.status.Status object currently resets the spinner animation in the object. This set of code changes adds a new function to spinners to update parameters, and leverages this function in Status objects to prevent the constant creation of spinner objects that causes an animation reset.

Before:
https://user-images.githubusercontent.com/8658674/112775847-e9111300-8ff2-11eb-8e4c-c42e7c62e7a3.mov

After:
https://user-images.githubusercontent.com/8658674/112775866-f4643e80-8ff2-11eb-8f69-6664b3a18ac8.mov


SandmanDP:
This PR makes a slightly invasive change to the expected type of the status parameter to Status to align with the text field of Spinner, and subsequently changes the renderable type of Status (a table is no longer required as the status text field now leverages spinner text).

Can revert this latter change if desired, but seeing as how the doc strings previously indicated that str/Text was the expected type anyways I saw this as a possible amendment

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improved layout.tree
 - Changed default theme color for repr.number to cyan
 - `__rich_measure__` signature changed to accept ConsoleOptions rather than max_width
+- `status` parameter to rich.status.Status changed to TextType
```
willmcgugan, line 20: Why do this? It works with any renderable.

```python
from rich.panel import Panel
c=Console()
from time import sleep
with c.status(Panel("foo")):
    sleep(10)
```

SandmanDP: Wanted to see if this was viable change based on the docstrings and text type of the wrapped spinner object. That is, since the spinner text already expects a Text object and since the status object documentation seems to prefer text input, was hoping that this modification to status would be permitted.

The alternative solution would be to allow the spinner object to fully manage its own updating and text field would be to change the spinner object itself to take in a RenderableType. Sounds like that's what you'd prefer for this situation, so I've gone ahead and done that as the latest commit to this PR.

willmcgugan: Calling the Spinner arg "text" was probably an error on my part, but it would be a shame to prohibit other renderables since it works.

willmcgugan:
I think there may be an alternative way of handling the speed change.

If at the point of update you calculate `self.offset_frame_no` as `(time * self.speed) / (self.interval / 1000.0)`, then add that to the `frame_no` calculation in `render()` it should correctly pick the next frame when speed changes.

You will also need to reset `start_time` to the current time on update.

I suspect that will ultimately be simpler. Would you mind trying that?

SandmanDP:
@willmcgugan wouldn't be able to do that in the `update` function unfortunately, since that function doesn't receive access to the current time since it's not in the direct rendering path. Would still need to perform the calculation in `__rich_console__` or `render`, which would mean that we'd still need to keep track of `_update_speed` from the update function and a new `frame_offset_no` attribute, which I don't think reduces complexity too much =/

Still want me to try this out?

willmcgugan:
I see what you mean.

Still, I think that calculation may be preferable. I don't quite follow how your adjustment to start_time works. I assume it keeps the animation frame correct, but it would mean that `start_time` is no longer the time it was started or updated?

SandmanDP:
@willmcgugan sorry for the delay, I believe I've addressed your comment. Can't for the life of me figure out why there's a reported coverage decrease on progress.py, I literally only made a typing change in that file.

willmcgugan:
Thanks, I'll check this out shortly. Don't worry about codecov, for some reason it gives false positives. Sometimes it says coverage has gone down after changing a docstring.

willmcgugan:
Thanks. Looks good. There is a merge conflict, but that will be a quick fix.

PR #1145, opened 2021-03-30
Add checkMark and crossMark status

davidbrochart:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This allow to mark a status as "done" or "failed" with a check mark or a cross mark instead of a spinner:
```python
status.update("Done", spinner="checkMark")
status.update("Failed", spinner="crossMark")
```
✅ Done
❌ Failed

### DIFF
--- a/rich/_spinners.py
+++ b/rich/_spinners.py
@@ -845,4 +845,12 @@ SPINNERS = {
             "▰▱▱▱▱▱▱",
         ],
     },
+    "checkMark": {
+        "interval": 1000,
+        "frames": ["✅"],
+    },
+    "crossMark": {
+        "interval": 1000,
+        "frames": ["❌"],
+    },
 }


willmcgugan:
Not sure why you would need that. Presumably once you are done, you can exit the status context and print whatever you like?

davidbrochart:
I'm using a live tree that represents running tasks, with subtasks down in the hierarchy, so I need to flag tasks as done or failed, but the tree rendering still goes on.

davidbrochart:
> you can exit the status context and print whatever you like

Maybe there's a better way to do it, but I want to change the status in-place to "done" or "failed".

willmcgugan:
Fair enough. Seems like a reasonable solution.

willmcgugan:
Actually, it may be a little top specific. I can imagine other users wanting different unicode symbols and I couldn't add a spinner for all requirements.

Could you replace the spinner with `Text("✅ Done")` (or cross) as needed?

davidbrochart:
I'm not sure to understand, do you mean something like:
```python
status.update("✅ Done")
```
But then we still have a leading spinner?

willmcgugan:
I meant replace status instance with a new Text instance.

If you have a Tree node (returned from Tree.add) you can do this:

```python
tree_node.label = "✅ Done"
```

Which will replace the status with Text.

Alternatively you could implement a renderable that returns a Spinner or the desired text. Something along these lines:

```python
class MyLabel:
    def __init__(self):
        self.spinner = Spinner()
    
    def __rich__(self):
        if is_finished():
            return "Done" if status == "done" else "Fail"
        else:
            return self.spinner

tree = Tree()
tree.add(MyLabel())
```

davidbrochart:
> If you have a Tree node (returned from Tree.add) you can do this:
> 
> ```python
> tree_node.label = "✅ Done"
> ```
> 
> Which will replace the status with Text.

Works like a charm, thanks a lot.
This is my first day using Rich, and it's awesome :smile:

PR #1165, opened 2021-04-10
Stop showing colon after exception type when there is no message

jack1142:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This updates the shown traceback so that it does not include a colon if there is no exception message. This is in line with Python's default traceback handler.
Here's a screenshot that will probably explain it even better:
image


willmcgugan:
Thanks.

PR #1167, opened 2021-04-13
Fixes a split_row() bug in examples/fullscreen.py

chief8192:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Fixes the `fullscreen.py` example to use `Layout.split(direction='horizontal')` instead of `Layout.split_row()`. Verified to work locally via `python3 fullscreen.py`.

### DIFF
--- a/examples/fullscreen.py
+++ b/examples/fullscreen.py
@@ -27,9 +27,10 @@ def make_layout() -> Layout:
         Layout(name="main", ratio=1),
         Layout(name="footer", size=7),
     )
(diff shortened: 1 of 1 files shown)


willmcgugan:
I think you may have an older version of Rich installed, because the direction argument was removed in version 10.0.0

chief8192:
Ahh, gotcha. Sorry for the noise.

PR #1187, opened 2021-04-24
Use typing-extensions on py<3.8 

mgorny:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Use `typing-extensions` package only in Python < 3.8. Python 3.8+ has everything in the built-in `typing` module.

Note: I didn't update `poetry.lock` as this causes a major change to everything. I don't really know how poetry-based workflow works.

### DIFF
--- a/examples/top_lite_simulator.py
+++ b/examples/top_lite_simulator.py
@@ -9,7 +9,11 @@ from rich import box
 from rich.console import Console
 from rich.live import Live
 from rich.table import Table
-from typing_extensions import Literal
+
+try:
+    from typing import Literal
+except ImportError:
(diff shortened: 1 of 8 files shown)


willmcgugan:
Are there any benefits of this other than reducing a dependency on 3.8? It's not like anything is broken in 3.8 or above.

mgorny:
We're dropping support for Python 3.7 in Gentoo soon. This helps people who don't need old Pythons to stop caring about one more dependency.

willmcgugan:
Ok, fair enough.

PR #1203, opened 2021-05-02
mypy strict compliant

nathanrpage97:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

This is a first draft of implementing Mypy strict. I tried to minimize using `type:ignore`, `casts`, and `Any` features. Let me know what you think.

### DIFF
--- a/rich/__main__.py
+++ b/rich/__main__.py
@@ -1,10 +1,17 @@
 import colorsys
 import io
 from time import process_time
+from typing import Any
 
 from rich import box
 from rich.color import Color
-from rich.console import Console, ConsoleOptions, RenderGroup, RenderResult
+from rich.console import (
+    Console,
+    ConsoleOptions,
+    ConsoleRenderable,
+    RenderGroup,
+    RenderResult,
+)
 from rich.markdown import Markdown
 from rich.measure import Measurement
 from rich.pretty import Pretty
@@ -88,7 +95,7 @@ def make_test_card() -> Table:
         ),
     )
 
-    def comparison(renderable1, renderable2) -> Table:
+    def comparison(renderable1: Any, renderable2: Any) -> Table:
         table = Table(show_header=False, pad_edge=False, box=None, expand=True)
         table.add_column("1", ratio=1)
         table.add_column("2", ratio=1)
--- a/rich/__main__.py
+++ b/rich/__main__.py
@@ -11,6 +11,7 @@ from rich.console import (
     ConsoleRenderable,
     RenderGroup,
     RenderResult,
+    RenderableType,
 )
 from rich.markdown import Markdown
 from rich.measure import Measurement
@@ -95,7 +96,7 @@ def make_test_card() -> Table:
         ),
     )
 
-    def comparison(renderable1: Any, renderable2: Any) -> Table:
+    def comparison(renderable1: RenderableType, renderable2: RenderableType) -> Table:
         table = Table(show_header=False, pad_edge=False, box=None, expand=True)
         table.add_column("1", ratio=1)
         table.add_column("2", ratio=1)
--- a/rich/_emoji_replace.py
+++ b/rich/_emoji_replace.py
@@ -5,11 +5,13 @@ import re
 from ._emoji_codes import EMOJI
 
 
-_EmojiSub = Callable[[Callable[[Match[str]], str], str], str]
+_ReStringMatch = Match[str]  # regex match object
+_ReSubCallable = Callable[[_ReStringMatch], str]  # Callable invoked by re.sub
+_EmojiSubMethod = Callable[[_ReSubCallable, str], str]  # Sub method of a compiled re
 
 
 def _emoji_replace(
-    text: str, _emoji_sub: _EmojiSub = re.compile(r"(:(\S*?):)").sub
+    text: str, _emoji_sub: _EmojiSubMethod = re.compile(r"(:(\S*?):)").sub
 ) -> str:
     """Replace emoji code in text."""
     get_emoji = EMOJI.get
--- a/rich/_emoji_replace.py
+++ b/rich/_emoji_replace.py
@@ -1,11 +1,16 @@
-from typing import Match
+from typing import Callable, Match
 
 import re
 
 from ._emoji_codes import EMOJI
 
 
-def _emoji_replace(text: str, _emoji_sub=re.compile(r"(:(\S*?):)").sub) -> str:
+_EmojiSub = Callable[[Callable[[Match[str]], str], str], str]
+
+
+def _emoji_replace(
+    text: str, _emoji_sub: _EmojiSub = re.compile(r"(:(\S*?):)").sub
+) -> str:
     """Replace emoji code in text."""
     get_emoji = EMOJI.get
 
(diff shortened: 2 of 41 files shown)


rich/__main__.py
```diff
@@ -88,7 +95,7 @@ def make_test_card() -> Table:
         ),
     )
 
-    def comparison(renderable1, renderable2) -> Table:
+    def comparison(renderable1: Any, renderable2: Any) -> Table:
```
willmcgugan, line 98: Should really be `RenderableType` rather than Any.

rich/_emoji_replace.py
```diff
@@ -1,11 +1,16 @@
-from typing import Match
+from typing import Callable, Match
 
 import re
 
 from ._emoji_codes import EMOJI
 
 
-def _emoji_replace(text: str, _emoji_sub=re.compile(r"(:(\S*?):)").sub) -> str:
+_EmojiSub = Callable[[Callable[[Match[str]], str], str], str]
```
willmcgugan, line 8: This was hard to parse in my head, but I double checked it and it's entirely correct. Bravo.

The thing is that readability has been thrown out for the sake of correctness, which I'm no fan of. Could we break this up a bit to make it more readable? Something along these lines:

```python
_ReStringMatch = Match[str]  # regex match object
_ReSubCallable = Callable[[_ReStringMatch], str] # Callable invoked by re.sub
_EmojiSubMethod = Callable[_ReSubCallable, str]. # Sub method of a compiled re
```

rich/_timer.py
```diff
@@ -6,10 +6,11 @@
 from time import time
 
 import contextlib
+from typing import Iterator
 
 
 @contextlib.contextmanager
-def timer(subject: str = "time"):
+def timer(subject: str = "time") -> Iterator[None]:
```
willmcgugan, line 13: I think `typing.Generator` would be more accurate here.

rich/_windows.py
```diff
@@ -18,7 +19,10 @@ class WindowsConsoleFeatures:
     from ctypes import wintypes
     from ctypes import LibraryLoader
 
-    windll = LibraryLoader(ctypes.WinDLL)  # type: ignore
+    if sys.platform == "win32":
```
willmcgugan, line 22: What was this solving

I suspect that if windll is None, it will break on L38.

rich/_windows.py
```diff
@@ -52,8 +56,8 @@ def get_windows_console_features() -> WindowsConsoleFeatures:
         result = _GetConsoleMode(handle, console_mode)
         vt = bool(result and console_mode.value & ENABLE_VIRTUAL_TERMINAL_PROCESSING)
         truecolor = False
-        if vt:
-            win_version = sys.getwindowsversion()  # type: ignore
+        if sys.platform == "win32" and vt:
```
willmcgugan, line 59: Can we get here if we are not on Windows?

rich/console.py
```diff
@@ -294,7 +294,7 @@ def __enter__(self) -> "Capture":
         self._console.begin_capture()
         return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
```
willmcgugan, line 297: I can never remember the types here, which is why omitted them but let's be accurate since we are --strict.

This is the types, copied from another project. Python2 syntax. 

``` python
    def __exit__(
        self,
        exc_type,  # type: Optional[Type[BaseException]]
        exc_value,  # type: Optional[BaseException]
        traceback,  # type: Optional[TracebackType]
    ):
```

rich/console.py
```diff
@@ -856,7 +858,7 @@ def is_terminal(self) -> bool:
         """
         if self._force_terminal is not None:
             return self._force_terminal
-        isatty = getattr(self.file, "isatty", None)
+        isatty: Optional[Callable[..., bool]] = getattr(self.file, "isatty", None)
```
willmcgugan, line 861: This takes no parameters, so I think the type should be this:

```
Optional[Callable[[], bool]] 
```

rich/containers.py
```diff
@@ -82,8 +84,8 @@ def __getitem__(self, index: int) -> "Text":
     def __getitem__(self, index: slice) -> "Lines":
         ...
 
-    def __getitem__(self, index):
-        return self._lines[index]
+    def __getitem__(self, index: Union[slice, int]) -> Union["Text", "Lines"]:
```
willmcgugan, line 87: I think the return types should be Text or List.

rich/control.py
```diff
@@ -157,7 +157,7 @@ def __rich_console__(
             yield self.segment
 
 
-def strip_control_codes(text: str, _translate_table=_CONTROL_TRANSLATE) -> str:
+def strip_control_codes(text: str, _translate_table: Any = _CONTROL_TRANSLATE) -> str:
```
willmcgugan, line 160: I think that table should be `Dict[int, None]`

rich/markdown.py
```diff
@@ -107,7 +108,7 @@ class Paragraph(TextElement):
     justify: JustifyMethod
 
     @classmethod
-    def create(cls, markdown: "Markdown", node) -> "Paragraph":
+    def create(cls, markdown: "Markdown", node: Any) -> "Paragraph":
```
willmcgugan, line 111: I think node should be "MarkdownElement"

rich/markup.py
```diff
@@ -36,7 +36,12 @@ def markup(self) -> str:
         )
 
 
-def escape(markup: str, _escape=re.compile(r"(\\*)(\[[a-z#\/].*?\])").sub) -> str:
+_EscapeSub = Callable[[Callable[[Match[str]], str], str], str]
```
willmcgugan, line 39: Could also use breaking up in to bite size chunks, for readability sake.

rich/traceback.py
```diff
@@ -47,7 +47,7 @@ def install(
     word_wrap: bool = False,
     show_locals: bool = False,
     indent_guides: bool = True,
-) -> Callable:
+) -> Callable[..., Any]:
```
willmcgugan, line 50: Let's type this. See L72 for the signature.

willmcgugan (changes requested):
Forgive my initial reluctance, but I think this exercise is worthwhile. Mostly requests for extra detail in the typing, and maybe a hypothetical bug in _windows.

nathanrpage97:
Thanks for the feedback, I will work on these changes later this week.

rich/_timer.py
nathanrpage97: https://github.com/python/typeshed/blob/master/stdlib/contextlib.pyi#L25 shows it takes in an iterator.

rich/_windows.py
nathanrpage97: Reverting to old method with a type ignore

willmcgugan:
HI @nathanrpage97 I've created a merge conflict for you. Would you mind fixing that? Is this ready for another review?

I'm going to publish a new version soon, and it would be nice to get this in.

willmcgugan:
Fixed merge conflict. This looks good, thanks!

PR #1217, opened 2021-05-07
Pretty attrs

willmcgugan:
Adds pretty printing support for attrs module.

### DIFF
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -11,7 +11,7 @@ from typing import (
     Callable,
     Dict,
     Iterable,
-    List,
+    List,    
     Optional,
     Set,
     Union,
@@ -21,7 +21,7 @@ from typing import (
 try:
     import attr as _attr_module
 except ImportError:  # pragma: no cover
-    _attr_module = None
+    _attr_module = None  # type: ignore
 
 
 def _is_attr_object(obj: Any) -> bool:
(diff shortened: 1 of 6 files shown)


rich/pretty.py
```diff
@@ -500,6 +515,50 @@ def iter_rich_args(rich_args) -> Iterable[Union[Any, Tuple[str, Any]]]:
...
+                            try:
+                                value = getattr(obj, attr.name)
+                            except Exception as error:
+                                # Can happen, albeit rarely
+                                yield (attr.name, error, None)
+                            else:
+                                yield (
+                                    attr.name,
+                                    value,
+                                    attr.repr if callable(attr.repr) else None,
```
hynek, line 546: What exactly does this logic do? `repr` being a callable doesn't mean that you have to call it yourself, it's used by the object's `__repr__`. Or do you use it as a marker, that _you_ shouldn't try to render it?

willmcgugan: Rich will only call `repr` on atomic values. For containers, Rich will take over the rendering of the repr so that it can know how to expand it on to several lines if required.

This logic is so that the custom `repr` will be called if supplied (L551), otherwise it will recurse in to the value (L553), potentially expanding lists and dicts etc. on to several lines.

Does that make sense, or have I misunderstood the purpose of Attribute.repr ?

hynek: No I think you've got it. A truish repr for an attribute means that `attrs` include it in `__repr__` and if it's a callable, it will use that callable to generate said repr instead of calling `repr()` on the value: https://github.com/python-attrs/attrs/blob/8ae2d6f1c3c067c941fd9d212e52972c7d3382cc/src/attr/_make.py#L1855-L1858

Issue #1149, opened 2021-04-01
[BUG] Console Status not working with Spyder


FabioRosado:
**Read the docs**
You might find a solution to your problem in the [docs](https://rich.readthedocs.io/en/latest/introduction.html) -- consider using the search function there.

**Describe the bug**

I am using status to get a spinner while things are being done in the background. I'm essentially running:

```python
 console = Console()

        with console.status(
            "[bold green] Creating Cluster. This takes about a minute..."
        ):
```

That spews a lot of these in a Spyder console:

```
<rich.jupyter.JupyterRenderable at **0x1c0ffeaaee0**>
```

I'm guessing that perhaps Rich is detecting the Spyder IPyython window as a Jupyter one.

**To Reproduce**
A minimal code example that reproduces the problem would be a big help if you can provide it. If the issue is visual in nature, consider posting a screenshot.

- Install Anaconda (or just Spyder) on a Windows machine
- Create a dummy code similar to above
- See the output

**Platform**
What platform (Win/Linux/Mac) are you running on? What terminal software are you using?

Windows, IPython console on Spyder

**Diagnose**
I may ask you to cut and paste the output of the following commands. It may save some time if you do it now.

```
python -m rich.diagnose
python -m rich._windows
pip freeze | grep rich
```

```
# python -m rich.diagnose
+---------------------- <class 'rich.console.Console'> -----------------------+
| A high level console interface.                                             |
|                                                                             |
| +-------------------------------------------------------------------------+ |
| | <console width=79 None>                                                 | |
| +-------------------------------------------------------------------------+ |
|                                                                             |
|     color_system = None                                                     |
|         encoding = 'cp1252'                                                 |
|             file = <colorama.ansitowin32.StreamWrapper object at            |
|                    0x000001843F9A0EB0>                                      |
|           height = 24                                                       |
|    is_alt_screen = False                                                    |
| is_dumb_terminal = False                                                    |
|   is_interactive = False                                                    |
|       is_jupyter = False                                                    |
|      is_terminal = False                                                    |
|   legacy_windows = True                                                     |
|         no_color = False                                                    |
|          options = ConsoleOptions(                                          |
|                        size=ConsoleDimensions(width=79, height=24),         |
|                        legacy_windows=True,                                 |
|                        min_width=1,                                         |
|                        max_width=79,                                        |
|                        is_terminal=False,                                   |
|                        encoding='cp1252',                                   |
|                        justify=None,                                        |
|                        overflow=None,                                       |
[...]

PR #1218, opened 2021-05-07
fix color output in spyder console

eendebakpt:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Enable colored output with [Spyder](https://github.com/spyder-ide/spyder)

This PR addresses two issues: 

* `colorama` needs to be initialized with the option `strip=False` in Spyder
* In the spyder environment the method `is_terminal` returns False, where it should return True. This is solved by creating a new method `detect_spyder_environment`

Related: #1149, https://github.com/spyder-ide/spyder/issues/15496, https://github.com/spyder-ide/spyder/issues/1917

A screenshot of Spyder with the PR:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed initial blank lines removed from Syntax https://github.com/willmcgugan/rich/issues/1214
 - Added Console.measure as a convenient alias for Measurement.get
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


willmcgugan:
I can't accept a fudge for a specific IDE. Colorama is a headache for Rich as it makes global changes, whereas Rich can have different settings per-Console instance.

I'd suggest setting `force_terminal` on the constructor if Spyder is reporting it is not a real terminal. You can also reinit colorama after importing rich to fix that issue.

PR #1252, opened 2021-05-22
Add Python 3.10 to test matrix

sondrelg:
## Type of changes

- [x] Bug fix

## Checklist

## Description

wip: Just checking that tests pass before finalizing


.github/workflows/pythonpackage.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10.0-alpha.4"]
```
willmcgugan, line 11: I think there is a beta version now

sondrelg: Yep unfortunately it doesn't seem to have been added to Github's workflows yet 🙂 Tried the beta version [here](https://github.com/willmcgugan/rich/runs/2647076941?check_suite_focus=true) and the workflow raises an error saying `Version 3.10.0b1 with arch x64 not found`. Maybe they have to add them in manually and there's a lag?

For context on the PR, I was hoping to verify that the [install poetry](https://github.com/snok/install-poetry) action now worked well with 3.10, related to [this issue](https://github.com/willmcgugan/rich/issues/1216), and the action looks to install OK in [this run](https://github.com/willmcgugan/rich/runs/2647562295?check_suite_focus=true), but there seems to also be a resolver issue. It does looks fixable though, so I think I'll try and tinker with it a little bit to see if I can get it to work 👍

willmcgugan: Should the beta version not being given as "3.10.0-beta.1" ? It looks like in the workflow it is given as "3.10.0-b1"

I found the list of Python versions [here](https://github.com/actions/python-versions/blob/main/versions-manifest.json)

sondrelg: Nice 👍 Added to the matrix, you're absolutely right

.github/workflows/pythonpackage.yml
```diff
@@ -20,21 +20,25 @@ jobs:
           python-version: ${{ matrix.python-version }}
           architecture: x64
       - name: Install and configure Poetry
-        uses: snok/install-poetry@v1.1.1
+        uses: snok/install-poetry@v1.1.6
         with:
-          version: 1.1.4
-          virtualenvs-create: false
+          version: 1.1.6
```
sondrelg, line 25: This is unintentionally possibly a little confusing. Just to clarify - the `version` parameter here specifies the Poetry version, while the `@1.1.6` specifies the action version. They just happen to be the same.

.github/workflows/pythonpackage.yml
```diff
@@ -20,21 +20,25 @@ jobs:
...
+          virtualenvs-in-project: true
       - name: Install dependencies
         run: poetry install
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
       - name: Format check with black
-        run: make format-check
+        run: |
+          source $VENV
+          make format-check
```
sondrelg, line 33: Ran into some problems with the old setup, so switched from installing dependencies directly inside the runner environment, to installing dependencies to a venv and activating the venv in consequent steps.

Just for context, instead of `source $VENV`, we *could* have used `source .venv/bin/activate`, but since there's a windows combination in the matrix you would need to introduce conditional steps; on Windows you'd want ~`source .venv/scripts/activate`. 

The `$VENV` environment variable is set inside the `install-poetry` action so upstream workflows don't need to worry about which OS you're on 🙂

willmcgugan:
Thanks!

PR #1274, opened 2021-06-06
add ipython extension for rich

nathanrpage97:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

I think that we can provide better integration with IPython by creating an extension. This provides a more conventional way of adding functionality to IPython. For now I have kept it rather basic, but later on there could be custom magic commands that could be added.


nathanrpage97:
If you do not want it in the `__init__.py` is to rename `_extension.py` to `ipython.py` and have the user call it via `%load_ext rich.ipython`. This is a common convention I see being used by the IPython community.

willmcgugan:
Thanks!

PR #1288, opened 2021-06-11
Exploratory Contribution: MUD Support (MXP/Pueblo)

volundmush:
I don't expect this Pull Request to pass as-is. It would definitely need a bit more polish for that. But see the description!

Basically, I'm looking for input on whether this SORT of contribution would be desired, before I commit to going further.

Please let me know if it's not; then I might have to fork. I tried to integrate these features via monkey-patching, but it caused a lot of problems.

## Type of changes

- [ ] Bug fix
- [X ] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [X ] I've run the latest [black](https://github.com/psf/black) with default args on new code.

## Description

This patch would add support for MXP (Mud eXtension Protocol) and Pueblo. This required some minor edits to Text, Style, and Console. Because I'm working on a project that relies heavily on Text objects and treating them largely like normal strings, the Text object API was also expanded to handle much more of the Python String API.

Also added the ability to specify BOX type for Traceback printing, because the default one looks terrible in monospace MUD setups.

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -141,6 +141,12 @@ class ConsoleOptions:
     """Enable markup when rendering strings."""
     height: Optional[int] = None
     """Height available, or None for no height limit."""
+    mxp: Optional[bool] = False
+    """Enable MXP/MUD HTML when printing. For MUDs only."""
+    pueblo: Optional[bool] = False
+    """Enable Pueblo/MUD HTML when printing. For MUDs only."""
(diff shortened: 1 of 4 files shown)


willmcgugan:
You've done some very interesting work there, but MUD support is probably too niche to add to the core library. You are probably better off with a fork of the lib. You could even publish it on Pypy as rich-mud or something.

PR #1364, opened 2021-07-24
Adds missing prop to `__slots__` of `Emoji`

sobolevn:
Refs https://github.com/python/mypy/pull/10864

## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This prop was missing from `__slots__` 🙂 

There's one more thing: current `__slots__` setup does not work:

```python
>>> e = Emoji('2nd_place_medal', variant='text')
>>> e.__slots__
['name', 'style', '_char']
>>> e.__dict__
{'variant': 'text', 'a': 1, 'bb': 1}

>>> e.a = 1  # should fail, but does not
>>> e.bb = 1  # should fail, but does not
```

This happens due to the fact that `JupiterMixin` does not have `__slots__ = ()` defined.
I can add it if you wish 🙂

### DIFF
--- a/rich/emoji.py
+++ b/rich/emoji.py
@@ -25,7 +25,7 @@ class NoEmoji(Exception):
 
 
 class Emoji(JupyterMixin):
-    __slots__ = ["name", "style", "_char"]
+    __slots__ = ["name", "style", "_char", "variant"]
 
     VARIANTS = {"text": "\uFE0E", "emoji": "\uFE0F"}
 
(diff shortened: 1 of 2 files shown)


willmcgugan:
Good catch. Yes, please add the missing slots.

sobolevn:
Whitespace issues are the worst! 😒

sobolevn:
This looks like a bug in `mypy` that I am going to fix soon (hopefully):

```
mypy -p rich --strict
rich/jupyter.py:34: error: Need type annotation for "__slots__" (hint: "__slots__: List[<type>] = ...")
```

willmcgugan:
Ah, it's complaining about the empty slots? If an empty tuple works, may as well go with that...

willmcgugan:
Thanks.

PR #1382, opened 2021-08-03
Optimisation suggestions for Segment.divide()

AlexWaygood:
Three suggested changes:

- `from __future__ import annotations` might give a slight performance increase as it means type hints don't need to be evaluated.
- Assign `iter(cuts).__next__` to a local function rather than having to look `next` up in the global scope every time (same as in the `itertools` recipe for `roundrobin`).
- Put your whole loop inside a `try/except` clause in lines 477ff, rather than having to do a `try/except` every iteration in the loop. Same as the `itertools` recipe for `iter_except`, and should be just as safe given that you're only catching `StopIteration`.

## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

### DIFF
--- a/rich/segment.py
+++ b/rich/segment.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from enum import IntEnum
 from logging import getLogger
 from typing import Dict, NamedTuple, Optional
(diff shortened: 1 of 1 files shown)


willmcgugan:
The annotations are only evaluated at import time, so shouldn't have any impact.

Good suggestions, but these micro-optimizations probably won't make a measurable impact. I'm looking for a more algorithmic improvement.

AlexWaygood:
Fair enough!

PR #1388, opened 2021-08-06
revert expected test_repr value for py3.10 following upstream

mgorny:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

All Enum changes were reverted from Python 3.10 just before rc1.  Revert
the test changes to fix test failures with 3.10.0rc1.


mgorny:
I don't think the original change to the test has been mentioned in the changelog, so I skipped it here too.

willmcgugan:
👍

PR #1446, opened 2021-08-29
Added support to print specific key in a json 

sureshdsk:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

print_json method can take an optional key and pretty prints only specific key in the json

```
rich.print_json(
            """{"user":"suresh","skill":{"languages":["python","javascript","golang"],"cloud":["google cloud","aws"]}}""",
            key="skill"
        )
```


willmcgugan:
Why do you need this? I don't see this as being particularly useful in its current form.  Perhaps if it was more sophisticated than a single key, but then it would be something outside of the scope of Rich.

PR #1467, opened 2021-09-03
Use builtin html.escape

tusharsadhwani:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Replace hand-rolled html escape function with builtin `html.escape`.

Has the added benefit of escaping `"` with `&quot;` and `'` with `&#x27;`.

Also deleted a stray print in the test.

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -10,6 +10,7 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from functools import wraps
 from getpass import getpass
+from html import escape
 from inspect import isclass
 from itertools import islice
 from time import monotonic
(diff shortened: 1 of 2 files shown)


tusharsadhwani:
What's with test coverage :thinking: There's one less line of code (the print statement), so it thinks the coverage went down? Is that it? or am I missing something

willmcgugan:
I’ve noticed that before. Coverage can be weird.

tusharsadhwani:
I would suggest using pytest-cov instead

willmcgugan:
Thanks.

PR #1473, opened 2021-09-09
Add arrows option to rich.tree

tusharsadhwani:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Saw some arrows in another project. Looked pretty neat, so I tried adding those into rich.

<details><summary>Code snippet:</summary>

```python
from rich import print
from rich.tree import Tree

tree = Tree("my tree", guide_style="red", arrows=True)
for i in range(5):
    tree.add(f"item {i}")
print(tree)

tree = Tree("my tree", guide_style="bold", arrows=True)
for i in range(5):
    tree.add(f"item {i}")
print(tree)
```

</details>

Example output:

image

Thoughts?

### DIFF
--- a/rich/tree.py
+++ b/rich/tree.py
@@ -18,6 +18,7 @@ class Tree(JupyterMixin):
         guide_style (StyleType, optional): Style of the guide lines. Defaults to "tree.line".
         expanded (bool, optional): Also display children. Defaults to True.
         highlight (bool, optional): Highlight renderable (if str). Defaults to False.
+        arrows (Optional[bool], optional): Show arrows at the end of guide lines. Defaults to False.
     """
 
     def __init__(
@@ -28,6 +29,7 @@ class Tree(JupyterMixin):
         guide_style: StyleType = "tree.line",
         expanded: bool = True,
         highlight: bool = False,
+        arrows: bool = False,
     ) -> None:
         self.label = label
         self.style = style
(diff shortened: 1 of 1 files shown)


willmcgugan:
I've never seen trees with arrows before. Where did you see them?

tusharsadhwani:
On twitter:
https://twitter.com/zkat__/status/1435820136326328326

willmcgugan:
Could you add some way to specify the arrows per-node? In-fact, maybe there are other ways of terminating the guides?

tusharsadhwani:
Yeah, I just realised that the `arrows` attribute inside `Tree.add` isn't doing what I thought it would. I'll take a look.

I'm not sure what other endings could be used. Should the guide endings be turned into an enum then? `ARROW` being one option?

willmcgugan:
That's what I was thinking. There may be [other symbols](https://www.compart.com/en/unicode/block/U+2B00) that make sense.

tusharsadhwani:
The only other symbol that might make sense imo is a filled circle, and even that might leave a gap on the left edge with the bold line.

willmcgugan:
You may be right! But even so, it would be nice to specify them per node.

tusharsadhwani:
I might have to run a debugger over the code to understand what's being rendered where. The rendering flow doesn't seem to be what I expected.

Mind if I also add some clarifying comments?

willmcgugan:
Be my guest.

willmcgugan:
Hi Tushar. Do you want to persevere with this one? Unless it is per-node I don't think it will be generally useful.

tusharsadhwani:
Yeah, I want to get to this eventually. I should wrap up a few of my projects in march, hopefully then!

PR #1490, opened 2021-09-14
Bump codecov-action (v1 is deprecated) and actions/checkout 

yashrathi-git:
[codecov action v1](https://github.com/codecov/codecov-action#migration-from-v1-to-v2) has been deprecated and it don't seems like it would be breaking change for rich.
Also updates `checkout` [action to v2](https://github.com/actions/checkout#whats-new)


willmcgugan:
Thanks

Issue #1492, opened 2021-09-16
[BUG] TypeError not callable : rich.print_exception


yoann9344:
Hello !

**Describe the bug**
I use print_exception with uvicorn and print_exception got an exception "Field not Callable", I patched [uvicorn/protocols/http/httptools_impl.py:404](https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/http/httptools_impl.py#L376) this way
```python
        except BaseException as exc:
            msg = "Exception in ASGI application\n"
            # PATCH start
            from rich.console import Console

            self.console = Console()
            self.console.print_exception(extra_lines=5, show_locals=True)
            # PATCH end
            self.logger.error(msg, exc_info=exc)
            if not self.response_started:
                await self.send_500_response()
            else:
                self.transport.close()
```
The exception which occured (the first part is the error in my side ; the second is the error in rich, both printed by uvicorn)
```
future: <Task finished name='Task-5' coro=<RequestResponseCycle.run_asgi() done, defined at .venv/lib/python3.9/site-packages/uvicorn/protocols/http/httptools_impl.py:396> exception=TypeError("'Field' object is not callable")>
Traceback (most recent call last):
  File ".venv/lib/python3.9/site-packages/uvicorn/protocols/http/httptools_impl.py", line 398, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File ".venv/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File ".venv/lib/python3.9/site-packages/fastapi/applications.py", line 199, in __call__
    await super().__call__(scope, receive, send)
  File ".venv/lib/python3.9/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File ".venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File ".venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File ".venv/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File ".venv/lib/python3.9/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File ".venv/lib/python3.9/site-packages/starlette/routing.py", line 580, in __call__
[...]

PR #1494, opened 2021-09-17
Pretty.traverse : patch obj.__getattr__ if exists

yoann9344:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

- [><"] I accept that @willmcgugan may be pedantic in the code review.

## Description
Some library can return a  `__rich_repr__` which is not callable by doing some stuff in __getattr__. This requires to patch __getattr__'s method to allow __rich_repr__ to act normally.
#1492

### DIFF
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -503,6 +503,25 @@ def traverse(
                 else:
                     yield arg
 
+        original_getattr_method = None
+
+        if (
+            hasattr(obj, "__rich_repr__")
+            and hasattr(obj, "__getattr__")
+            and not isclass(obj)
+        ):
+            original_getattr_method = obj.__getattr__
+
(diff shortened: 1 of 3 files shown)


rich/pretty.py
```diff
@@ -503,6 +503,25 @@ def iter_rich_args(rich_args: Any) -> Iterable[Union[Any, Tuple[str, Any]]]:
                 else:
                     yield arg
 
+        original_getattr_method = None
```
willmcgugan, line 506: Not keen on this solution. It is probably better to respect that an object is doing magic, but be more robust in handling errors.

yoann9344: I saw your patch, that's way better ! Thx !

PR #1540, opened 2021-10-03
Added Calendar layout example

sureshdsk:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other - Examples

## Checklist

## Description

Added example for calendar layout using Columns and Tables. 

calendar3
calendar4

### DIFF
--- a/examples/columns2.py
+++ b/examples/columns2.py
@@ -1,0 +1,56 @@
+"""
+Builds calendar layout using Columns and Tables.
+Usage:
+python columns2.py [YEAR]
+Example:
+python columns2.py 2021
+"""
+import argparse
+
+import calendar
+from datetime import datetime
+
+from rich.table import Table
+from rich.columns import Columns
+from rich import print
+
+
+def display_calendar(year=None):
+    
+    today = datetime.today()
+    year = int(year or today.year)
+    cal = calendar.Calendar()
+
+    tables = []
+    for month in range(1,13):
+
+        table = Table(title=f"""{calendar.month_name[month]} {year}""", style="green")
+
+        for wd in cal.iterweekdays():
+            table.add_column("{:.3}".format(calendar.day_name[wd]))
+
+        month_days = cal.monthdayscalendar(year, month)
+        for weekdays in month_days:
+            days = []
+            for day in weekdays:
+                today_tuple = today.day, today.month, today.year
+                
+                if day and (day, month, year) == today_tuple :
+                    _day = "[bold underline red]%s[/]" % day   
+                else:
+                    _day = "%s" % ('' if day == 0 else day) 
+                days.append(_day)
+            table.add_row(*days)
+
+        tables.append(table)
+
+    columns = Columns(tables, equal=True, expand=True)
+    print(columns)
+
+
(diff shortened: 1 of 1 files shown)


sureshdsk:
@willmcgugan do I need to add myself to CONTRIBUTORS.md?

examples/columns2.py
```diff
@@ -0,0 +1,56 @@
...
+    today = datetime.today()
+    year = int(year)
+    cal = calendar.Calendar()
+
+    tables = []
+    for month in range(1, 13):
+
+        table = Table(title=f"""{calendar.month_name[month]} {year}""", style="green")
+
+        for wd in cal.iterweekdays():
```
willmcgugan, line 29: Can we call this `week_day`. I have a no abbreviation policy.

examples/columns2.py
```diff
@@ -0,0 +1,56 @@
...
+            table.add_column("{:.3}".format(calendar.day_name[wd]))
+
+        month_days = cal.monthdayscalendar(year, month)
+        for weekdays in month_days:
+            days = []
+            for day in weekdays:
+                today_tuple = today.day, today.month, today.year
+
+                if day and (day, month, year) == today_tuple:
+                    _day = "[bold underline red]%s[/]" % day
```
willmcgugan, line 39: You won't need the `[/]` here.

Can you replace the string interpolation with f strings?

examples/columns2.py
```diff
@@ -0,0 +1,56 @@
...
+                else:
+                    _day = "%s" % ("" if day == 0 else day)
```
willmcgugan, line 41: I don't think you need to interpolate at all here.

How bout this?

```python
_day = "" if day == 0 else str(day)
```

examples/columns2.py
```diff
@@ -0,0 +1,56 @@
...
+def display_calendar(year):
+
+    today = datetime.today()
+    year = int(year)
+    cal = calendar.Calendar()
+
+    tables = []
+    for month in range(1, 13):
+
+        table = Table(title=f"""{calendar.month_name[month]} {year}""", style="green")
```
willmcgugan, line 27: Don't think we need triple quotes here.

Have you run black on this?

willmcgugan (changes requested):
Looks great, a few requests. Could you rename this to `calendar.py`...

sureshdsk:
@willmcgugan calendar.py has name collision with std python calendar module. can we go with calendar2.py or calendar_layout.py ?

examples/columns2.py
sureshdsk: done

examples/columns2.py
sureshdsk: done

examples/columns2.py
sureshdsk: good find.

examples/columns2.py
sureshdsk: changed triple quotes into single quotes. black didn't change this.

willmcgugan:
Thanks!

PR #1544, opened 2021-10-04
Correct typing of Iterables and small narrowing

GBeauregard:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

A lot of the Iterators in the code were typed as Iterable when they were more appropriately typed as an Iterator (i.e they support next). I also narrowed a type a bit.

There were a lot of instances in the code where usage of `Any` can or should be replaced with `object`, but I'm not sure if you have strong opinions on this. If you're open to all instances of more narrowly typing the library I may work on this in my free time.

### DIFF
--- a/rich/_loop.py
+++ b/rich/_loop.py
@@ -1,9 +1,9 @@
-from typing import Iterable, Tuple, TypeVar
+from typing import Iterable, Iterator, Tuple, TypeVar
 
 T = TypeVar("T")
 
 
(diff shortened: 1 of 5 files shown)


willmcgugan:
I don't want to promise that the return type has a `__next__` method, all I want to promise is that I'm returning something that may be iterated over. This allows me to change the return type to some other iterable in the future, safe in the knowledge that code using it won't break.

There may be some room for changing `Any` to `object`. It could conceivably catch bugs in the future.

PR #1551, opened 2021-10-07
Fix all reportPrivateImportUsage in pyright for library users

GBeauregard:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This PR adds `TYPE_CHECKING` guards to the `__main__` sections of all files that imported/redefined part of their own API in this section. This hides the redefinition from type checkers as needed to avoid accidentally signalling the API as private and causing pyright errors for everyone using rich as a library.

Note there are many more `__main__` sections in the library; it was only necessary to add the guard to files importing their own API.

This PR resolves https://github.com/willmcgugan/rich/issues/1523

I didn't update CONTRIBUTORS since the change is minor, but if you have a policy of always updating it let me know and I'll add my name.

### DIFF
--- a/rich/box.py
+++ b/rich/box.py
@@ -430,7 +430,7 @@ LEGACY_WINDOWS_SUBSTITUTIONS = {
 }
 
 
-if __name__ == "__main__":  # pragma: no cover
+if __name__ == "__main__" and not TYPE_CHECKING:  # pragma: no cover
 
     from rich.columns import Columns
     from rich.panel import Panel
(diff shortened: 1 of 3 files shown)


GBeauregard:
I brought up the possibility of excluding `__main__` sections with upstream, but the feeling was that it's really a library issue. The proper `TYPE_CHECKING` guard is simple enough anyway.

willmcgugan:
Thanks. I think I'll go with 2. 

3. Isn't really an option. The main blocks are fairly integral and popular feature.

PR #1586, opened 2021-10-14
fix return type for `Table.expand()` getter

akettmann-e24:
return type for `Table.expand()` was `int` should be `bool` based on the type annotation for `expand` argument for `Table.__init__`

## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description
Fix type annotation for Table.expand getter.


akettmann-e24:
Fixing type annotation, so I didn't add tests, and didn't add to changelog

willmcgugan:
Thanks. I guess Mypy allows for automatic conflation of a bool with an int. Makes sense, but I'd have preferred it to be strict here.

willmcgugan:
Thanks

PR #1600, opened 2021-10-16
fix: Dataclass pretty repr can contain a trailing separator and whitespace

ochilly:
Dataclass pretty repr can contain a trailing separator and whitespace

Fixed pretty_repr() for dataclass fields with repr=False.
Fixed corresponding test.

## Type of changes

- [X] Bug fix
- [ ] New feature
- [X] Tests
- [ ] Other

## Checklist

## Description
Issue: https://github.com/willmcgugan/rich/issues/1599

Changed logic for looping over dataclass fields so now there will always be a child_node marked as last, even if the final field has its repr disabled.

Filter the fields by field.repr is True before passing to loop_last, instead of afterwards.

Modified the corresponding test so the current undesireable behavior would be caught.

### DIFF
--- a/tests/test_pretty.py
+++ b/tests/test_pretty.py
@@ -67,6 +67,7 @@ class ExampleDataclass:
     bar: str
     ignore: int = field(repr=False)
     baz: List[str] = field(default_factory=list)
+    _ignore: int = field(init=False, repr=False)
 
 
 def test_pretty_dataclass():
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -647,11 +647,11 @@ def traverse(
 
             repr_fields = (field for field in fields(obj) if field.repr)
             for last, field in loop_last(repr_fields):
-                    child_node = _traverse(getattr(obj, field.name))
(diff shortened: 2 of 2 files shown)


tests/test_pretty.py
```diff
@@ -67,6 +67,7 @@ class ExampleDataclass:
     bar: str
     ignore: int = field(repr=False)
     baz: List[str] = field(default_factory=list)
+    _ignore: int = field(init=False, repr=False)
```
willmcgugan, line 70: What is this for?

willmcgugan:
Fixed in 10.13.0. THanks.

tests/test_pretty.py
ochilly: This would cause the test to fail if the dataclass pretty repr didn't handle the last field having repr=false properly. Which was the issue, it would have a trailing comma and whitespace.

ochilly: This caused the test to fail without the changes proposed for pretty.py

PR #1602, opened 2021-10-16
feat(bars):  pause and resume methods to progress bars

### LINKED ISSUE #1535: Add pause() and resume() methods to progress bars
A couple problems I'm noticing:
1.  If prompted while a progress bar is in use, the user typing is done at the end of the progress bar. If auto_refresh is True, you cannot even see what you type as you type it. Example output below is with auto_refresh=False
2. When using choices=[] and default="" There are escape characters displayed, so it is pretty un-readable.  

```
What value would you like to enter?:
Playbook running... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% -:--:--foobarbaz
demo action requiring user prompt: The user responded with foobarbaz
Choose a value. [1;35m[Scott/Alex/Sam][0m [1;36m(Scott)[0m:
Playbook running... ━━━━━━━━━━━━━━━━━━━━╺━━━━━━━━━━━━━━━━━━━  50% -:--:--Sam
demo prompt for user choice: The user chose Sam
```

I suspect both of these things are related to the fact that the prompt is occurring within a
`with Progress() as progress:` context

v0lp3:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

I added code for pause and resume progress bars.
Actually _progress.py_ and _progress_bar.py_ pause and resume methods are implemented in different way because Progress use ProgressBar indirectly (render the progress bar as new instance every time):
- *progress.py*: Set a flag in Task object, so the update method not add the advance to `task.completed`.
- *progres_bar.py*: Set a flag `paused` in ProgressBar class and track advance in `paused_delta`. 

Briefly, i couldn't reuse the methods pause and resume of ProgressBar into Progress, but I think that the result is enough good.
 
progress

progress bar

close #1535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [10.12.1] - unreleased
 
+- Added pause and resume methods to progress bars
+
 ### Fixed
 
 - Fixed an edge case bug when console module try to detect if they are in a tty at the end of a pytest run
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -21,3 +21,4 @@ The following people have contributed to the development of Rich:
 - [Clément Robert](https://github.com/neutrinoceros)
 - [Tushar Sadhwani](https://github.com/tusharsadhwani)
 - [Gabriele N. Tornetta](https://github.com/p403n1x87)
+- [Andrea Maugeri](https://github.com/v0lp3)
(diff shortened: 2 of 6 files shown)


v0lp3:
@willmcgugan let me know if there is something wrong

willmcgugan:
This isn't what I meant in the issue.

The pause and resume I had in mind would pause the rendering of the progress bar(s), allowing input() to use stdin. Trickier than it seems because it would have to interact with a threads. Feel free to tackle that if you like.

PR #1604, opened 2021-10-17
Many Miscellaneous Typing Fixes

GBeauregard:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

This PR fixes almost the entirety of the remaining bugs that would prevent adding pyright to rich's CI if desired. All of these changes are genuine typing bugs that just aren't flagged by `mypy --strict` due to limitations in its design or particular quirks with how it works. Some of these changes are especially subtle, but I've included detailed explanations in the commit message for each change.

Are you interested in adding pyright to the CI? It's not necessary, but going through it has helped me find a number of genuinely incorrect things in the codebase.

The remaining 10 errors and 3 warnings thrown by pyright when run on the codebase (but not addressed by this PR and left for possible future work) fall into one of these categories:
- 4 errors: these are normal typing errors that are errors when using `mypy` too, but they're incidentally hidden due to the [project configuration](https://github.com/willmcgugan/rich/blob/eb673d1204340738d3084ebc2e4c789a35a4e49b/mypy.ini#L3) and CI not having the `types-pygments` package installed. If you install `types-pygments` locally and run `mypy --strict` on the codebase you will see the errors. At their surface they seem like they could be indicative of genuine bugs, but I haven't had a chance to look into them since validating the errors would require me learning some internals of the pygments package beyond the scope of this PR.
- 1 error: There is a `padding` property in `table.py` where the getter/setter don't have symmetric types; pyright is set to lint this asymmetry by default and throw an error. I've inspected the code in question and there is no bug or typing error; this is just a stylistic complaint from pyright. I see no way to change this without breaking rich's public API so an ignore for this error would need to be added to the file or project if pyright was added to CI.
- 5 errors: A variable is possibly unbound: these occur in cases where there is code that uses the iterating variable `x` after a `for x in y` has closed, but without setting a default value for `x`. I wasn't sure how you'd feel about merging code to work around this type of error. Leave it? Add code to init `x`? Look at it to see if it can be refactored?
- 3 warnings: `warning: TypeVar "DefaultType" appears only once in generic function signature`: These occur in the PromptBase class where the class is not a generic of `DefaultType`, but nonetheless uses these variables alone in class methods. I've inspected the methods in question and the `DefaultType` in these cases can safely be replaced with `object`; another possibility is to make the class a proper generic of `DefaultType` too. However, this latter option would be a bit awkward for usage. These warnings are addressed with #1607

I still intend to at some point make a pass on the codebase narrowing `Any` to `object` where possible, but I'd like to address the existing typing stuff first.

### DIFF
--- a/rich/_lru_cache.py
+++ b/rich/_lru_cache.py
@@ -1,6 +1,5 @@
 from collections import OrderedDict
-from typing import Dict, Generic, TypeVar
-
+from typing import Generic, TypeVar
 
 CacheKey = TypeVar("CacheKey")
 CacheValue = TypeVar("CacheValue")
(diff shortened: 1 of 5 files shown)


GBeauregard:
Branch rebased with the commit removed. Apologies, I wasn't intending to make any backward-incompatible changes there and should have tested as carefully as I did the other things.

GBeauregard:
I think I have a good handle now of typed descriptors, their interactions with the project, and what the best approach for us is here. In particular, typing the descriptor to indicate it can only belong to the class we intend to use it with obviates the need to make separate getter methods or pepper the implementation with excessive casts. I'm going to open a new PR with it when it's ready.

GBeauregard:
The `pick_bool` function seems to be only used as `A ?? B ?? False` if PEP 505 is ever accepted, or `A ?? bool(B)` I guess.

Issue #1528, opened 2021-09-28
[REQUEST] Progress for generators with unknown length


SeanNaren:
Related https://github.com/willmcgugan/rich/issues/1022

Hey @willmcgugan!

We're trying to make Rich the main progress bar in PyTorch Lightning, superseding TQDM: https://github.com/PyTorchLightning/pytorch-lightning/pull/9647

However we noticed that generators are not supported for the progress object, which means IterableDatasets cannot be supported (a common use case for PyTorch). Do you have any suggestions on something that could replace this and give atleast some iteration progress or know if support for generators is possible?

willmcgugan:
Generators absolutely are supported by Progress.

However, there is no way to get the total number of steps from a generator. Without it, theres no way to calculate total progress, time remaining etc. In this case you would need to specify the `total` manually.

So what would you want Rich to do when total is not available?

SeanNaren:
thanks @willmcgugan! appreciate you and this awesome library ❤️  

we came up with a solution, but maybe you have ideas to deal with my hacky lil solution :) the goal is to support an unknown total, as it cannot be defined

Currently it looks like this now:
![demo](https://user-images.githubusercontent.com/6707363/135078029-1e12d95b-a44b-476f-a788-c91b2dbf7b73.gif)

It involved overriding a few methods of a couple of components to support `total=float('inf')` as an input. These can be seen here:

https://github.com/PyTorchLightning/pytorch-lightning/pull/9647/files#diff-1ed8aa90c90b04a74eeb942b0ad5483dff76b05a7de68787495dddbc6aaa40c0R29-R93

Overall it seems to work great, but I wonder if you have any feedback!

jnoortheen:
> So what would you want Rich to do when total is not available?

`rich` can follow the `tqdm`'s approach. When total is not available, it can show the number of tasks executed along with the time it spent, but hiding progress bar and remaining time.

willmcgugan:
Dupe of #1054

Issue #1054, opened 2021-02-24
[REQUEST] Progress bars with undetermined 'total'


azmeuk:
Hi. The documentation describes the usage of [indeterminate progress bars](https://rich.readthedocs.io/en/latest/progress.html#indeterminate-progress) that displays a waiting animation during the time the number of steps is computed.

![indeterminate state](https://media.moyaproject.com/willmcgugan/uploads/thumbnails/techblog/788d9cb6-9e8f-11ea-b769-f23c91845b44.png/788d9cb6-9e8f-11ea-b769-f23c91845b44png.lg.3.jpeg)

However, sometimes you never know how many steps you except (when reading a stream for instance), but this is something rich does not accept:
https://github.com/willmcgugan/rich/blob/1c1d72ccad965a41500253680c73e6606924cd13/rich/progress.py#L667-L673

I suggest in those cases to display a waiting animation instead.

Of course in this case the percentage and the ETA would not be very usable, but the bar could display the number of items processed (this is what tqdm does).

What do you think?

willmcgugan:
Would a [status](https://rich.readthedocs.io/en/latest/console.html#status) not be more appropriate for that?

azmeuk:
Status can do the job if you know the *thing* you iterate on has always an undetermined size. Sometimes you may want to use a progress bar without knowing in advance if the total size could be determined.

geoffsmith:
Just to note a few things that you'd need to do manually with a status that could perhaps be done more cleanly with a progress bar: elapsed time so far, number of units complete so far and the rate of units. TQDM is a nice example of this, if you just leave out a `total` (really enjoying rich so far btw - really lovely library 😄 )

Daddy2200:
> Hi. The documentation describes the usage of [indeterminate progress bars](https://rich.readthedocs.io/en/latest/progress.html#indeterminate-progress) that displays a waiting animation during the time the number of steps is computed.
> 
> 
> 
> ![indeterminate state](https://media.moyaproject.com/willmcgugan/uploads/thumbnails/techblog/788d9cb6-9e8f-11ea-b769-f23c91845b44.png/788d9cb6-9e8f-11ea-b769-f23c91845b44png.lg.3.jpeg)
> 
> 
> 
> However, sometimes you never know how many steps you except (when reading a stream for instance), but this is something rich does not accept:
> 
[...]

PR #1609, opened 2021-10-19
Indeterminate progress tracking

mrvkino:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

A task tracking the progress of an indeterminate process can be done by adding `float("inf")` as a value to the `total` attribute of the `progress.add_task` method or `track` function.

This will display a pulsing progress bar. Also, columns can be updated as normal with calls to `progress.update`. The number of element completed and the time elapse will work as expected. The value for the total in columns will be `?` and any completion percentage will be calculated as `0%` or equivalent.

The new progress bar has been added in the example provided by the module `rich.progress`.


mrvkino:
@azmeuk I have added the option of tracking a sequence of indeterminate length when the total is `None`. This will reuse the `float("inf")` underneath.

This should work:
```python
from itertools import cycle
from rich.progress import track

test = ["foo", "bar", "baz"]
for value in track(cycle(test), total=None):
    print(value)
```

mrvkino:
Whit this latest round of commit. It should wrap all request for indeterminate progress tracking.

close #1054 and close #1528

willmcgugan:
I'm afraid I can't accept the `float("inf")`. A significant number of people will find this surprising.

mrvkino:
Do you want to do a complete removal (internally also)?

We would only support a total of `None`. I will get on it.

mrvkino:
Updated the code to remove the `float("inf")` support.
For specific case, I convert the `None` value to `float("inf")` internally to simplify check and calculation.

If you have any comment, it will be my pleasure to change the code.

mrvkino:
@willmcgugan, I updated the code fixed some typing issue that arose with my code and the latest commits/merges. It should be ready for review.

willmcgugan:
We now support total=None for indeterminate progress bars.

PR #1612, opened 2021-10-19
moved properties w/ asymmetric interfaces to descriptors

GBeauregard:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

This PR implements the descriptor workaround discussed in #1604. With this change we should no longer be breaking downstream mypy users, and we incidentally fix the pyright lint as well. I've tried to take a minimally disruptive and highly maintainable approach. With this change I also keep the autodoc generated with the same docstrings.

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -30,6 +30,7 @@ from typing import (
     Type,
     Union,
     cast,
+    overload,
 )
 
 if sys.version_info >= (3, 8):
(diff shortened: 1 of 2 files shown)


GBeauregard:
There is some hot off the press news on asymmetric property interfaces.

Guido says his views have evolved on them and has convinced pyright maintainer to support them https://github.com/microsoft/pyright/issues/2626

See https://github.com/python/mypy/issues/3004 (particularly the discussion from the past day). This presumably means mypy will be changing to support this pattern as well, but neither change has happened yet.

GBeauregard:
Note this PR is intended to fix *consumers* of rich's API and the fix on the library side was incidental. Consumers were already unbroken in pyright and being broken in mypy is still a bug we need to wait to be fixed.

PR #1621, opened 2021-10-24
allow for callable in rich.traceback.install(suppress=…)

jettero:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

I think this suite of readability improvements is just great. <3 ... but rich lacks the thing I was actually looking for when I found it. I was actually looking for a module that would limit the stack traces to code I actually wrote so I don't have to scroll back through 15 pages of stuff I didn't write to get to the actually relevant error on the lines I did write.

So, while `rich.traceback.install(suppress=[])` _almost_ does what I want, there was no way to actually achieve my goal without really annoying acrobatics; since, what I really want is `rich.traceback.install(suppress=[lambda fname: recognize_my_own_modules(fname)])`

This PR adds callable suppress_entity support. It's already doing what I want in my project by simply installing my modified version; but I doubt I'm the only person that would want this feature so I cleaned it up for sharing.

### DIFF
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -243,7 +243,7 @@ class Traceback:
         self.locals_max_length = locals_max_length
         self.locals_max_string = locals_max_string
 
-        self.suppress: Sequence[str, callable] = []
+        self.suppress: Sequence[str, Callable] = []
         for suppress_entity in suppress:
             if callable(suppress_entity):
                 self.suppress.append(suppress_entity)
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -65,6 +65,8 @@ def install(
         show_locals (bool, optional): Enable display of local variables. Defaults to False.
         indent_guides (bool, optional): Enable indent guides in code and locals. Defaults to True.
         suppress (Sequence[Union[str, ModuleType]]): Optional sequence of modules or paths to exclude from traceback.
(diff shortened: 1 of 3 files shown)


rich/traceback.py
```diff
@@ -243,8 +243,11 @@ def __init__(
         self.locals_max_length = locals_max_length
         self.locals_max_string = locals_max_string
 
-        self.suppress: Sequence[str] = []
+        self.suppress: Sequence[str, callable] = []
```
willmcgugan, line 246: Suspect his should be `Union[str, Callable]` (note Callable imported from the typing module).

willmcgugan (changes requested):

willmcgugan:
Docs and tests please.

rich/traceback.py
jettero: ok. will do.  FYI, I got that lower case spelling from a doc entry in the same file under from_exception.

jettero:
> Docs and tests please.

I did update a test to show callables work. It's not like extensive testing or anything, but it does show it works imo.

I thought I had added docs too, but apparently not. Here's another commit.

willmcgugan:
Closing. Assumed stale.

PR #1634, opened 2021-10-31
Mention the usage of Python's builtin readline module in the documentation

### LINKED ISSUE #1633: [REQUEST] Add instructions about readline in the docs
- [x] Have you checked the issues for similar suggestions.

I believe the docs (and the community in general) would benefit a **lot** from having some basic guidelines on how to use the standard [readline module](https://docs.python.org/3/library/readline.html) with Rich. It's a single import after all.

Without `readline` (weird stuff are the result of arrow keys):
withoutreadline

With `readline`:
withreadline

It turns out there's no need to change a single line in Rich, just mention it in the docs (such as [here](https://rich.readthedocs.io/en/latest/console.html?highlight=input#input)):

```python
import readline  # just add this
console = Console()
console.input("What is [i]your[/i] [bold red]name[/]? :smiley: ")
```

The reason I think it is useful is because
1. it might not be known by beginners (I didn't!)
2. it's [nowhere to be found in the docs](https://rich.readthedocs.io/en/latest/search.html?q=readline&check_keywords=yes&area=default)
3. it's dead easy (just `import readline`)
4. similar stuff seems to have been requested before (e.g. #306), but without any definite answer

schneiderfelipe:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [x] ~I've added tests for new code.~ No new code.

## Description

This attempts to resolve #1633.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -43,6 +43,7 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.viewcode",
     "sphinx.ext.napoleon",
+    "sphinx.ext.intersphinx",
     "sphinx.ext.autosectionlabel",
     "sphinx_copybutton",
 ]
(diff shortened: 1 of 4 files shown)


willmcgugan:
Thanks

PR #1637, opened 2021-11-01
Have logger link to proper file line.

Carreau:
Some terminal emulators like iterm2, support anchoring to the line
number:

https://iterm2.com/documentation-one-page.html see Anchor (OSC 8)
section:

     Note: in iTerm2 version 3.4 and later, if the URL has the file scheme and
     a # fragment is present then the semantic history rules will apply
     for opening the file. It may optionally include a line number, like
     file:///tmp/file.txt#123 or line number and column number like
     file:///tmp/file.txt#123:45

This update the logic in the logger so that the line number links the
the corresponding file on the same line.

As I'm not sure whether all terminals support that, I let the  clik on
the file name point to the file itself w/o line numbers.

And for discoverability and user friendliness the colon itself is
unstyled so that it's easier to discover that click on the  filename and
line number are two different targets.

## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

I'm not sure it is worth documenting as it is a minor change, and not sure for tests, as it's mostly external software interaction.

## Checklist

### DIFF
--- a/rich/_log_render.py
+++ b/rich/_log_render.py
@@ -75,7 +75,11 @@ class LogRender:
                 path, style=f"link file://{link_path}" if link_path else ""
             )
             if line_no:
-                path_text.append(f":{line_no}")
+                path_text.append(":")
+                path_text.append(
+                    f"{line_no}",
+                    style=f"link file://{link_path}#{line_no}" if link_path else "",
+                )
             row.append(path_text)
 
         output.add_row(*row)
--- a/tests/test_log.py
+++ b/tests/test_log.py
@@ -6,8 +6,7 @@ import re
 
 from rich.console import Console
 
-
-re_link_ids = re.compile(r"id=[\d\.\-]*?;.*?\x1b")
+re_link_ids = re.compile(r"id=((\d+);(file:///(\w+\/)+(\w+\.py)(#\d+)?))\x1b")
 
 
 def replace_link_ids(render: str) -> str:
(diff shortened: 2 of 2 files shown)


willmcgugan:
That seems like a nice feature. Any chance of a screenshot and  a test?

Carreau:
Pushed updates to tests.

Screenshot is not going to change much, here is `python examples/log.py` the only difference is that the lines number are now underlined and clickable.

Screen Shot 2021-11-06 at 12 22 48 PM

while they are not on the main branch:

Screen Shot 2021-11-06 at 12 24 58 PM

You can see the screen recording of hovering the links maybe ?

https://user-images.githubusercontent.com/335567/140621550-2f626684-721e-4ea3-ad5b-23eb73293a9f.mov

willmcgugan:
I think this is a nice addition. I would like to test it on a number of terminals to see what they do with the line number.

If it doesn't break other terminals, I think it may be worth making the entire thing link to the line number.

willmcgugan:
Just a black format required...

willmcgugan:
I'll sort the formatting. Thanks!

Carreau:
> If it doesn't break other terminals, I think it may be worth making the entire thing link to the line number.

My understanding is that it's not standard, so I was careful.

> I'll sort the formatting. Thanks!

hum, thanks, I don't know what's wrong on my end , i'm supposed to have reformat as a save hook in my editor.

willmcgugan:
It's in 10.13.0 If there aren't any problems reported I'll make the line number span the path in 10.14.0

I've tried it on a number of terminals, and it seems solid.

Issue #1670, opened 2021-11-09
[BUG] Text objects pre-formatted with ANSI escapes miscalculate their width


0xDEC0DE:
When attempting to use "pre-formatted" strings from a third-party library with Rich tables, it seems to use the "whole width" of the string for header calculation rather than the "visible width"

Steps to reproduce
==============

A simple test case, I'm using [colored](https://pypi.org/project/colored/) to illustrate:

```
from colored import bg, fg, attr

from rich.box import ASCII
from rich.console import Console
from rich.table import Table
from rich.text import Text

# simulate getting a "cooked" string from a library
text = bg(28) + fg(186) + 'this is test text' + attr(0)

table = Table(box=ASCII)
table.add_column('column')
table.add_row(Text(text))

Console().print(table)
```

Expected result
============

```
+-------------------+
| column            |
|-------------------|
| this is test text |
+-------------------+
```

Actual behavior
===========

```
+-----------------------------------------+
| column                                  |
|-----------------------------------------|
| this is test text |
+-----------------------------------------+
```

Errata
=====

I see class methods like `Text.from_markup` and `Text.styled` that would appear to be useful for creating `Text` objects from "cooked" inputs, but they don't appear to be made for this particular flavor.  And attempting to adjust the `Text._length` property implicitly truncates the string.

Platform
======
```
╭───────────────────────── <class 'rich.console.Console'> ─────────────────────────╮
│ A high level console interface.                                                  │
│                                                                                  │
│ ╭──────────────────────────────────────────────────────────────────────────────╮ │
│ │ <console width=120 ColorSystem.TRUECOLOR>                                    │ │
│ ╰──────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                  │
│     color_system = 'truecolor'                                                   │
│         encoding = 'utf-8'                                                       │
│             file = <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'> │
│           height = 77                                                            │
│    is_alt_screen = False                                                         │
│ is_dumb_terminal = False                                                         │
│   is_interactive = True                                                          │
│       is_jupyter = False                                                         │
│      is_terminal = True                                                          │
│   legacy_windows = False                                                         │
│         no_color = False                                                         │
│          options = ConsoleOptions(                                               │
│                        size=ConsoleDimensions(width=120, height=77),             │
│                        legacy_windows=False,                                     │
│                        min_width=1,                                              │
│                        max_width=120,                                            │
[...]

PR #1675, opened 2021-11-10
Add a Text.from_ansi helper method

0xDEC0DE:
Add a simple little helper to run `AnsiDecoder.decode_line` over "pre-cooked" inputs.

Fixes issue: #1670

## Type of changes

- [x] Bug fix
- [x] New feature
- [x] Tests

## Checklist

### DIFF
--- a/rich/text.py
+++ b/rich/text.py
@@ -243,6 +243,28 @@ class Text(JupyterMixin):
         return rendered_text
 
     @classmethod
+    def from_ansi(
+        cls,
+        text: str,
+        *,
+        justify: Optional["JustifyMethod"] = None,
+        overflow: Optional["OverflowMethod"] = None,
+    ) -> "Text":
+        """Create a Text object from pre-formatted ANSI.
+
+        Args:
+            text (str): A string containing ANSI color codes.
+            justify (str, optional): Justify method: "left", "center", "full", "right". Defaults to None.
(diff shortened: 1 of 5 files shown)


rich/text.py
```diff
@@ -242,6 +242,28 @@ def from_markup(
         rendered_text.overflow = overflow
         return rendered_text
 
+    @classmethod
+    def from_ansi(
+        cls,
+        text: str,
+        *,
+        justify: Optional["JustifyMethod"] = None,
+        overflow: Optional["OverflowMethod"] = None,
```
willmcgugan, line 251: Could you copy the other arguments from the Text constructor, i.e. style, no_wrap, end, and tab_size (no need for spans).

willmcgugan (changes requested):
I think this is a good idea. And a nice way of exposing that ANSIDecoder class. Just one request.

rich/text.py
0xDEC0DE: If the `style=` kwarg is provided, what should that do, exactly?  Add to the existing styling, or replace it?

willmcgugan: Set it in the returned Text instance, it defines the base style.

0xDEC0DE:
Pushing changes to my branch isn't updating this pull request.

Replacing this PR with #1706

PR #1705, opened 2021-11-17
Add support for showing threads with RichHandler

hemna:
This patch adds the ability to optionally show the thread name
with the RichHandler logger.   This is quite useful for python
applications that have multiple threads and trying to make sense
of the log output.

The thread name is colorized to a color which is created from
a simple hash of the thread name itself.  So a thread name of 'Main'
will always have the same color, but another thread say "socket"
will have a different color.

## Type of changes

- [ ] Bug fix
- [X] New feature
- [X] Tests
- [ ] Other

## Checklist

## Description

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added file protocol to URL highlighter https://github.com/willmcgugan/rich/issues/1681
 - Added rich.protocol.rich_cast
+- Added show_thread to RichHandler to enable showing thread names in logs
 
 ### Changed
 
--- a/rich/_log_render.py
+++ b/rich/_log_render.py
@@ -20,6 +20,8 @@ class LogRender:
         time_format: Union[str, FormatTimeCallable] = "[%x %X]",
         omit_repeated_times: bool = True,
         level_width: Optional[int] = 8,
+        show_thread: bool = False,
+        thread_width: Optional[int] = 10,
     ) -> None:
         self.show_time = show_time
         self.show_level = show_level
@@ -28,6 +30,8 @@ class LogRender:
         self.omit_repeated_times = omit_repeated_times
         self.level_width = level_width
         self._last_time: Optional[Text] = None
+        self.show_thread = show_thread
+        self.thread_width = thread_width
 
     def __call__(
         self,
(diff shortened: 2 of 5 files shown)


willmcgugan:
How can you be sure that it won't generate unreadable colors on a given background?

hemna:
> How can you be sure that it won't generate unreadable colors on a given background?

actually, no, but I am open to suggestions.   In my testing one of my thread names was dark blue on my black background and it was hard to read, but so is the color for INFO level debug level, which is also dark blue.

Screen Shot 2021-11-17 at 8 33 37 PM

hemna:
I removed the dynamic colorization of the thread name and default it to yellow now.  

I have been trying to see if there was a built in mechanism to override that style with my own style class.  That way my app that uses this feature could just implement the dynamic color itself via a theme.   From what I can tell there is no way to get the text from the style object, or to override the style object.

I have this working, but it still only forces an override to 1 particular color

```
cs = Console(theme=Theme({'log.thread_name': "red"}))
rh = RichHandler(rich_tracebacks=True, omit_repeated_times=False,
                         show_thread=True, thread_width=10, console=cs)
...
```

That code overrides the yellow color to red.  But I still want the color based on the actual thread_name text itself.

hemna:
I added the missing doc strings and 2 unit tests

willmcgugan:
I'm afraid I don't think this is a good. I can see why you might want it, but it's not generally useful enough to go in the core lib. For the moment, I would suggest just copying your custom RichHandler to your own project.

In the future, maybe RichHandler will grow some way of customizing columns in a more data-driven way.

hemna:
showing the thread is not good?  how do you figure that?

hemna:
how do you figure not being able to show the thread is 'not good' ?

willmcgugan:
I meant to say "good idea"

hemna:
So, it's not a good idea to be able to log which thread a log entry is coming from in a multi threaded app.   SMH

willmcgugan:
It's not a good idea to add one specific use case to the core library. I can't satisfy every use case, so I need to draw a line somewhere, to keep the interface manageable.

I'm not preventing you from achieving what you need, since you clearly know how to extend the RichHandler.

hemna:
Fair enough.  I guess this is one of the downsides to the current RichHandler not being able to be configurable in the format of the output like the standard logging that comes with python.

If rich had the ability to use the log formatter text

```
formatter = logging.Formatter("[%(asctime)s] [%(threadName)-20.20s] [%(levelname)-5.5s] %(message)s - [%(pathname)s:%(lineno)d]")
```

Then one could simple add the threadName in there like that and it would just show up.  

Sorry if I'm being a pain, I just thought it would be a good idea to enable a standard python idiom to the log output.

Issue #1725, opened 2021-11-28
[Bug/feature] Filename detection poor for paths/filenames with spaces


aaronrudkin:
rich's automatic highlighting for filenames is generally quite useful, but it tends to fail when the filename has a space in it. The highlighting breaks on the space.

```
my_filename = "/folder/file name.txt"
console.log(f"Scanning file {my_filename}")
```

This seems like a fairly tricky thing to get right, because obviously there's some ambiguity here that makes the grammar non-regular. For instance, you could image the string "/folder/file name.txt" could be either a folder named "/folder/file" and then an unrelated string "name.txt" or it could all be a filename.  I had a look at how rich.highlighter seems to operate and unsurprisingly it's just a battery of regexes.

I could think of two possible partial solutions here. The first would be to allow for a particular magic string that could surround a variable to indicate the entire variable should be parsed as a single syntax highlight, e.g. `console.log(f"Scanning file ||{my_filename}||")` or some such. The second would be -- is it possible to overload the default f-string processing (I don't know that I have a clear understanding of how f-strings are substituted at runtime to know if this is possible) to be able to make assumptions there.

Maybe these sorts of solutions add too much cruft to be worthwhile, but I figured I'd still raise the issue in case you have a cleverer sense than I do.

(Really love the package, I've converted a bunch of stuff to use it recently and am just totally impressed!)

nathanrpage97:
ForyYour example, rich has no way of differentiating between two cases:

1. It is all one file path as you describe
2. `/folder/file` is a path and `name.txt` is a separate file name

You can explicitly provide highlighting like so:

```python
my_filename = "/folder/file name.txt"
console.log(f"Scanning file [repr.path]{my_filename}[/repr.path]")
```

willmcgugan:
@aaronrudkin I'm afraid there is no way to hook in to the f string interpolation. @nathanrpage97 's solution is good, but note that you should escape the filename, if there is a chance it will contain square brackets.

```python
from rich.markup import escape
```

PR #1730, opened 2021-11-30
Docs add default theme appendix

nathanrpage97:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Add a reference to all of the default theme stylings available. Useful for cases such as #1725.


willmcgugan:
The defaults styles will change from version to version. I don't want the burden of updating that HTML in the repos.

I like the default_styles command though. How about we reference that command, but don't add the html to the appendix?

willmcgugan (approved):

willmcgugan:
Thanks

PR #1748, opened 2021-12-09
Add lexer parameter to Syntax

patrick91:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [ ] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate. **Will do this when the PR has been reviewed**

## Description

I was working on some code that needed to show highlighted code and I needed to use a lexer instance as I was using some custom Pygment filters and I noticed that rich doesn't allow to pass an instance of lexer, so I decided to see if I can add support for that 😊

I'm not 100% happy with having both `lexer` and `lexer_name` as arguments, but that's the easiest way to maintain compatibilty. Another option is rename `lexer_name` to `lexer` and accept `Union[str, Lexer]`, but that would be a breaking change (which might be ok), let me know what you think 😊

Related discussion: https://github.com/willmcgugan/rich/discussions/985

### DIFF
--- a/rich/syntax.py
+++ b/rich/syntax.py
@@ -5,6 +5,7 @@ import textwrap
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Iterable, List, Optional, Set, Tuple, Type, Union
 
+from pygments.lexer import Lexer
 from pygments.lexers import get_lexer_by_name, guess_lexer_for_filename
 from pygments.style import Style as PygmentsStyle
 from pygments.styles import get_style_by_name
@@ -195,6 +196,7 @@ class Syntax(JupyterMixin):
     Args:
         code (str): Code to highlight.
         lexer_name (str): Lexer to use (see https://pygments.org/docs/lexers/)
+        lexer (Lexer): Lexer instance to use (see https://pygments.org/docs/lexers/)
         theme (str, optional): Color theme, aka Pygments style (see https://pygments.org/docs/styles/#getting-a-list-of-available-styles). Defaults to "monokai".
         dedent (bool, optional): Enable stripping of initial whitespace. Defaults to False.
         line_numbers (bool, optional): Enable rendering of line numbers. Defaults to False.
@@ -226,8 +228,9 @@ class Syntax(JupyterMixin):
     def __init__(
         self,
         code: str,
-        lexer_name: str,
+        lexer_name: Optional[str] = None,
         *,
+        lexer: Optional[Lexer] = None,
         theme: Union[str, SyntaxTheme] = DEFAULT_THEME,
         dedent: bool = False,
         line_numbers: bool = False,
@@ -240,8 +243,12 @@ class Syntax(JupyterMixin):
         background_color: Optional[str] = None,
         indent_guides: bool = False,
     ) -> None:
+        if all((lexer_name, lexer)):
+            raise ValueError("Cannot pass both lexer_name and lexer")
+
         self.code = code
         self.lexer_name = lexer_name
+        self.lexer = lexer
         self.dedent = dedent
         self.line_numbers = line_numbers
         self.start_line = start_line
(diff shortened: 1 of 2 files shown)


tests/test_syntax.py
```diff
@@ -74,6 +89,21 @@ def test_python_render_simple():
...
+    syntax = Syntax(
+        CODE,
+        lexer=PythonLexer(),
+        line_numbers=False,
+        theme="foo",
+        code_width=60,
+        word_wrap=False,
+    )
+    rendered_syntax = render(syntax)
+    print(repr(rendered_syntax))
```
adamchainz, line 102:
```suggestion
```

?

patrick91: uh, good spot :D I've copied this from the test below! maybe @willmcgugan knows if we should remove it or not (if so I can remove it from all the tests in this file)

rich/syntax.py
```diff
@@ -240,8 +243,12 @@ def __init__(
         background_color: Optional[str] = None,
         indent_guides: bool = False,
     ) -> None:
+        if all((lexer_name, lexer)):
```
willmcgugan, line 246: I think `lexer_name and lexer` probably expresses intent better.

rich/syntax.py
```diff
@@ -374,8 +381,8 @@ def highlight(
         )
         _get_theme_style = self._theme.get_style_for_token
         try:
-            lexer = get_lexer_by_name(
-                self.lexer_name,
+            lexer = self.lexer or get_lexer_by_name(
```
willmcgugan, line 384: Breaking this in to an `if` rather than an expression would allow test coverage to span both branches, and avoid that cast.

tests/test_syntax.py
willmcgugan: I sometimes leave those in tests, so that pytest shows the full output if there is a failure.

I don't mind either way.

willmcgugan (changes requested):
Looks good. A couple of minor requests.

tests/test_syntax.py
patrick91: makes sense! I'll leave it then 😊

rich/syntax.py
patrick91: I kinda prefer the previous version 😊 ideally the if would live outside the try/except, but there's logic using the lexer inside the else statement, so I've kept it there. We still need a cast (or an asser as I've done now) in this case :)

willmcgugan:
Just thinking this through.

I think your comment about not being 100% happy with having `lexer` and `lexer_name` was a good instinct. It would be a better interface to have a single `lexer` argument that may be a str or a Lexer instance.

That could be construed as a breaking change, but it was always intended to be a positional argument. So it might warrant a major version bump.

Let's go with the single `lexer` argument. I'll decide wether it warrants a major version bump.

patrick91:
> Just thinking this through.
> 
> I think your comment about not being 100% happy with having `lexer` and `lexer_name` was a good instinct. It would be a better interface to have a single `lexer` argument that may be a str or a Lexer instance.

I think this looks much better! Let me know what you think 😊

rich/syntax.py
```diff
@@ -348,6 +349,20 @@ def _get_token_color(self, token_type: TokenType) -> Optional[Color]:
         style = self._theme.get_style_for_token(token_type)
         return style.color
 
+    @property
+    def lexer(self) -> Optional[Lexer]:
```
willmcgugan, line 353: Docstring please!

willmcgugan (changes requested):
Great. Just a request for a docstring and this is good to go.

rich/syntax.py
patrick91: done, hope it is good!

willmcgugan:
Thanks!

Issue #1749, opened 2021-12-10
[BUG] rich/examples/downloader.py does not wait for downloading


hankyul2:
You may find a solution to your problem in the [docs](https://rich.readthedocs.io/en/latest/introduction.html) or [issues](https://github.com/willmcgugan/rich/issues).

**Describe the bug**

In `rich/example/downloader.py`, process does not wait for downloading file done. It just exit within 1 second.

```
python3 url.py https://releases.ubuntu.com/20.04/ubuntu-20.04.2.0-desktop-amd64.iso
```
![image](https://user-images.githubusercontent.com/31476895/145548548-cb47b49d-b512-4b4a-9994-52df1ddcc472.png)

**Platform**

I am using linux 18.04, python 3.7.10

![image](https://user-images.githubusercontent.com/31476895/145548965-373940ed-a14d-4413-837f-a25220af10fa.png)
![image](https://user-images.githubusercontent.com/31476895/145549038-eee976f9-7961-45df-befc-8e6a1aba3391.png)

sanders41:
The problem is the url `https://releases.ubuntu.com/20.04/ubuntu-20.04.2.0-desktop-amd64.iso` returns a 404. Try `https://releases.ubuntu.com/20.04/ubuntu-20.04.3-desktop-amd64.iso` instead.

hankyul2:
@sanders41  Thank you. It works
![image](https://user-images.githubusercontent.com/31476895/145657747-ff318e7c-6c8e-4a9d-8c48-ebc5226edc12.png)

PR #1750, opened 2021-12-11
Update example url

sanders41:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Related to #1749

The URL specified in the doc string as an example to try no longer exists and returns a 404. I updated it to the current working URL.

### DIFF
--- a/examples/downloader.py
+++ b/examples/downloader.py
@@ -73,7 +73,7 @@ def download(urls: Iterable[str], dest_dir: str):
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


willmcgugan:
This isn't the first time this has happened. I wonder if there is a better URL to suggest that won't change?

sanders41:
I thought the same about the URL changing again. I tried to think of something that would be big enough to show the progress and not change, but so far haven’t come up with anything.

Issue #1776, opened 2021-12-23
[REQUEST] option to respond to choices with digits


tlambert03:
**How would you improve Rich?**

related to #306, It would be nice to make it easier to pick from a variety of annoying-to-type choices:

```py
choices = ['Some really long and annoying thing to type', 'Another really long...']
Prompt.ask('pick one', choices=choices, with_digits=True)
```

could prompt as follows (or something like this ... where the valid responses are now digits)

```sh
[1] Some really long and annoying thing to type
[2] Another really long...
pick one [1/2]: 
```

**What problem does it solve for you?**

makes it much easier to request categorical values

in  #306, `readline` was mentioned, but I didn't follow, so let me know if there's already a better way to do this. 
thanks for rich!!

TheTrio:
So I don't know if I should've done this lol but I had a [stab](https://github.com/TheTrio/rich/tree/promptDigits) at this. (First time doing this :eyes:)

tlambert03:
nice!  just tried it out, and it works for me.  exactly what I was looking for.

only thought I had (and i'm not sure about this) is whether it should be possible to use _either_ the number or the text.  So, given the prompt
```
[1] choice a
[2] choice b
pick_one (1): 
```
entering _either_ "1" or "choice a" would be valid and give "choice a", etc...

that's more of a stylistic API choice that's up to @willmcgugan

but this would be great for me.  maybe make a PR and see if it's ok with @willmcgugan?

TheTrio:
> only thought I had (and i'm not sure about this) is whether it should be possible to use either the number or the text. 

Yes, that could be an option. 

> but this would be great for me. maybe make a PR and see if it's ok with @willmcgugan?

I was just waiting for him to green-light the idea before we got into implementation details of the PR. Not really sure of how this all is supposed to work lol.

tlambert03:
> I was just waiting for him to green-light the idea before we got into implementation details of the PR.

opening a PR is a very reasonable thing to do, even if you're just proposing something :)  It will make it easier for him to directly comment on your implementation, to voice concerns or request changes inline.  If he's opposed, it's trivial to say thanks but no thanks and close it :)

tlambert03:
side note: there's a couple places in the code where you need to check whether `self.choices` is not None (e.g. before indexing into it with `self.choices.index`).  not sure if you have a linter, but it will show up for sure with mypy/pylance/etc...  thanks again!

TheTrio:
[...]

PR #1779, opened 2021-12-24
Add support for indices in Prompt

TheTrio:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Addresses #1776 

Adds a new parameter `show_digits` which when set to `True` (`False` by default), allows the user to select by the item's index rather than its text. 

For example

```
from rich.prompt import Prompt

choices = ["Some really long and annoying thing to type", "Another really long..."]
print(
    Prompt.ask(
        "pick one", choices=choices, show_digits=True, default="Another really long..."
    )
)
```

has the following output 

image

Currently, if `show_digits` is set to `True`, only selections based on the index are allowed.

### DIFF
--- a/rich/prompt.py
+++ b/rich/prompt.py
@@ -178,7 +178,7 @@ class PromptBase(Generic[PromptType]):
             prompt.append(choices, "prompt.choices")
         elif self.show_choices and self.choices and self.show_digits:
             prompt = Text()
-            _choices_list = []
+            _choices_list: List[str] = []
             for index, choice in enumerate(self.choices):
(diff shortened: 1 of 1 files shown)


willmcgugan:
Sorry, I don't think this offers much over the current functionality. You can display a list of things in the prompt. If you really need it, a subclass is probably the way to go.

coeur85:
i like the idea, the user can simply type the choice number instead of typing the hole word

PR #1781, opened 2021-12-27
Improve installation snippets

MaxG87:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

I found that the German README gave the installation snippet wrong. This is fixed now. While working at that, I noticed that all other READMEs made use of `pip` directly, while this is discouraged. See [here](https://snarky.ca/why-you-should-use-python-m-pip/) for a reference. Last but not least I tried to give the correct language specification to the  code blocks.


willmcgugan:
THanks.

PR #1792, opened 2022-01-02
Update year to 2022

Awilum:
Update year to 2022


willmcgugan:
Thanks, but this isn't required. https://hynek.me/til/copyright-years/

PR #1832, opened 2022-01-13
Fix width measurement of 353 emoji

darrenburns:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

The following test fails on `master`, but passes on this branch:

```python
def test_cell_len_emoji_presentation_sequence():
    # Handled different from standard emoji - we vary from wcwidth here, which reports width 1.
    assert cells.cell_len("🏵️") == 2
```

In the example above, I used the "rosette" emoji, but the same applies to all of the emoji listed on [this](https://unicode.org/emoji/charts/emoji-variants.html) page.

The Rich "codepoint to cell width" mapping is based off of the wcwidth database. For determining the width of an emoji, wcwidth looks at the East Asian Width Unicode database, which specifies whether characters in that database are whole/half width etc.

In constructing its database, wcwidth discards emojis if they don't have a `W` or `F` East Asian Width property type (see here: https://github.com/jquast/wcwidth/blob/4d4548048275ee25ba698cb0bd411296ff3f2238/bin/update-tables.py#L129), which results in their width being treated as 1.

However, the Unicode specification states that emoji presentation sequences (there are 353 of these, listed [here](http://www.unicode.org/Public/emoji/12.1/emoji-variation-sequences.txt)) should all be treated as East Asian Wide (`W`), regardless of the EAW property type.

> [UTS51] emoji presentation sequences behave as though they were East Asian Wide, regardless of their assigned East_Asian_Width property value.


willmcgugan (approved):
LGTM

willmcgugan (approved):

PR #1834, opened 2022-01-13
Stop using functools.lru_cache decorator on instance methods

### LINKED ISSUE #1831: Remove lru_cache from methods
Using lru_cache on a method may cause Python to hold on to references (via self) for longer than strictly necessary.

We should review instances of lru_cache decorator in Rich, and if affected we might want to replace them with the LRUCache object.

It may not be a huge problem since I think must uses of lru_cache set a sensible maximum size.

darrenburns:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [x] I've added tests for new code. (there is no new code, just shuffling things around mostly)

## Description

Mostly extracting the body of a method and shoving it into a function. There's a couple of smaller changes e.g. converting list to tuple since list isn't hashable too.

Fixes #1831


pbsds:
Alternative approach: https://pypi.org/project/methodtools/

willmcgugan (approved):
LGTM

PR #1852, opened 2022-01-19
Default datetime Highlighting

aaronst:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This PR adds default highlighting for [`datetime`](https://docs.python.org/3/library/datetime.html#datetime-objects) objects. It applies to the standard `repr` as well as the string returned by `.isoformat()`, which as far as I can tell just uses a `T` instead of a space between the date and time. The microseconds and time zone portions are optional.

For a quick example:
```python
from datetime import datetime
from rich import print

print(f'The current time is {datetime.now()}.')
```


willmcgugan:
I'm not keen on adding much more to that list of regexes, as it is quite large already and each new one slows highlighting down a little.

You could always create a custom highlighter if you need this.

aaronst:
Ah, makes sense. Thanks for taking a look.

Issue #1836, opened 2022-01-14
[REQUEST] Add suppress to RichHandler


daddycocoaman:
**How would you improve Rich?**

Hi. Thanks for the continued work and solving the recent issue (https://github.com/Textualize/rich/issues/1510) for adding suppress to `traceback.install`. Would it be possible to also add this to RichHandler as well so that the Traceback object there also suppresses modules? 

**What problem does it solve for you?**

Performs the same intended actions as addressed in the previous issue. It would also be convenient to do this in the logging handler too.

daddycocoaman:
I can make a PR for this. One thing to settle though; should RichHandler and traceback.install() have different Traceback objects or should they reference the same object with the ability to overwrite `suppress` in either call?

willmcgugan:
Yeah, good idea.

I think that RichHandler should probably grow `tracebacks_suppress` parameter, which gets passed to `emit`.

Thanks for the offer of a PR!

willmcgugan:
Thanks for the PR!

PR #1858, opened 2022-01-21
Add RichHandler suppress

daddycocoaman:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This PR address #1836 and adds the `suppress` functionality for Traceback to Rich Handler as well. The new `tracebacks_suppress` parameter is passed to the `Traceback` object created when `rich_tracebacks=True`.

One issue I had was creating a test. The test for Traceback suppress only counts the length of `suppress` after the modules/paths resolves. For this test, I tried to `emit` a custom `LogRecord` but I think the way Traceback handles rendering suppressed messages isn't compatible with an non-existent file. In other words, to ensure `RichHandler(console=console, rich_tracebacks=True, tracebacks_suppress=["foo.bar"])` is properly tested, an actual `foo` file should exist but I'm not sure how you want to handle extra files in the `tests` folder. I also tried supressing `pytest` but failed at that too.


docs/source/logging.rst
```diff
@@ -49,3 +49,23 @@ The :class:`~rich.logging.RichHandler` class may be configured to use Rich's :cl
...
+
+    import click
+    import logging
+    from rich.logging import RichHandler
+
+    logging.basicConfig(
+        level="NOTSET",
+        format="%(message)s",
+        datefmt="[%X]",
+        handlers=[RichHandler(rich_tracebacks=True, tracebacks_suppress)]
```
willmcgugan, line 68: Did you mean?

```python
tracebacks_suppress=[click]
```

docs/source/logging.rst
```diff
@@ -49,3 +49,23 @@ The :class:`~rich.logging.RichHandler` class may be configured to use Rich's :cl
 
 
 There are a number of other options you can use to configure logging output, see the :class:`~rich.logging.RichHandler` reference for details.
+
+Suppressing Frames
+------------------
+
+If you are working with a framework (click, django etc), you may only be interested in seeing the code from your own application within the traceback. You can exclude framework code by setting the `suppress` argument on `Traceback`, `install`, and `Console.print_exception`, which should be a list of modules or str paths.
```
willmcgugan, line 56: Could you update this to refer to the RichHandler.

willmcgugan (changes requested):
I'm happy with not writing an explicit test for this. It's such a small surface area for any faults.

Just a couple requests re doc.

willmcgugan (approved):

willmcgugan:
I think black was updated since you did this. Would you mind reformatting? And then should be good to go.

daddycocoaman:
> I think black was updated since you did this. Would you mind reformatting? And then should be good to go.

Sorry for the delay...haven't been paying attention to my GH feed lately. I've updated as requested.

willmcgugan:
Thanks!

PR #1862, opened 2022-01-23
Add Traditional Chinese readme and corresponding texts in __main__.py

toto6038:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [ ] I've added tests for new code. (Not checked because it's a documentation change.)

## Description

Add Traditional Chinese readme and texts in `__main__.py`'s **Asian language support** section. The original Simplified Chinese readme file is renamed to `README.zh-cn.md` in order to reflect that both versions of readme in Chinese are available. Changes to the navigation links in other readme files are also applied to follow mentioned renaming.
Traditional Chinese readme is reviewed by @jw910731
New screenshot may be needed because `__main__.py` are changed.

Edit: Update information about the reviewer


willmcgugan:
This is great. Thank you so much.

My only concern is that links to the original Chinese readme will be broken with this change. Perhaps we should keep the original Chinese readme, and add the traditional under a new filename?

toto6038:
To verify the change won't break links to the original Chinese readme, I created a new branch [`readme_test`](https://github.com/l10n-tw/rich/tree/readme_test) in my organization to test. It turns out that the experiment is successful. Since the links in readme were hard coded to the master branch of Textualize's rich repo, I modify them to the readme_test branch in l10n-tw's rich repo. The modification was only applied to English version for testing purpose, which means the links to both Chinese readmes in other language are, of courese, still broken. Hope this alleviate your concern.

By the way, would you have a plan to change links from absolute URL to relative path like `./README.de.md`?

willmcgugan:
It was actually links from outside of the project I was concerned about. If somewhere on the web links to the Chinese README, it would break with this change.

toto6038:
Understood. The latest commit reverts Simplified Chinese readme to its original file name. Links in other readme are also changed. It should work without problems.

willmcgugan:
謝謝!

rich/__main__.py
```diff
@@ -97,7 +97,7 @@ def comparison(renderable1: RenderableType, renderable2: RenderableType) -> Tabl
 
     table.add_row(
         "Asian\nlanguage\nsupport",
-        ":flag_for_china:  该库支持中文，日文和韩文文本！\n:flag_for_japan:  ライブラリは中国語、日本語、韓国語のテキストをサポートしています\n:flag_for_south_korea:  이 라이브러리는 중국어, 일본어 및 한국어 텍스트를 지원합니다",
+        ":flag_for_china:  该库支持中文，日文和韩文文本！\n:flag_for_japan:  ライブラリは中国語、日本語、韓国語のテキストをサポートしています\n:flag_for_south_korea:  이 라이브러리는 중국어, 일본어 및 한국어 텍스트를 지원합니다\n:flag_for_taiwan:  本函式庫支援中日韓文字元！",
```
willmcgugan, line 100: Sorry, this will break the tests which use this content. And it will require the screenshots to be generated. These were just examples and not meant to be exhuastive.

willmcgugan (changes requested):

rich/__main__.py
toto6038: That's sad, but I understand this change requires extra work. It's removed.

toto6038:
Conflict on CHANGELOG.me is resolved.

willmcgugan:
Thanks!

PR #1869, opened 2022-01-25
Add `Syntax.guess_lexer`, add support for more lexers (e.g. Django templates etc.)

darrenburns:
## Type of changes

- [x] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

The code which determines which lexer to use was always using filename. If you gave it a file with `.html` extension, it'd use the HTML lexer. This means that if you used e.g. a Django template with a file extension of `.html`, it would just use the HTML lexer, and not the Django one.

This change ensures we take advantage of the analysis Pygments can perform to select the most appropriate lexer in cases like this.


willmcgugan (approved):
LGTM

PR #1874, opened 2022-01-27
Handle classes in inspect when methods=True

darrenburns:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Removes assumption that all callables should be prefixed with `def`/ account for classes in inspect().

### DIFF
--- a/rich/_inspect.py
+++ b/rich/_inspect.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 
+import inspect
 from inspect import cleandoc, getdoc, getfile, isclass, ismodule, signature
 from typing import Any, Iterable, Optional, Tuple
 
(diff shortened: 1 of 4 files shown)


rich/_inspect.py
```diff
@@ -106,8 +107,17 @@ def _get_signature(self, name: str, obj: Any) -> Optional[Text]:
         signature_text = self.highlighter(_signature)
 
         qualname = name or getattr(obj, "__qualname__", name)
+
+        # If obj is a module, there may be classes (which are callable) to display
+        if inspect.isclass(obj):
```
willmcgugan, line 112: Will this not need a test?

darrenburns: Yeah, I've added one now.

willmcgugan (approved):

PR #1894, opened 2022-01-31
Add default_columns classmethod to Progress class

ptmcg:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Added new default_columns() classmethod to the Progress class, so that client code does not need to replicate the defaults literally, but can just add `*Progress.default_columns()`

### DIFF
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -588,12 +588,7 @@ class Progress(JupyterMixin):
             refresh_per_second is None or refresh_per_second > 0
         ), "refresh_per_second must be > 0"
         self._lock = RLock()
-        self.columns = columns or (
-            TextColumn("[progress.description]{task.description}"),
-            BarColumn(),
-            TextColumn("[progress.percentage]{task.percentage:>3.0f}%"),
-            TimeRemainingColumn(),
-        )
+        self.columns = columns or self.default_columns()
         self.speed_estimate_period = speed_estimate_period
 
         self.disable = disable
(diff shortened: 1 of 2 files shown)


rich/progress.py
```diff
@@ -613,6 +608,15 @@ def __init__(
         self.print = self.console.print
         self.log = self.console.log
 
+    @classmethod
+    def default_columns(cls) -> Tuple[ProgressColumn, ...]:
+        return (
```
willmcgugan, line 613: Could you change this to `get_default_changes`  and add a brief docstring...

willmcgugan (changes requested):
Great! Just a request for name and docstring.

rich/progress.py
ptmcg: Did you mean get_default_columns?

willmcgugan: Yeah, `get_default_columns`.  Another case of the Mondays...

ptmcg: I've got a decent docstring, but just want to double-check on the name. Do you *really* want get_default_columns()? I usually associate getters with instances, not the class. For this method, I'm not sure what value the "get_" prefix adds.

willmcgugan: Leaning towards `get`. Implies it does something more than expose data.

ptmcg: See updates

ptmcg:
This should be ready whenever you are. Unless you have further comments, I'll assume this is in your queue for the next release.

willmcgugan:
Thanks!

PR #1902, opened 2022-02-01
Fix #1530

### LINKED ISSUE #1530: [BUG] live displays and console printing are not thread safe
There seems to be some sort of threading/race issue with the console and the rich handler (or maybe with live dislays in general?) that causes log lines to get lost or misplaced when used with live displays or status displays.

**To Reproduce**
Here's a code example that will print out the logs from a queue in a separate thread. Note the log lines aren't then all expected to be above the status, but I was expecting the status to not interfere with it if it's thread safe. In particular the status isn't even getting properly wiped here.

```python
import atexit
import logging
import logging.handlers
import queue

import rich.logging
import rich.console

logger = logging.getLogger(__name__)
console = rich.console.Console()

root_logger = logging.getLogger("")
rhandler = rich.logging.RichHandler(console=console)
que = queue.Queue()
queue_handler = logging.handlers.QueueHandler(que)
listener = logging.handlers.QueueListener(que, rhandler)
root_logger.addHandler(queue_handler)
listener.start()
atexit.register(lambda x: x.stop(), listener)

with console.status("testing"):
    for n in range(1000):
        logger.error(n)
```
With this a variety of things can happen, the status can get leftover in a line printed at the same time as a log message or sometimes it can even eat a log message entirely.

Here's an example where it ate a line entirely
failure

a lot of the time I get other goofy stuff like this
other failure

**Platform**
Gentoo Linux / kitty 0.23.1

**Diagnose**

```console
pyffstream ❯ python -m rich.diagnose
python -m rich._windows
pip freeze | grep rich
│ │ <console width=147 ColorSystem.TRUECOLOR>                                    │ │
│     color_system = 'truecolor'                                                   │
│           height = 41                                                            │
│                        size=ConsoleDimensions(width=147, height=41),             │
│                        max_width=147,                                            │
│                        max_height=41,                                            │
│             size = ConsoleDimensions(width=147, height=41)                       │
│            width = 147                                                           │
platform="Linux"
WindowsConsoleFeatures(vt=False, truecolor=False)
rich==10.11.0

```

GBeauregard:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This is a temporary fix for #1530 rolled up into a PR so I can keep it up to date.


GBeauregard:
ah apparently I'll need to make my own branch :\

willmcgugan:
@GBeauregard This may fix the error, but it breaks the intended behaviour.

The context manager in Console does *not* lock any threads. It is intended to prevent interleaving in output when writing from multiple threads. It does this by writing to a thread-local buffer and only write to the console on exit.

It is more efficient this way because there is no waiting for exclusive access from threads.

Adding the thread lock will break this behaviour.

GBeauregard:
One idea I had is writing to the console while a live display is active could be disallowed; any valid usage that won't run into #1530 can get by with writing to the live display's console instance directly

PR #1918, opened 2022-02-04
Fixed force_terminal on Jupyter

verdimrc:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Fixes behavior after running `rich.pretty.install()` Jupyter Lab:

- Do not show `<Figure size 432x288 with 1 Axes>` on Matplotlib figures
- With `force_terminal=True`. and `force_jupyter=None` (or `True`), pandas dataframe must still be rendered once (i.e., the html repr).

### DIFF
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -36,6 +36,7 @@ from ._loop import loop_last
 from ._pick import pick_bool
 from .abc import RichRenderable
 from .cells import cell_len
+from .console import Console
 from .highlighter import ReprHighlighter
 from .jupyter import JupyterMixin, JupyterRenderable
 from .measure import Measurement
@@ -119,12 +120,17 @@ def _ipy_display_hook(
                     repr_result = method()
                 except Exception:
                     continue  # If the method raises, treat it as if it doesn't exist, try any others
-                if repr_result is not None:
-                    return  # Delegate rendering to IPython
+                else:
+                    # Delegated rendering to IPython, so no further action needed.
+                    return
 
     # certain renderables should start on a new line
     if _safe_isinstance(value, ConsoleRenderable):
         console.line()
+
+    if console.is_terminal:
+        # Force render as colored text, rather than html.
+        console = Console(force_terminal=True, force_jupyter=False)
 
     console.print(
         value


willmcgugan:
Why would you `force_terminal` in Jupyter lab?

Please don't ask me to run something in Jupyter when you can paste a screenshot.

verdimrc:
Thank your for taking a look. Added a screenshot to the PR description, with markers on the behaviors that this PR tries to fix.

willmcgugan:
I'm going to need an answer to my question, and some code I can cut and paste (or a notebook).

verdimrc:
Hi, I thought the quoted question was posted accidentally, my apology.

The motivation to let Jupyter display cell outputs as colored ASCII rather than HTML are:

1. to use the same font as the code cells. Otherwise, HTML output will be displayed with a noticeably different font.
2. to automatically wrap long ASCII texts (which is how Jupyter behaves without rich). Otherwise, long HTML output will need horizontal scrolling, which makes rich notebook to behave differently than stock Jupyter.

Below are the Python fragments to demonstrate the effect of this PR. You'll need to `pip install pandas` to run this example in Jupyter.

This fragment shows the effect on Jupyter "display_data".

```python
import pandas as pd
import rich
import rich.pretty

# Stuffs to display
d = {"A" * 200, "B" * 200}
df = pd.DataFrame({"a": range(3), "b": range(3)})

# Render "display_data" as colored text.
rich.reconfigure(force_terminal=True)
rich.pretty.install()
display(d, df, df["a"])

# Jupyter may surpress certain "display_data". For example, the semicolon makes Jupyter hide the
# "<AxesSubplot:...>" message.
df.plot();
```

Here's the side-by-side comparison:

patched

One thing to also note is this PR changes only the ipython display hook, and makes no attempt to change the Console behavior. In the next example, the Jupyter "stream_output" still defaults to HTML, and will have to be specifically instructed for colored ASCII. 

```python
# By default, render "stream_output" (e.g., stdout) as HTML.
rich.get_console().out("a", d)

# However, "stream_output" can be force rendered as colored text.
rich.console.Console(force_terminal=True, force_jupyter=False).out("a", d)
```

tdung (approved):

DrBenton:
@verdimrc  Note that in order to automatically wrap long ASCII texts we just have to specify a `fold` "overflow" to Rich: i.e. with `rich.pretty.install(overflow="fold")` the text is wrapped when the Jupyter notebook displays it :slightly_smiling_face:

Issue #1906, opened 2022-02-02
[REQUEST] Optional RichHandler Keywords


aaronst:
**How would you improve Rich?**

Currently, `RichHandler.KEYWORDS` is a class variable containing common HTTP methods.
https://github.com/Textualize/rich/blob/94273ee7030be773542110c0c70809535a08c6eb/rich/logging.py#L46-L55

It would be cool if `RichHandler()` had an optional `keywords` parameter that allowed for a custom set of keywords.

I'd be happy to submit a PR for this, let me know!

willmcgugan:
I don't see why not. Feel free to submit a PR.

PR #1919, opened 2022-02-04
feat: optional RichHandler keywords

aaronst:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

From https://github.com/Textualize/rich/issues/1906, add an optional `keywords` parameter to `RichHandler` to use instead of `RichHandler.KEYWORDS`.


rich/logging.py
```diff
@@ -171,7 +174,9 @@ def render_message(self, record: LogRecord, message: str) -> "ConsoleRenderable"
         if highlighter:
             message_text = highlighter(message_text)
 
-        if self.KEYWORDS:
+        if self.keywords:
+            message_text.highlight_words(self.keywords, "logging.keyword")
+        elif self.KEYWORDS:
```
aaronst, line 179: Tried to keep changes minimal, `render_message()` will first check to see if `self.keywords` is set, and fall back on `self.KEYWORDS` if it's not.

rich/logging.py
```diff
@@ -171,7 +174,9 @@ def render_message(self, record: LogRecord, message: str) -> "ConsoleRenderable"
         if highlighter:
             message_text = highlighter(message_text)
 
-        if self.KEYWORDS:
+        if self.keywords:
```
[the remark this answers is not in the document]
aaronst: Yeah makes sense, good idea! I'll make the changes now.

aaronst:
@willmcgugan, a couple files that I didn't touch in this PR were failing the automated `black` check, so I got those fixed. I'm not sure what to do about this `mypy` error though, I thought it was a difference between `List` and `Iterable`, but it doesn't like that `self.keywords` is `Optional`.

aaronst:
> @willmcgugan, a couple files that I didn't touch in this PR were failing the automated `black` check, so I got those fixed. I'm not sure what to do about this `mypy` error though, I thought it was a difference between `List` and `Iterable`, but it doesn't like that `self.keywords` is `Optional`.

So turns out it was actually the `Optional` on `self.KEYWORDS` that `mypy` didn't like. I removed it in the latest commit, but if we need to keep it `Optional` I can change the logic to something like:
```python
if self.keywords is None:
    if self.KEYWORDS is not None:
        message_text.highlight_words(self.KEYWORDS, "logging.keyword")
else:
    message_text.highlight_words(self.keywords, "logging.keyword")
```

rich/logging.py
```diff
@@ -41,9 +41,10 @@ class RichHandler(Handler):
             Defaults to 10.
         locals_max_string (int, optional): Maximum length of string before truncating, or None to disable. Defaults to 80.
         log_time_format (Union[str, TimeFormatterCallable], optional): If ``log_time`` is enabled, either string for strftime or callable that formats the time. Defaults to "[%x %X] ".
+        keywords (List[str], optional): List of words to highlight instead of ``RichHandler.KEYWORDS``.
     """
 
-    KEYWORDS: ClassVar[Optional[List[str]]] = [
+    KEYWORDS: ClassVar[List[str]] = [
```
willmcgugan, line 47: The typing can't be changed now. Since its been optional since day one, people may have reasonably set it to None for no keywords.

rich/logging.py
```diff
@@ -171,8 +174,10 @@ def render_message(self, record: LogRecord, message: str) -> "ConsoleRenderable"
         if highlighter:
             message_text = highlighter(message_text)
 
-        if self.KEYWORDS:
-            message_text.highlight_words(self.KEYWORDS, "logging.keyword")
+        message_text.highlight_words(
```
willmcgugan, line 177: How about this, to account for KEYWORDS being Optional?

```python
if keywords is None:
    keywords = self.KEYWORDS
if keywords:
    message_text.highlight_words(keywords, "logging.keyword")
```

willmcgugan (changes requested):

willmcgugan (approved):

willmcgugan:
Thanks.

PR #1957, opened 2022-02-12
fix link

brakhane:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Description

I just noticed that somehow a merge in my PR #1920 broke the link of a contributor. Not sure how it happened, but here's the fix.


willmcgugan:
Suspect that was my fault. Bad merge. Test fails are unrelated. Thanks

PR #1985, opened 2022-02-20
Hy support during inspection

shadowrylander:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Added support for [hy](https://docs.hylang.org/en/alpha/) in inspect submodule, and accounted for `ModuleNotFoundError`.


shadowrylander:
Ah; that's unfortunate. Am I allowed to post my fork to pypi?

willmcgugan:
If you wish. But equally you could just extract inspect.py

shadowrylander:
Ah, right; the inspect file technically doesn't have any local imports, does it? Such as `from . import module`, or `from .module import submodule`?

shadowrylander:
True; thanks for the advice! I'll just include the inspect file itself in one of my global modules, then, and use it wherever necessary.

shadowrylander:
Although I just realized that other rich submodules would not be able to use the separated inspect file. Would you happen to have any solutions for that?

PR #1992, opened 2022-02-22
Add condensed time column

bhrutledge:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

While [replacing Twine's progress bar with Rich](https://github.com/pypa/twine/pull/877), I got into some bike-shedding/yak-shaving to declutter the time column. After looking at several download progress indicators (e.g. Chrome, Firefox, `wget`, `scp`), it seems more common to show time remaining than time elapsed. Both `wget` and `scp` show time remaining until the download is finished, then switch to time elapsed. So, I implemented that behavior here.

For example:

https://user-images.githubusercontent.com/1326704/155128465-783d6866-006b-4e9c-bcae-45267b787e24.mp4

### DIFF
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -357,6 +357,30 @@ class TimeRemainingColumn(ProgressColumn):
         return Text(str(remaining_delta), style="progress.remaining")
 
 
+class CondensedTimeColumn(ProgressColumn):
+    """Renders estimated time remaining, or elapsed time when the task is finished."""
+
+    # Only refresh twice a second to prevent jitter
+    max_refresh = 0.5
+
+    def render(self, task: "Task") -> Text:
+        """Show time."""
+        style = "progress.elapsed" if task.finished else "progress.remaining"
+        task_time = task.finished_time if task.finished else task.time_remaining
+        if task_time is None:
+            return Text("--:--", style=style)
+
+        # Based on https://github.com/tqdm/tqdm/blob/master/tqdm/std.py
+        minutes, seconds = divmod(int(task_time), 60)
+        hours, minutes = divmod(minutes, 60)
+        if hours:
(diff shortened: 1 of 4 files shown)


rich/progress.py
```diff
@@ -357,6 +357,30 @@ def render(self, task: "Task") -> Text:
         return Text(str(remaining_delta), style="progress.remaining")
 
 
+class CondensedTimeColumn(ProgressColumn):
```
willmcgugan, line 360: The naming isn't clear. What is condensed time? How is it different from TimeRemaining or TimeElapsed? I know the answers, but I don't think it would be obvious to a user.

I do like the functionality. But if we add it I think it should be rolled in to the `TimeRemainingColumn` class so that the defaults continue to work as before, but with options for the new features.

willmcgugan (changes requested):

rich/progress.py
bhrutledge: Yeah, I don't love `Condensed`, but I chose that thinking of "a shorter time column with more information". Maybe `Compact` would be clearer.

However, I'm happy to roll it into `TimeRemainingColumn`. How about:

```python
class TimeRemainingColumn(ProgressColumn):
    """Renders estimated time remaining.

    Args:
        compact (bool, optional): Render MM:SS when time remaining is less than an hour. Defaults to False.
        elapsed_when_finished (bool, optional): Render time elapsed when the task is finished. Defaults to False.
    """
```

Or, should `compact` (or another name) encompass both behaviors? Or ???

bhrutledge: Done!

willmcgugan (approved):

bhrutledge:
My pleasure; thanks for accepting it. Any idea of when it might be released? No pressure, just curious when I can start using this in Twine. 😉

Issue #1991, opened 2022-02-22
Replace Colorama with internal solution for windows legacy mode


willmcgugan:
Colorama is a wonderful hack that wraps stdout to process ansi codes. It is problematic today because it is essentially a global setting and interferes with the ability to enable ansi escape sequences with newer Windows.

We should replace Colorama with equivalent functionality within Rich itself that generates the appropriate win32 api calls. This will hopefully be a little faster, and give us much more granular control over when to enable it.

I think we can do this at the Segment level. In Console, there is a `_render_buffer` method which renders segments in to text with ansi codes. I think we can generate the appropriate calls to the win32 api call here, if `self.legacy_windows` is True.

Since the win32 calls are only available on Windows, this should probably implemented within a conditional import. Perhaps `_windows_renderer.py`

See the Win32 docs for a reference on the appropriate API calls. It may also be worth referring to the colorama source.

https://docs.microsoft.com/en-us/windows/console/console-screen-buffers#character-attributes

I suggested we might have to process ansi codes to extract the win32 calls. On reflection I don't think we will need this. There is enough information in the `Segment` object that we don't need to parse anything. This should make it significantly faster.

PR #1993, opened 2022-02-22
Conversion of Segments to Windows Console API calls on legacy Windows platform

darrenburns:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Currently we depend on Colorama, which converts ANSI to Windows Console API calls in the context of legacy Windows terminals. This change removes the Colorama dependency, and has Rich directly call the Windows Console API.

- This should speed things up a little on older systems, since Colorama is no longer parsing ANSI codes.
- This change is implemented at the Segment level - so the calls to the API are done slightly earlier than before.
- We support a larger surface of the Console API now. For example, control sequences for showing and hiding the cursor are now handled where they previously weren't (we may see less flickering as a result), and the title of the terminal can be set.

Closes https://github.com/Textualize/rich/issues/1991

## Comparing new vs old

`rich.syntax` - New on left, old (Colorama) on right

image

`rich.color` - left column is this branch, right column is master

image
image

`rich.palette` - left this branch, right master

image

`rich.progress` - left this branch, right master - no difference in the amount of flickering, however, on master the cursor occasionally flashes on screen.
image


UltraStudioLTD:
@darrenburns I have a question.
Why do you define wrappers over Win32 API calls? I don't think it is needed

darrenburns:
@UltraStudioLTD The wrapper abstracts legacy Windows console API calls behind an easier to understand and more Pythonic interface, so that the rest of Rich doesn't need to know anything about legacy Windows APIs. The purpose of this PR is to drop the dependency on Colorama, and handle older Windows systems in a slightly more performative way.

rich/_win32_console.py
```diff
@@ -0,0 +1,514 @@
...
+def GetStdHandle(handle: int = STDOUT) -> wintypes.HANDLE:
+    return cast(wintypes.HANDLE, _GetStdHandle(handle))
+
+
+_GetConsoleMode = windll.kernel32.GetConsoleMode
+_GetConsoleMode.argtypes = [wintypes.HANDLE, wintypes.LPDWORD]
+_GetConsoleMode.restype = wintypes.BOOL
+
+
+def GetConsoleMode(std_handle: wintypes.HANDLE, console_mode: wintypes.DWORD) -> bool:
```
willmcgugan, line 70: I know this is a straightforward wrapping of a win32 call, but it is very surprising that console_mode is returned by modifying the parameter.

Suggest returning the DWORD and raising an exception if there is an error. 

Alternatively,  returning the DWORD or None if there is an error.

rich/_win32_console.py
```diff
@@ -0,0 +1,514 @@
...
+    def __init__(self, file: IO[str] = sys.stdout):
+        self.file = file
+        handle = GetStdHandle(STDOUT)
+        self._handle = handle
+        default_text = GetConsoleScreenBufferInfo(handle).wAttributes
+        self._default_text = default_text
+
+        self._default_fore = default_text & 7
+        self._default_back = (default_text >> 4) & 7
+        self._default_attrs = self._default_fore + self._default_back * 16
```
willmcgugan, line 237: Could we use bitwise operator here.

rich/_win32_console.py
```diff
@@ -0,0 +1,514 @@
...
+            row=cast(int, screen_size.Y), col=cast(int, screen_size.X)
+        )
+
+    def write_text(self, text: str) -> None:
+        """Write text directly to the terminal without any modification of styles
+
+        Args:
+            text (str): The text to write to the console
+        """
+        self.write(text)
```
willmcgugan, line 270: I have a feeling that if we use win32 api exclusively to write it may avoid certain edge cases, and possibly reduce flickering.

Can you give that a try?

I could very well be wrong about this but lets give it a shot.

willmcgugan (changes requested):
Great! Just a few requests.

willmcgugan:
@darrenburns Just looking in to styles a bit more. 

According to the colorama docs, it will interpret bold as bright, and also dim as normal intensity. We should probably do the same. Colorama doesn't support the "reverse" style, but I think we could do that quite easily by swapping "fore" and "back". Would you mind making those changes?

darrenburns:
@willmcgugan Made those changes in https://github.com/Textualize/rich/pull/2019

willmcgugan (approved):
LGTM

PR #2013, opened 2022-02-28
Add the `log_omit_repeated_times` option to `Console`

apcamargo:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This PR adds a `log_omit_repeated_times` option to the `Console` constructor. When set to `False`, the omission of repeated times by `log()` methods is disabled.


apcamargo:
Hey @willmcgugan. Is this a feature that could be considered? If so, do I need to add anything to the PR?

willmcgugan:
The constructor already has a large argument list. I'd prefer not to add to it without good reason.

What is your use case? Why do you not want identical times removed from the log function?

apcamargo:
In my case, I wanted to be able to grep a log and be able to check the time it was recorded. But my motivation to write this PR was just because it made sense to me to have feature parity.

willmcgugan:
This option makes sense for logging, but less so for `Console.log`. Not worth adding yet another option. Thanks for the contribution.

PR #2019, opened 2022-03-03
Add support for bold, dim, and reverse on legacy Windows consoles

darrenburns:
Some fixes for the PR #1993

* Adds support for bold, dim, and reverse on legacy Windows consoles.
* Calls the Windows Console API in order to output text, instead of writing to the `Console.file`
* More docstrings


willmcgugan (approved):
LGTM

PR #2031, opened 2022-03-07
Add support for named tuples to pretty

### LINKED ISSUE #2012: [REQUEST] Add explicit support for NamedTuples
The Pretty class doesn't support pretty printing NamedTuples.

We should attempt to auto detect named tuples in order to expand these on to multiple lines as appropriate.

NamedTuples inherit from `tuple` and don't have a common base class. They have a `_fields` attribute which could identify them as a named tuple. 

We will also need to check if a custom __repr__ has been provided (in the same was as we do dataclasses).

darrenburns:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Closes #2012 

Named tuples will be formatted similarly to dataclasses when formatted using the `pretty` module. The code for handling this is very similar to dataclasses, but I opted not to try and factor it out as I think it will reduce readability.

### DIFF
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -524,6 +524,18 @@ class _Line:
             )
 
 
+def _is_namedtuple(obj: Any) -> bool:
+    base_classes = getattr(type(obj), "__bases__", [])
+    if len(base_classes) != 1 or base_classes[0] != tuple:
+        return False
+
+    fields = getattr(obj, "_fields", None)
+    if not fields or not isinstance(fields, tuple):
+        return False
+
+    return all(type(field) == str for field in fields)
+
+
 def traverse(
     _object: Any,
     max_length: Optional[int] = None,
@@ -731,7 +743,32 @@ def traverse(
                     append(child_node)
 
                 pop_visited(obj_id)
+        elif _is_namedtuple(obj):
+            obj_id = id(obj)
+            if obj_id in visited_ids:
+                # Recursion detected
+                return Node(value_repr="...")
+            push_visited(obj_id)
 
+            children = []
+            append = children.append
+            if reached_max_depth:
+                node = Node(value_repr="...")
+            else:
+                node = Node(
+                    open_brace=f"{obj.__class__.__name__}(",
+                    close_brace=")",
+                    children=children,
+                )
+
+                for last, (key, value) in loop_last(obj._asdict().items()):
+                    child_node = _traverse(value, depth=depth + 1)
+                    child_node.key_repr = key
+                    child_node.last = last
+                    child_node.key_separator = "="
+                    append(child_node)
+
+                pop_visited(obj_id)
         elif _safe_isinstance(obj, _CONTAINERS):
             for container_type in _CONTAINERS:
                 if _safe_isinstance(obj, container_type):
@@ -878,6 +915,15 @@ if __name__ == "__main__":  # pragma: no cover
             1 / 0
             return "this will fail"
 
+    from typing import NamedTuple
+
+    class StockKeepingUnit(NamedTuple):
+        name: str
+        description: str
+        price: float
+        category: str
+        reviews: List[str]
+
     d = defaultdict(int)
     d["foo"] = 5
     data = {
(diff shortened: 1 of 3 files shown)


rich/pretty.py
```diff
@@ -731,7 +753,23 @@ def iter_attrs() -> Iterable[
                     append(child_node)
 
                 pop_visited(obj_id)
-
+        elif _is_namedtuple(obj):
```
willmcgugan, line 756: We are also going to need a check to see if the default `__repr__` was overriden. If it was, Rich should respect that. Have a look at `_is_dataclass_repr`, maybe we can do something similar for NamedTuple.

rich/pretty.py
```diff
@@ -524,6 +524,28 @@ def __str__(self) -> str:
             )
 
 
+def _is_namedtuple(obj: Any) -> bool:
```
willmcgugan, line 527: I wonder if we can be more precise than this. I know there is no `is_namedtuple` function, alas.

willmcgugan (changes requested):
Mainly just detecting if `__repr__` has been overriden.

rich/pretty.py
```diff
@@ -79,6 +79,24 @@ def _is_dataclass_repr(obj: object) -> bool:
...
+
+    Args:
+        obj (object): A namedtuple
+
+    Returns:
+        bool: True if the default repr is used, False if there's a custom repr.
+    """
+    obj_file = inspect.getfile(obj.__repr__)
+    default_repr_file = inspect.getfile(_dummy_namedtuple.__repr__)
+    print(obj_file, default_repr_file)
```
willmcgugan, line 96: Debug print snuck in there.

rich/pretty.py
```diff
@@ -524,6 +543,28 @@ def __str__(self) -> str:
...
+    there is only a minuscule chance of this happening unintentionally.
+
+    Args:
+        obj (Any): The object to test
+
+    Returns:
+        bool: True if the object is a namedtuple. False otherwise.
+    """
+    base_classes = getattr(type(obj), "__bases__", [])
+    if len(base_classes) != 1 or base_classes[0] != tuple:
```
willmcgugan, line 558: You can extend tuples, and named tuples. Perhaps this should just be `if not isinstance(tuple)` ?

rich/pretty.py
```diff
@@ -524,6 +543,28 @@ def __str__(self) -> str:
...
+        return False
+
+    fields = getattr(obj, "_fields", None)
+    if not fields or not isinstance(fields, tuple):
```
willmcgugan, line 562: `_fields` *could* be an empty tuple here. For the strange edge-case of a named tuple with no fieds.

Perhaps just return `isinstance(fields, tuple)` ?

rich/pretty.py
```diff
@@ -731,7 +772,24 @@ def iter_attrs() -> Iterable[
...
+            if reached_max_depth:
+                node = Node(value_repr="...")
+            else:
+                node = Node(
+                    open_brace=f"{obj.__class__.__name__}(",
+                    close_brace=")",
+                    children=children,
+                )
+                for last, (key, value) in loop_last(obj._asdict().items()):
+                    print(last, key, value)
```
willmcgugan, line 787: print statement...

rich/pretty.py
```diff
@@ -79,6 +79,24 @@ def _is_dataclass_repr(obj: object) -> bool:
...
+def _has_default_namedtuple_repr(obj: object) -> bool:
+    """Check if an instance of namedtuple contains the default repr
+
+    Args:
+        obj (object): A namedtuple
+
+    Returns:
+        bool: True if the default repr is used, False if there's a custom repr.
+    """
+    obj_file = inspect.getfile(obj.__repr__)
```
willmcgugan, line 94: This can throw a `TypeError` according to the docs. Is that an issue here?

willmcgugan (changes requested):
Introspection is tricky

rich/pretty.py
darrenburns: I've handled `TypeError` and `OSError` now.

willmcgugan (approved):
LGTM

PR #2037, opened 2022-03-08
Handle case inspect fails on repl-defined objects

darrenburns:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Objects defined in a REPL are part of the `__main__` scope, which doesn't have a corresponding file, and so `rich.inspect` was failing when trying to find this filename.

`inspect.getfile` raises an `OSError` in this case.

We catch the `OSError` and ignore it, as the rest of the code already handles the case where there is no `source_filename` available.

### DIFF
--- a/rich/_inspect.py
+++ b/rich/_inspect.py
@@ -98,7 +98,8 @@ class Inspect(JupyterMixin):
         source_filename: Optional[str] = None
         try:
             source_filename = getfile(obj)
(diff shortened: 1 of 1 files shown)


willmcgugan (approved):
That was simple :)

Issue #2042, opened 2022-03-09
[BUG] Tracebacks don't expand to fit height properly


henryiii:
**Describe the bug**

If you put a Traceback inside a container with a height, it exactly chops off the message at the bottom (regardless of how many lines it has, tested with 1 and a longer message producing 2), and instead matches the frame around the traceback. This is currently causing an issue when placing a traceback in a panel in textual.

MWE:
```python
import rich.traceback
import rich.panel

try:
    1/0
except Exception:
    rich.print(rich.panel.Panel(rich.traceback.Traceback(), height=20))
```

The output has chopped off the bottom of the traceback:

```
╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮                                                   │
│ │                                                                                                  │                                                   │
│ │ /Users/henryschreiner/git/scikit-hep/uproot-browser/tmp2.py:5 in <module>                        │                                                   │
│ │                                                                                                  │                                                   │
│ │   2 import rich.panel                                                                            │                                                   │
│ │   3                                                                                              │                                                   │
│ │   4 try:                                                                                         │                                                   │
│ │ ❱ 5 │   1/0                                                                                      │                                                   │
│ │   6 except Exception:                                                                            │                                                   │
│ │   7 │   rich.print(rich.panel.Panel(rich.traceback.Traceback(), height=20))                      │                                                   │
│ │   8                                                                                              │                                                   │
│ │                                                                                                  │                                                   │
│ │                                                                                                  │                                                   │
│ │                                                                                                  │                                                   │
│ │                                                                                                  │                                                   │
│ │                                                                                                  │                                                   │
│ │                                                                                                  │                                                   │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────────╯                                                   │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

If you do not set the height externally (remove `height=20` above), then the correct height is computed:

```
╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮                                                   │
│ │                                                                                                  │                                                   │
│ │ /Users/henryschreiner/git/scikit-hep/uproot-browser/tmp2.py:5 in <module>                        │                                                   │
[...]

PR #2043, opened 2022-03-09
reset height of children

willmcgugan:
Reset height of children in render.

Fix for https://github.com/Textualize/rich/issues/2042

When rendering an object with a height, that height is passed to the renderable via options. The problem was that any renderables yielded by `__rich_console__` were passed that same height, which was not intended.

This change ensures that height is reset for these sub objects.

Also removes tabulate.py which was marked for deprecation a while ago.


CHANGELOG.md
```diff
@@ -25,14 +24,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix syntax lexer guessing.
 - Fixed Pretty measure not respecting expand_all https://github.com/Textualize/rich/issues/1998
 - Collapsed definitions for single-character spinners, to save memory and reduce import time.
-- Fix print_json indent type in __init__.py
+- Fix print_json indent type in **init**.py
```
darrenburns, line 27: Not exactly an issue but just curious - is this some markdown formatter thing that caused this?

rich/console.py
```diff
@@ -224,6 +224,16 @@ def update_height(self, height: int) -> "ConsoleOptions":
         options.max_height = options.height = height
         return options
 
+    def reset_height(self) -> "ConsoleOptions":
+        """Resets the height to None.
```
darrenburns, line 228: The docstring makes it sound like the method is doing self.height = None
```suggestion
        """Returns a copy of the ConsoleOptions with height=None
```

darrenburns (approved):
Minor docstring change suggested, take it or leave it :)

CHANGELOG.md
willmcgugan: Yeah, I think that may have been vscode. I need to put it inside backticks.

rich/console.py
willmcgugan: Good point.

I had auto-merge on. Feel free to "request changes" if there is a worthwhile change.

PR #2066, opened 2022-03-15
Handle recording on legacy mode Windows consoles

### LINKED ISSUE #2053: [BUG] Rich 12.0.0 fails to write to stdout attached to a pipe on Windows
### Describe the bug

Since rich 12.0.0, printing to stdout fails running under a subprocess attached to a pipe on Windows. 

To reproduce, save the following reproducer and run it first with rich==12.0.0

```python
import subprocess
import sys

code = """\
import sys

console = Console(file=sys.stdout)
console.print("spam")
"""
err_code = code.replace("sys.stdout", "sys.stderr")
cmd = [sys.executable, "-"]

print("[capture_output=True via sys.stdout]")
proc = subprocess.run(cmd, input=code, capture_output=True, encoding="utf-8")
print(f"{proc.stdout=} {proc.stderr=}")

print("[capture_output=True via sys.stderr]")
proc = subprocess.run(cmd, input=err_code, capture_output=True, encoding="utf-8")
print(f"{proc.stdout=} {proc.stderr=}")

print("[capture_output=False via sys.stdout]")
proc = subprocess.run(cmd, input=code, capture_output=False, encoding="utf-8")
```

image

There's a few observations here:

- With rich==12.0.0 installed, the captured stdout is empty. With an older version it works fine (see orange / blue)
- Standard error is completely unaffected. It's also fine letting the subprocess inherit the streams (see green)

### Platform

**Python:** CPython 3.8.5
**OS:** Windows 10 Home 21H2 (build 19044.1526)
**Terminal:** Windows Terminal 1.11.3471.0

<details>
<summary>rich.diagnose</summary>

```
(venv) PS R:\Programming\diff-shades> python -m rich.diagnose
│ │ <console width=148 ColorSystem.TRUECOLOR>                                    │ │
│     color_system = 'truecolor'                                                   │
│           height = 35                                                            │
│                        size=ConsoleDimensions(width=148, height=35),             │
│                        max_width=148,                                            │
│                        max_height=35,                                            │
│             size = ConsoleDimensions(width=148, height=35)                       │
│            width = 148                                                           │
╭── <class 'rich._windows.WindowsConsoleFeatures'> ───╮
│ Windows features available.                         │
│                                                     │
│ ╭─────────────────────────────────────────────────╮ │
│ │ WindowsConsoleFeatures(vt=True, truecolor=True) │ │
│ ╰─────────────────────────────────────────────────╯ │
│                                                     │
│ truecolor = True                                    │
│        vt = True                                    │
╰─────────────────────────────────────────────────────╯
╭────── Environment Variables ───────╮
│     'TERM': None,                  │
╰────────────────────────────────────╯
platform="Windows"
```

</details>

### Additional context

[This was discovered by GHA Windows CI](https://github.com/ichard26/diff-shades/runs/5528992639?check_suite_focus=true#step:5:235) (the problem is hidden and buried in this test, but this shows it has been reproduced under several environments[^1]). Original investigation into this strange behaviour happened in [the #black-formatter channel on Python Discord](https://discord.com/channels/267624335836053506/846434317021741086/952628086535520367).

According to @jack1142 who helped me investigate it, it's related to the recent move to replace colorama with first-party code that interacts with the Windows APIs:

> I'm pretty sure it's because it can't use WriteConsole API (the one that's used for legacy console printing) when it's piped because output from WriteConsole is not capturable 

They then shared this excerpt from Windows console API documentation:

> WriteConsole fails if it is used with a standard handle that is redirected to a file. If an application processes multilingual output that can be redirected, determine whether the output handle is a console handle (one method is to call the GetConsoleMode function and check whether it succeeds). If the handle is a console handle, call WriteConsole. If the handle is not a console handle, the output is redirected and you should call WriteFile to perform the I/O.
https://docs.microsoft.com/en-us/windows/console/writeconsole 

[^1]: if you're curious, it's basically testing whether recording the output (via `Console(record=True)`) and then dumping it as HTML works as expected with the rest of the tool

darrenburns:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Console.record=True was only being handled inside `_render_buffer`, which was not called on legacy mode consoles.

I've extracted the code for recording the buffer out of this method, to ensure it gets called regardless of environment.

Closes #2053


rich/console.py
```diff
@@ -1924,21 +1929,24 @@ def _check_buffer(self) -> None:
                     if WINDOWS:
                         try:
                             file_no = self.file.fileno()
+                            stdout_num = sys.stdout.fileno()
```
willmcgugan, line 1932: These probably don't need to be in a try except blog. Suggest replacing them with constants. `pty.STDOUT_FILENO` and `pty.STDERR_FILENO` contain the values so they don't need to be hard coded.

willmcgugan (changes requested):

willmcgugan (approved):
LGTM

darrenburns:
@willmcgugan `pty` module doesn't work on Windows. Has a comment at the top of it too saying it's only tested on Unix and MacOS.

willmcgugan (approved):

PR #2084, opened 2022-03-18
fix: full static typing, fix issues discovered

### LINKED ISSUE #2083: [BUG] typing: Progress in Group isn't happy
**Describe the bug**

Running mypy on the following code:

```python
from rich.console import Group

outer_progress = Progress()
inner_progress = Progress()
live_group = Group(outer_progress, inner_progress)
```

Produces:


```console
$ mypy --strict tmp.py
tmp.py:6: error: Argument 1 to "Group" has incompatible type "Progress"; expected "Union[ConsoleRenderable, RichCast, str]"
tmp.py:6: note: Following member(s) of "Progress" have conflicts:
tmp.py:6: note:     Expected:
tmp.py:6: note:         def __rich__(self) -> Union[ConsoleRenderable, str]
tmp.py:6: note:     Got:
tmp.py:6: note:         def __rich__(self) -> Union[ConsoleRenderable, RichCast, str]
tmp.py:6: error: Argument 2 to "Group" has incompatible type "Progress"; expected "Union[ConsoleRenderable, RichCast, str]"
tmp.py:6: note:     Expected:
tmp.py:6: note:         def __rich__(self) -> Union[ConsoleRenderable, str]
tmp.py:6: note:     Got:
tmp.py:6: note:         def __rich__(self) -> Union[ConsoleRenderable, RichCast, str]
Found 2 errors in 1 file (checked 1 source file)
```

I think `RichCast` should also be in the Protocol, that is, `__rich__` is allowed to return an object with `__rich__`, ~~or it should not be in `__rich__`, that is, `__rich__(self) -> Union[ConsoleRenderable, str]` should be used for all `__rich__` methods. Which is correct depends on runtime; can a `__rich__` return a `__rich__` which can return a `__rich__`, etc?~~. Ahah, I see `CHANGELOG.md:167:- Allowed `__rich__` to work recursively`, so it's the former.

I'm preparing a PR.

**Platform**
<details>

What platform (Win/Linux/Mac) are you running on? What terminal software are you using?

I may ask you to copy and paste the output of the following commands. It may save some time if you do it now.

If you're using Rich in a terminal:

```
╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── <class 'rich.console.Console'> ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ A high level console interface.                                                                                                                                                                                                                                            │
│                                                                                                                                                                                                                                                                            │
│ ╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮ │
│ │ <console width=298 ColorSystem.TRUECOLOR>                                                                                                                                                                                                                              │ │
│ ╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                                                                                                                                                                                            │
│     color_system = 'truecolor'                                                                                                                                                                                                                                             │
│         encoding = 'utf-8'                                                                                                                                                                                                                                                 │
│             file = <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>                                                                                                                                                                                           │
│           height = 68                                                                                                                                                                                                                                                      │
│    is_alt_screen = False                                                                                                                                                                                                                                                   │
│ is_dumb_terminal = False                                                                                                                                                                                                                                                   │
│   is_interactive = True                                                                                                                                                                                                                                                    │
│       is_jupyter = False                                                                                                                                                                                                                                                   │
│      is_terminal = True                                                                                                                                                                                                                                                    │
│   legacy_windows = False                                                                                                                                                                                                                                                   │
│         no_color = False                                                                                                                                                                                                                                                   │
│          options = ConsoleOptions(size=ConsoleDimensions(width=298, height=68), legacy_windows=False, min_width=1, max_width=298, is_terminal=True, encoding='utf-8', max_height=68, justify=None, overflow=None, no_wrap=False, highlight=None, markup=None, height=None) │
│            quiet = False                                                                                                                                                                                                                                                   │
│           record = False                                                                                                                                                                                                                                                   │
│         safe_box = True                                                                                                                                                                                                                                                    │
│             size = ConsoleDimensions(width=298, height=68)                                                                                                                                                                                                                 │
│        soft_wrap = False                                                                                                                                                                                                                                                   │
│           stderr = False                                                                                                                                                                                                                                                   │
│            style = None                                                                                                                                                                                                                                                    │
│         tab_size = 8                                                                                                                                                                                                                                                       │
│            width = 298                                                                                                                                                                                                                                                     │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─── <class 'rich._windows.WindowsConsoleFeatures'> ────╮
│ ╭───────────────────────────────────────────────────╮ │
│ ╰───────────────────────────────────────────────────╯ │
╰───────────────────────────────────────────────────────╯
╭──────────────────────────────────────────────────────────────────────────────────────────── Environment Variables ────────────────────────────────────────────────────────────────────────────────────────────╮
│ {'TERM': 'xterm-256color', 'COLORTERM': 'truecolor', 'CLICOLOR': None, 'NO_COLOR': None, 'TERM_PROGRAM': 'iTerm.app', 'COLUMNS': None, 'LINES': None, 'JPY_PARENT_PID': None, 'VSCODE_VERBOSE_LOGGING': None} │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
platform="Darwin"
rich==11.2.0
```

(Same issue after upgrading to Rich 12)

</details>

henryiii:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

This updates mypy, ~~uses pre-commit's controlled environment for mypy more consistently~~ (reverted for now), and fixes a variety of issues found. One feature in mypy 0.940 is the ability to show an error code if a blanket type-ignore is applied; instead of just adding the suggested codes, I've removed all the type ignores, the carefully readded them, trying to fix the underlying problem if possible. I was able to fix several. Also, mypy seems a little buggy when not running with strict, so I've made sure that always runs. I'm using the new(ish) pyproject.toml configuration, which is much nicer than the old ini configuration.

There were a few typing bugs discovered (highlighted via a "code-review", in a minute). Let me know if you'd like this broken up into different commits or anything.

This now checks a bit more of the code. I've mostly added type ignores to the examples rather than adding types, assuming you wanted to keep them "untyped/simple", unless there were already types in the same file.

Fixes #2083


Makefile
```diff
@@ -5,7 +5,7 @@ format-check:
 format:
 	black .
 typecheck:
-	mypy -p rich --strict --no-incremental
+	pre-commit run --all-files mypy
 typecheck-report:
```
henryiii, line 9: This `typecheck-report`) is not used anywhere. Can it be removed? Then `mypy` could be dropped for the dev deps, which would simplify the install/update process in poetry.

examples/table_movie.py
```diff
@@ -60,7 +61,7 @@
 
 
 @contextmanager
-def beat(length: int = 1) -> None:
+def beat(length: int = 1) -> Generator[None, None, None]:
```
henryiii, line 64: Incorrect type before.

rich/spinner.py
```diff
@@ -10,6 +10,9 @@
     from .style import StyleType
 
 
+__all__ = ["Spinner", "SPINNERS"]
```
henryiii, line 13: SPINNERS was missing, used in the examples. Needs to be explicitly exported.

tools/make_emoji.py
```diff
@@ -4,9 +4,10 @@
     print("pip install emoji")
     raise
 
-from emoji.unicode_codes import EMOJI_ALIAS_UNICODE
+from tkinter import W
+from emoji.unicode_codes import EMOJI_ALIAS_UNICODE_ENGLISH
 
-emoji = {k.lower().strip(":"): v for k, v in EMOJI_ALIAS_UNICODE.items()}
+emoji_dict = {k.lower().strip(":"): v for k, v in EMOJI_ALIAS_UNICODE_ENGLISH.items()}
```
henryiii, line 10: This seems to be a bug, typing seems to be right, there's no matching attribute in https://github.com/carpedm20/emoji/blob/master/emoji/unicode_codes/__init__.py that I can see. Mayne this was what was meant?

Also avoiding overriding the imported `emoji` module with a dict.

pyproject.toml
```diff
@@ -50,3 +50,16 @@ asv = "^0.5.1"
 [build-system]
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
+
+
+[tool.mypy]
+files = ["benchmarks", "examples", "rich", "tools"]
+python_version = "3.6"
```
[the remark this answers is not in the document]
henryiii: Reverted for the moment, since this already runs per-version, let's keep that for the moment.

henryiii:
Addressed a few things and rebased on latest master.

rich/_lru_cache.py
```diff
@@ -1,12 +1,16 @@
-from collections import OrderedDict
-from typing import Dict, Generic, TypeVar
-
+from typing import Dict, Generic, TypeVar, TYPE_CHECKING
+import sys
 
 CacheKey = TypeVar("CacheKey")
 CacheValue = TypeVar("CacheValue")
 
+if sys.version_info < (3, 8):
```
henryiii, line 7:
```suggestion
if sys.version_info < (3, 9):
```

Oops, applied both changes, didn't mean to add this one.

henryiii:
(I'm running the CI on my fork for the moment, will post a comment here once it's passing. 3.9 & 3.10 are passing, of course (that's what I'm checking locally), but 3.7 doesn't like OrderedDict.)

henryiii:
Okay, passing on my fork, so should be good. Didn't notice that there were hacks inside of LRUCache that now needed to be removed since it's back to being a nice, normal class again.

rich/segment.py
```diff
@@ -141,6 +141,7 @@ def _split_cells(cls, segment: "Segment", cut: int) -> Tuple["Segment", "Segment
                     _Segment(before[: pos - 1] + " ", style, control),
                     _Segment(" " + text[pos:], style, control),
                 )
+        raise AssertionError("Can't reach here")
```
henryiii, line 144:
```suggestion
        raise AssertionError("Can't reach here")  # pragma: no cover
```

Can't see the coverage report, but this is one that could clearly use a no cover pragma.

rich/console.py
```diff
@@ -263,7 +263,9 @@ def update_dimensions(self, width: int, height: int) -> "ConsoleOptions":
 class RichCast(Protocol):
     """An object that may be 'cast' to a console renderable."""
 
-    def __rich__(self) -> Union["ConsoleRenderable", str]:  # pragma: no cover
+    def __rich__(
+        self,
+    ) -> Union["ConsoleRenderable", "RichCast", str]:  # pragma: no cover
```
henryiii, line 268:
```suggestion
    def __rich__(  # pragma: no cover
        self,
    ) -> Union["ConsoleRenderable", "RichCast", str]:
```
I'm getting this might be the other thing coverage is complaining about. Maybe the pragma should be on the first line rather than at the end of the def? Pragmas on all three lines would be silly...

Makefile
willmcgugan: What would be simplified exactly?

I use it to generate typecheck reports.

willmcgugan:
There is little point in typechecking examples. They are for educational purposes, and I don't want to assume the viewer is familiar with type annotations. That includes `mypy:` comments which have the potential to confuse.

If anything, the types should be removed from the examples, unless it is particularly relevant.

rich/console.py
```diff
@@ -1235,8 +1237,8 @@ def render(
         render_iterable: RenderResult
 
         renderable = rich_cast(renderable)
-        if hasattr(renderable, "__rich_console__") and not isclass(renderable):
-            render_iterable = renderable.__rich_console__(self, _options)  # type: ignore
+        if isinstance(renderable, ConsoleRenderable) and not isclass(renderable):
```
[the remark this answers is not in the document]
willmcgugan: This was they way I had it previously, however it was notable slower than the less elegant `hasattr` which is why I replaced it.

tools/make_emoji.py
willmcgugan: That might be a fix. The code in tools isn't meant for public consumption. There's little point in maintaining code that is only use very occasionally by us.

willmcgugan:
Hi Henry

Could you break this PR in to manageable chunks. Hard to see the wood for the trees.

examples and tools don't need typing.

Thanks

henryiii:
Absolutely, I'll break this up, address feedback, and resubmit in 2-3 smaller PRs. Might be a little bit of time (1-2 days), unrelated work deadline is interfering; might be able to fit it in sooner, though. Thanks for #2089!

rich/console.py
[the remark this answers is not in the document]
henryiii: I was afraid of that. Okay, will leave this as just an ignore then.

Makefile
henryiii: My comment is not very valid anymore anyway - I usually recommend running mypy via pre-commit only; mypy is very sensitive to its environment, and it is very sensitive to the target Python version (sometimes you can't make it pass on multiple versions due to type ignore's changing validity based on Python version - we gave up in `pypa/build`, IIRC). But this running a different way, and passing, so not going to suggest anything different for now.

PR #2110, opened 2022-03-25
fix(types): use Type[T] instead of T for cls

henryiii:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

Another typing cleanup/fix pulled out from
https://github.com/Textualize/rich/pull/2084. By using `cls: Type[T]` instead of
`T`, MyPy becomes smarter about handling this and less type ignores are needed,
since the typing is more precise.

### DIFF
--- a/rich/repr.py
+++ b/rich/repr.py
@@ -1,5 +1,6 @@
 from functools import partial
 import inspect
+import sys
 
 from typing import (
     Any,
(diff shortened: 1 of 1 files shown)


willmcgugan (approved):

willmcgugan:
Thanks

Issue #2119, opened 2022-03-26
[REQUEST] Allow setting default parameters for `Console` via environment variables


joouha:
Hello,

When using `rich` in [`euporie`](https://github.com/joouha/euporie) (a terminal Jupyter notebook client), `rich` detects that it is running in a Jupyter notebook and not running in a TTY, and so returns HTML output.

However, you get a far superior experience when using `rich` in `euporie` by by setting the following `Console` parameters:

```python
Console(force_jupyter=False, force_terminal=True)
```

I'd like to be able to pre-configure `rich` to use these parameters by default when running in `euporie`. 

The Jupyter kernel in which `rich` runs is a separate process to `euporie`, but will almost always be a sub-process of it. This means that if the default `Console` parameters could be set using environment variables, this would allow me to pre-configure `rich` to give the best user experience when used in `euporie`.

I'm happy to take stab at implementing this.

Let me know your thoughts,
Thanks

willmcgugan:
What are you proposing?

joouha:
I'm proposing that the some of the default parameters values for `Console.__init__` be configurable using environment variables

willmcgugan:
I got that, but what environment variables and what parameters?

joouha:
I think it makes sense to be able to override the default behaviour of `force_jupyter`, `force_terminal` & `force_interactive`.

The environment variables could be called "RICH_FORCE_JUPYTER", "RICH_FORCE_TERMINAL", etc.

I've made a PR: #2122

joouha:
Hello,

I've "fixed" my original issue by implementing `ipywidgets` support, and writing a HTML/CSS renderer for [euporie](https://github.com/joouha/euporie), so `rich`'s output is now handled in the same way as it is in JupyterLab.

It feels like a bit of an overly-complex solution, but it now means that `rich.live.Live` objects work without clearing the entire cell output, and that the HTML output of `JupyterRenderable`s / `JupyterMixin`s look good in the terminal.

So you can now do this:

[_rich.webm](https://user-images.githubusercontent.com/12154190/185574143-7e107a3f-a827-407a-acda-e9422e270395.webm)

Issue #2112, opened 2022-03-25
[BUG] Using status overwrites all previous output when in Jupyter notebook


ggydush-fn:
**Describe the bug**

When in a Jupyter notebook, using `status` overwrites all previous output.

```python
import time

from rich.console import Console

print("Previous output")
console = Console()

with console.status('This overwrites previous output'):
    time.sleep(2)
```

**Platform**
<details>
<summary>Click to expand</summary>

What platform (Win/Linux/Mac) are you running on? What terminal software are you using?

I may ask you to copy and paste the output of the following commands. It may save some time if you do it now.

If you're using Rich in a terminal:
```
╭────────────────────── <class 'rich.console.Console'> ──────────────────────╮
│ A high level console interface.                                            │
│                                                                            │
│ ╭────────────────────────────────────────────────────────────────────────╮ │
│ │ <console width=93 ColorSystem.TRUECOLOR>                               │ │
│ ╰────────────────────────────────────────────────────────────────────────╯ │
│                                                                            │
│     color_system = 'truecolor'                                             │
│         encoding = 'utf-8'                                                 │
│             file = <ipykernel.iostream.OutStream object at 0x7f9c0ea1eb20> │
│           height = 100                                                     │
│    is_alt_screen = False                                                   │
│ is_dumb_terminal = False                                                   │
│   is_interactive = False                                                   │
│       is_jupyter = True                                                    │
│      is_terminal = False                                                   │
│   legacy_windows = False                                                   │
│         no_color = False                                                   │
│          options = ConsoleOptions(                                         │
│                        size=ConsoleDimensions(width=93, height=100),       │
│                        legacy_windows=False,                               │
│                        min_width=1,                                        │
│                        max_width=93,                                       │
│                        is_terminal=False,                                  │
│                        encoding='utf-8',                                   │
│                        max_height=100,                                     │
│                        justify=None,                                       │
│                        overflow=None,                                      │
│                        no_wrap=False,                                      │
│                        highlight=None,                                     │
│                        markup=None,                                        │
│                        height=None                                         │
│                    )                                                       │
│            quiet = False                                                   │
│           record = False                                                   │
│         safe_box = True                                                    │
│             size = ConsoleDimensions(width=93, height=100)                 │
│        soft_wrap = False                                                   │
│           stderr = False                                                   │
│            style = None                                                    │
│         tab_size = 8                                                       │
│            width = 93                                                      │
╰────────────────────────────────────────────────────────────────────────────╯
[...]

PR #2122, opened 2022-03-27
Allow default Console parameters to be configured with environment variables

joouha:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This allows the default values of the `Console` parameters `force_jupyter`, `force_terminal` & `force_interactive` to be overridden by setting the environment variables `RICH_FORCE_JUPYTER`, `RICH_FORCE_TERMINAL` & `RICH_FORCE_INTERACTIVE`.

Fixes issue #2119

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add missing `end` keyword argument to `Text.from_markup`
 - Fixed issue with decoding ANSI reset https://github.com/Textualize/rich/issues/2112
 
+### Added
+
+- Allow the default values for `Console` prameters `force_jupyter`, `force_terminal` & `force_interactive` to be set using environment variables
+
 ## [12.0.1] - 2022-03-22
 
 ### Changed
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -14,6 +14,7 @@ The following people have contributed to the development of Rich:
 - [Oleksis Fraga](https://github.com/oleksis)
 - [Michał Górny](https://github.com/mgorny)
 - [Leron Gray](https://github.com/daddycocoaman)
+- [Josiah Outram Halstead](https://github.com/joouha)
 - [Kenneth Hoste](https://github.com/boegel)
 - [Finn Hughes](https://github.com/finnhughes)
 - [Josh Karpel](https://github.com/JoshKarpel)
--- a/rich/console.py
+++ b/rich/console.py
@@ -5,6 +5,7 @@ import platform
 import sys
 import threading
 from abc import ABC, abstractmethod
+from ast import literal_eval
 from dataclasses import dataclass, field
 from datetime import datetime
 from functools import wraps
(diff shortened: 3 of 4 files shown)


willmcgugan:
I'm not keen on adding env vars that aren't already a standard. As a library, Rich shouldn't impose its own env vars on to applications. It should be the application author that decides how to handle env vars.

joouha:
Fair enough.

I see that Spyder had the same issue I am experiencing here:

https://github.com/Textualize/rich/pull/1221#issuecomment-835734314

(In the link above you suggested adding a generic environment variable for exactly this purpose).

Spyder managed to work around the issue by creating their own kernel wrapper around `ipykernel`, which allows them to set `isatty` for `stdout` to True, and change `ipython.__class__.__name__`, meaning that the default values for `Console.is_jupyter` becomes `False` and `Console.is_terminal` becomes `True` (this is what I am trying to achieve).

However, creating my own kernel isn't really an option for me with `euporie`, as users need to be able to run any kernel of their choosing. 

I could override the kernel's configuration when it is launched to add hooks that make the required changes to the kernel. But, the kernel does not necessarily run in the same python environment or interpreter, so this would require users to install an extra package which contains the hooks (not ideal).

---

My other idea is that `rich` could add a [metadata dictionary](https://ipython.readthedocs.io/en/stable/config/integrating.html#MyObject._repr_mimebundle_) to the output of `JupyterRenderable._repr_mimebundle_` which would allow me to identify that the output comes from `rich` and therefore I probably want to choose to render the _text/plain_ output over the _text/html_ one.

I suppose that I could simply search the _text/plain_ output for terminal escape sequences and make a decision about what to render based on that.

Do you have any other ideas?

joouha:
I've done some testing, and rendering the `text/plain` output of `JupyterRenderable._repr_mimebundle_` from rich doesn't help. Since `euporie` doesn't support jupyter-widgets, the `clear_output` call made when a "live" rich renderable refreshes causes the entire cell output to clear. This means that you wouldn't be able to use `rich.progress.track` and display other output in the same cell, for example. (This is similar to what is going on in #2112).

>  As a library, Rich shouldn't impose its own env vars on to applications. It should be the application author that decides how to handle env vars.

I'm not sure how this should apply when `rich` is being use interactively in a Jupyter notebook. The author of a Jupyter notebook is the one writing the code which calls `rich`, but I don't think they would expect to have to change their code to reconfigure `rich`'s `Console` depending on which notebook editor they are using.

willmcgugan:
@joouha I think I need to understand your use case a bit more. Would you be free for a video chat, next week perhaps?

joouha:
@willmcgugan Yes, I can do a video call. My e-mail is in my profile if you'd like to set something up.

Commandcracker:
It would be nice to see this Implemented!
I can see many use cases for a force_terminal env var.

Examples:

- Some Terminals support colors but rich isn't detecting it and editing every package to force_terminal isn't worth the time. 

- in GitHub actions force_terminal is required to enable color. Setting an env var in a Dockerfile would be an easy fix. also i don't know how to tell pip to force_terminal when i run it from shell.

force_terminal=False
Screenshot_20220528_221839

force_terminal=True
Screenshot_20220528_221912

- Big packages like twine also have some issues not having an env var for it. wich can be seen [here](https://github.com/pypa/twine/blob/main/twine/cli.py#L34-L37)

willmcgugan:
Closing. Assumed stale.

PR #2125, opened 2022-03-28
fix(types): remove a few more type ignores

henryiii:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

A few type ignores can be pretty easily removed. Mostly discovered by enabling error codes (next PR).

### DIFF
--- a/rich/markdown.py
+++ b/rich/markdown.py
@@ -613,11 +613,12 @@ if __name__ == "__main__":  # pragma: no cover
         import pydoc
         import io
 
+        fileio = io.StringIO()
         console = Console(
-            file=io.StringIO(), force_terminal=args.force_color, width=args.width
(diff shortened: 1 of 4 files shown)


rich/markdown.py
```diff
@@ -613,11 +613,12 @@ def __rich_console__(
         import pydoc
         import io
 
+        fileio = io.StringIO()
         console = Console(
```
henryiii, line 617: `Console` hides the exact type of `file=`. Instead of making it Generic, the original object can be used here, which mypy does know the exact type of.

willmcgugan:
Looks like the CI will pass now. 🤷‍♂️

willmcgugan (approved):

PR #2130, opened 2022-03-29
chore(types): include error code on type ignores

henryiii:
Signed-off-by: Henry Schreiner <henryschreineriii@gmail.com>

## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

This adds type ignore codes to all type ignores, and activates the (new in MyPy
0.94x) feature to verify that all type ignores have error codes. Unlike
previous methods of requiring error codes (like pre-commit's pygrep hook), mypy
will actually tell you what the correct error code is in the error message,
which is quite nice for adding them.

Having the error codes improves the readability, since you get a hint as to why
the type ignore is there (actual human readable hint, unlike flake8's numeric
codes), and it also keeps one type ignore from hiding another problem. This was
used in the last few PRs to see type ignores that were pretty easy to work
around.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ files = ["rich"]
 warn_unused_configs = true
 show_error_codes = true
 strict = true
+enable_error_code = ["ignore-without-code"]
 
 [[tool.mypy.overrides]]
 module = ["pygments.*", "IPython.*", "commonmark.*", "ipywidgets.*"]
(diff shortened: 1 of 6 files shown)


willmcgugan (approved):

willmcgugan:
Great, Thanks.

PR #2131, opened 2022-03-29
fix(types): typing fixes exposed by extra checking

henryiii:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

This enables a few extra mypy checks, and fixes type bugs exposed by those. There were three features enabled, but I've left one off: `warn_unrachable = true`. While it is really useful, and helped expose a couple of the issues fixed here, it also requires type ignoring in about 5-7 places unavoidably, so I've left it off. The most irritating one is `if sys.platform.startswith("win")` constructs, which it will notice are unreachable if you are typing targeting a different platform. MyPy might fix this in https://github.com/python/mypy/issues/12286, which would run all platforms/versions in parallel and then combine the results (which would be really fantastic!), but until then, it's likely better just to manually add `--warn-unreachable` once in a while and make sure the errors reported are expected. (Depending on your preference, of course - I've been turning it on mostly, but I usually don't have quite as many ignores due to it - I thik I counted at least 7 here. If you want to see it, let me know, and I'll PR it).

I'll comment inline on the others, but this is one fix:

```python
x: Iterable[Any] = ...
len(x)
```

is invalid. You can't take the length of an iterator. Someone could correctly put an iterator function in here, and this will break. Either this should be `Sequence` (which is what I've used here), or you should iterate over the iterator and build a sequence like a list or a tuple and then use that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,10 +54,9 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.mypy]
 files = ["rich"]
-warn_unused_configs = true
 show_error_codes = true
 strict = true
-enable_error_code = ["ignore-without-code"]
+enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
 
 [[tool.mypy.overrides]]
 module = ["pygments.*", "IPython.*", "commonmark.*", "ipywidgets.*"]
--- a/rich/console.py
+++ b/rich/console.py
@@ -1947,7 +1947,7 @@ class Console:
         frame = currentframe()
         if frame is not None:
             # Use the faster currentframe where implemented
-            while offset and frame:
+            while offset and frame is not None:
                 frame = frame.f_back
                 offset -= 1
             assert frame is not None
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -19,6 +19,7 @@ from typing import (
     Iterable,
     List,
     Optional,
+    Sequence,
     Set,
     Tuple,
     Union,
@@ -28,8 +29,10 @@ from rich.repr import RichReprResult
 
 try:
     import attr as _attr_module
+
+    _has_attrs = True
 except ImportError:  # pragma: no cover
-    _attr_module = None  # type: ignore[assignment]
+    _has_attrs = False
 
 from . import get_console
 from ._loop import loop_last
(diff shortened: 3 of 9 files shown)


rich/console.py
```diff
@@ -1947,7 +1947,7 @@ def _caller_frame_info(
         frame = currentframe()
         if frame is not None:
             # Use the faster currentframe where implemented
-            while offset and frame:
+            while offset and frame is not None:
```
henryiii, line 1950: frame doesn't have `__bool__` or `__len__`, so it will always evaluate to true. This makes it explicit.

rich/pretty.py
```diff
@@ -28,8 +29,10 @@
 
 try:
     import attr as _attr_module
+
+    _has_attrs = True
 except ImportError:  # pragma: no cover
-    _attr_module = None  # type: ignore[assignment]
+    _has_attrs = False
```
henryiii, line 35: These are always a little painful in MyPy; using this method helped it it out below. I wish mypy had a bit better support try/except and setting to None.

rich/spinner.py
```diff
@@ -79,7 +79,8 @@ def render(self, time: float) -> "RenderableType":
             self.speed = self._update_speed
             self._update_speed = 0.0
 
-        if not self.text:
+        # This normally can't be str, unless someone assigned it later.
+        if not self.text:  # type: ignore[truthy-bool]
```
henryiii, line 83: Typewise this can't happen, but I could see it possibly happing at runtime, so just added an ignore and note here.

rich/style.py
```diff
@@ -740,7 +740,7 @@ def __add__(self, style: Optional["Style"]) -> "Style":
         new_style._link = style._link or self._link
         new_style._link_id = style._link_id or self._link_id
         new_style._hash = style._hash
-        new_style._null = self._null or style._null
+        new_style._null = style._null
```
henryiii, line 743: There's a guard about 10 lines above:

```python
if self._null:
    return style
```

So `self._null` cannot be truthy at this point.

rich/tree.py
```diff
@@ -45,7 +45,7 @@ def add(
         style: Optional[StyleType] = None,
         guide_style: Optional[StyleType] = None,
         expanded: bool = True,
-        highlight: bool = False,
+        highlight: Optional[bool] = False,
```
henryiii, line 48: This is checked for `None` just below.

pyproject.toml
```diff
@@ -54,10 +54,9 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.mypy]
 files = ["rich"]
-warn_unused_configs = true
```
henryiii, line 57: This is included in strict, which I didn't realize until checking the `mypy --help` output just now.

rich/progress.py
```diff
@@ -643,9 +643,7 @@ def __init__(
         disable: bool = False,
         expand: bool = False,
     ) -> None:
-        assert (
-            refresh_per_second is None or refresh_per_second > 0
-        ), "refresh_per_second must be > 0"
+        assert refresh_per_second > 0, "refresh_per_second must be > 0"
```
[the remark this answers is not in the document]
willmcgugan: Keep in mind not everyone uses Mypy. Asserting things at runtime that are impossible as far as the typing is concerned, can still provide helpful errors for beginners.

That said, I genuinely don't recall why we check for None here. 🤷‍♂️

rich/spinner.py
henryiii: Yes, that appears to be it, mypy assigns `self.text: RenderableType` because Text fits that Protocol. But the `__bool__` behavior is unique to Text, so this confuses it below. Making this explicit solves the issue.

willmcgugan (approved):

willmcgugan:
Thanks

PR #2143, opened 2022-04-02
feat: waves spinner

H999:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Suggest adding 2 new spinners: waves1, waves2
[discussions 2142](https://github.com/Textualize/rich/discussions/2142)


wasi-master:
Why does this spinner have 38k phases? I am no genius but wouldn't that have some performance impacts?

H999:
> Why does this spinner have 38k phases? I am no genius but wouldn't that have some performance impacts? probably not but 38k seems a lot, can't it be shortened?

Hi @wasi-master, first sorry for slow response second thank for your feedback.
I think you talk about wave2, first idea I want have a wave spinner like [this](https://www.youtube.com/watch?v=-qtZ5oiJVqE) so I don't care so much about length of phases. After take a look in code, and I think you are right, I know where I should fix, because the base `frames = "▁▂▃▄▅▆▇█▇▆▅▄▃▂"` have length is 14 so when use `itertools.product(frames, repeat=4)` it be 14^4 = 38416 and when use set it equal vs `itertools.product(frames[:8], repeat=4)` so then number of phases now have only 8^4 = 4096, I fix it in [this](https://github.com/Textualize/rich/pull/2143/commits/a8dc743d35ece44b4837f203574bd3003e6c6bf3).
Again, thanks @wasi-master

willmcgugan:
Sorry, can't accept code which runs at import time. This may have to remain something you could add to your own project, if you want it. But probably doesn't belong in the core lib.

H999:
> Sorry, can't accept code which runs at import time. This may have to remain something you could add to your own project, if you want it. But probably doesn't belong in the core lib.

Ok, it alright! Because I found the spinner growVertical similar with wave have 1 size length so I create the def to dynamic the length like I wrote [here](https://github.com/Textualize/rich/discussions/2142#discussioncomment-2489492)

> The first one can be converted to a list with 16 elements
> 
> ```python
> ['▁▂▃▄', '▂▃▄▅', '▃▄▅▆', '▄▅▆▇', '▅▆▇█', '▆▇█▇', '▇█▇▆', '█▇▆▅', '▇▆▅▄', '▆▅▄▃', '▅▄▃▂', '▄▃▂▁', '▃▂▁▂', '▂▁▂▃']
> ```
> 
> The second one, however, because the order is random, can't really be calculated before runtime.

Yes, @wasi-master that is the first version of wave [here](https://github.com/Textualize/rich/discussions/2142#discussioncomment-2487339)

So @willmcgugan do you want the first version of wave sine in the core lib?

Issue #2150, opened 2022-04-04
[BUG] ImportError OrderedDict


johannes-graeter:
You may find a solution to your problem in the [docs](https://rich.readthedocs.io/en/latest/introduction.html) or [issues](https://github.com/willmcgugan/rich/issues).

First: Thanks for your awesome project and the work you take to make devs lifes better :)!
**Describe the bug**

Some updates seem to have mixed up dependencies.
```bash
> rich README.md      
Traceback (most recent call last):
  File "/home/graeter/.local/bin/rich", line 5, in <module>
    from rich_cli.__main__ import run
  File "/home/graeter/.local/lib/python3.8/site-packages/rich_cli/__main__.py", line 7, in <module>
    from rich.console import Console, RenderableType
  File "/home/graeter/.local/lib/python3.8/site-packages/rich/console.py", line 46, in <module>
    from ._log_render import FormatTimeCallable, LogRender
  File "/home/graeter/.local/lib/python3.8/site-packages/rich/_log_render.py", line 5, in <module>
    from .text import Text, TextType
  File "/home/graeter/.local/lib/python3.8/site-packages/rich/text.py", line 5, in <module>
    from rich.emoji import EmojiVariant
  File "/home/graeter/.local/lib/python3.8/site-packages/rich/emoji.py", line 4, in <module>
    from .jupyter import JupyterMixin
  File "/home/graeter/.local/lib/python3.8/site-packages/rich/jupyter.py", line 4, in <module>
    from .segment import Segment
  File "/home/graeter/.local/lib/python3.8/site-packages/rich/segment.py", line 19, in <module>
    from .cells import (
  File "/home/graeter/.local/lib/python3.8/site-packages/rich/cells.py", line 6, in <module>
    from ._lru_cache import LRUCache
  File "/home/graeter/.local/lib/python3.8/site-packages/rich/_lru_cache.py", line 8, in <module>
    from typing_extensions import OrderedDict
ImportError: cannot import name 'OrderedDict' from 'typing_extensions' (/home/graeter/.local/lib/python3.8/site-packages/typing_extensions.py)
```
Can you point me to a working configuration?
Up to now I used rich very happily and would miss it a lot ;)

**Platform**
- Ubuntu 20.04
- python 3.8.10
- pip 22.0.4
- zsh with starship

<details>
```
python -m rich.diagnose
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
[...]

PR #2157, opened 2022-04-04
bump typing extension

willmcgugan:
Fixes dependancy issue https://github.com/Textualize/rich/issues/2150

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -585,7 +585,7 @@ test = ["check-manifest", "fastjsonschema", "testpath", "pytest", "pytest-cov"]
 
 [[package]]
 name = "nest-asyncio"
-version = "1.5.4"
+version = "1.5.5"
 description = "Patch asyncio to allow nested event loops"
 category = "main"
 optional = true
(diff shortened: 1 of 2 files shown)


cschlittPHI:
Is there any way to bump the version number from this? We can't seem to use it as it uses the old version number still I think?

willmcgugan:
New release coming soon

willmcgugan:
@henryiii FYI we had to bump typing-extensions because the minimum version didn't have OrderedDict

henryiii:
That's annoying, very sorry! According to https://github.com/python/typing/tree/master/typing_extensions it was added in 3.7. They don't have a proper changelog until "4.0", so it can be hard to tell what's in typing_extensions, so I thought it was in typing_extensions 3.7 too. Looking at the history, it was added in 3.10.0 in https://github.com/python/typing/commit/ea7f88e8d2dd28a5569b2385c21b17063c480c7f Please feel free to ping me if you need any help with anything related to what I've done.

In build we test all minimum versions too: https://github.com/pypa/build/blob/main/tests/constraints.txt (ironically, except for `typing_extensons`). I can set up a test like that if you want one.

Issue #2154, opened 2022-04-04
[BUG] Wrong terminal size when piping to a Rich app


willmcgugan:
When using a Console between two pipes, i.e `git diff | dunk | less -R` The output width is not the full width of the terminal by default.

Inside `dunk`, it's worked around by doing `Console(force_terminal=True, width=os.get_terminal_size(2))` which results in the content that gets piped into less being the whole terminal width.

PR #2160, opened 2022-04-04
Fall back to `sys.__stderr__` on non-Windows OS to get size

DrBenton:
This fixes the "pipe" bug, where we could not determine the available size correctly when the output of Rich is piped to another process

## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [x] I accept that @willmcgugan may be pedantic in the code review. :sweat_smile: 

## Description

This should (:crossed_fingers:) fix this bug: https://github.com/Textualize/rich/issues/2154

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -1102,12 +1102,16 @@ class Console:
             except OSError:  # Probably not a terminal
                 pass
         else:
-            try:
-                width, height = os.get_terminal_size(sys.__stdin__.fileno())
-            except (AttributeError, ValueError, OSError):
+            posix_std_descriptors = (
+                sys.__stdin__,  # try this one first...
+                sys.__stdout__,  # ...then that one...
+                sys.__stderr__,  # ...and ultimately try to fall back to this one
+            )
+            for descriptor in posix_std_descriptors:
                 try:
-                    width, height = os.get_terminal_size(sys.__stdout__.fileno())
-                except (AttributeError, ValueError, OSError):
+                    width, height = os.get_terminal_size(descriptor.fileno())
+                    break
+                except (AttributeError, ValueError, OSError) as err:
                     pass
 
         columns = self._environ.get("COLUMNS")
--- a/rich/console.py
+++ b/rich/console.py
@@ -1109,9 +1109,10 @@ class Console:
             for file_descriptor in _STD_STREAMS:
                 try:
                     width, height = os.get_terminal_size(file_descriptor)
-                    break
                 except (AttributeError, ValueError, OSError):
                     pass
+                else:
+                    break
 
         columns = self._environ.get("COLUMNS")
         if columns is not None and columns.isdigit():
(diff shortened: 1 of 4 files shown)


tests/test_console.py
```diff
@@ -122,6 +123,48 @@ def test_size():
...
+            sys.__stdout__.fileno(): stdout_size,
+            sys.__stderr__.fileno(): stderr_size,
+        }[fileno]
+        if value is ValueError:
+            raise value
+        return value
+
+    get_terminal_size_mock.side_effect = get_terminal_size_mock_impl
+
+    console = Console(legacy_windows=False)
```
DrBenton, line 162: looks like we always remove one pixel to the width if this is not forced to `False` ?

rich/console.py
```diff
@@ -1102,11 +1106,10 @@ def size(self) -> ConsoleDimensions:
...
                 pass
         else:
-            try:
-                width, height = os.get_terminal_size(sys.__stdin__.fileno())
-            except (AttributeError, ValueError, OSError):
+            for file_descriptor in _STD_STREAMS:
                 try:
-                    width, height = os.get_terminal_size(sys.__stdout__.fileno())
+                    width, height = os.get_terminal_size(file_descriptor)
+                    break
```
willmcgugan, line 1112: Suggest putting that break in an `else:` clause.

tests/test_console.py
willmcgugan: Yeah, legacy windows treats new lines differently, they actually take up a cell in the terminal. We need to subtract 1 so the new line character doesn't go on to the following line.

willmcgugan (changes requested):
Just one small request.

rich/console.py
willmcgugan: An else: clause on the try except block, avoids having too much code in the try block

```python
try:
    width,height =. ...
except:
    # exceptions
else:
    break
```

rich/console.py
```diff
@@ -82,17 +82,21 @@ class NoChange:
...
     _STDOUT_FILENO = 1
-
 try:
     _STDERR_FILENO = sys.__stderr__.fileno()
 except Exception:
     _STDERR_FILENO = 2
 
-_STD_STREAMS = (_STDOUT_FILENO, _STDERR_FILENO)
+_STD_STREAMS = (_STDIN_FILENO, _STDOUT_FILENO, _STDERR_FILENO)
+_STD_STREAMS_OUTPUT = (_STDOUT_FILENO, _STDERR_FILENO)
```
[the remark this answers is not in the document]
willmcgugan: Yup

rich/console.py
DrBenton: Ah, I see! Alright, let's do this :+1:  :slightly_smiling_face:

DrBenton:
@willmcgugan feedback (hopefully) addressed: [3eac1e4](https://github.com/Textualize/rich/pull/2160/commits/3eac1e47a814deb9bc14b0263070b32c7643d0d0)   :slightly_smiling_face:

willmcgugan (approved):
All good, thanks!

PR #2166, opened 2022-04-05
[progress] The `total` we pass to a Progress renderable can now be `None`

DrBenton:
## Type of changes

- [ ] Bug fix
- [x] New feature: https://github.com/Textualize/rich/issues/1054
- [ ] Tests
- [ ] Other

## Checklist

## Description

With this PR we now render a "pulse" progress bar when the given `total` is None. 
Which is a new possibility we now have: this parameter used to be a `float`, and is now an `Optional[float]` - so it shouldn't bring any compatibility issues, as the type is just widened to a value that was not accepted according to the documentation, and which would raise exceptions before if it was used?

#### Design considerations

To carry the semantic of _"we don't know the 'total' value for my progress bar"_, instead of `None` we could have used other kinds of values, such as:
 - A negative value, like `total=-1`. The benefit would have been that the signature would still have been `float` , but the drawback is that the meaning of this value is less clear when debugging
 - A sentinel value, such as `total=rich.progress.UNDETERMINED` : probably the most explicit option, but a bit more cumbersome for our users than a good old `None`? :thinking:

### DIFF
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -148,7 +148,7 @@ def track(
                 finished_style=finished_style,
                 pulse_style=pulse_style,
             ),
-            TextColumn("[progress.percentage]{task.percentage:>3.0f}%"),
+            TaskProgressColumn(),
             TimeRemainingColumn(),
         )
     )
@@ -675,6 +675,43 @@ class TimeElapsedColumn(ProgressColumn):
         return Text(str(delta), style="progress.elapsed")
 
 
+class TaskProgressColumn(TextColumn):
+    """A column displaying the progress of a task."""
+
+    def __init__(
+        self,
+        text_format: str = "[progress.percentage]{task.percentage:>3.0f}%",
+        text_format_no_percentage: str = "",
+        style: StyleType = "none",
+        justify: JustifyMethod = "left",
+        markup: bool = True,
+        highlighter: Optional[Highlighter] = None,
+        table_column: Optional[Column] = None,
+    ) -> None:
+        self.text_format_no_percentage = text_format_no_percentage
+        super().__init__(
+            text_format=text_format,
+            style=style,
+            justify=justify,
+            markup=markup,
+            highlighter=highlighter,
+            table_column=table_column,
+        )
+
+    def render(self, task: "Task") -> Text:
+        text_format = (
+            self.text_format_no_percentage if task.total is None else self.text_format
+        )
+        _text = text_format.format(task=task)
+        if self.markup:
+            text = Text.from_markup(_text, style=self.style, justify=self.justify)
+        else:
+            text = Text(_text, style=self.style, justify=self.justify)
+        if self.highlighter:
+            self.highlighter.highlight(text)
+        return text
+
+
 class TimeRemainingColumn(ProgressColumn):
     """Renders estimated time remaining.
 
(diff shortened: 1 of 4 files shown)


tests/test_progress.py
```diff
@@ -3,33 +3,32 @@
 import io
 import os
 import tempfile
-from time import sleep
 from types import SimpleNamespace
 
 import pytest
 
 import rich.progress
-from rich.progress_bar import ProgressBar
```
DrBenton, line 12: only changed these imports in order to pass the `isort` part of the pre-commit hook :slightly_smiling_face:

rich/progress.py
```diff
@@ -40,7 +40,7 @@
     from typing_extensions import Literal  # pragma: no cover
 
 from . import filesize, get_console
-from .console import Console, JustifyMethod, RenderableType, Group
+from .console import Console, Group, JustifyMethod, RenderableType
```
DrBenton, line 43: `isort` doing its things :slightly_smiling_face:

rich/progress.py
```diff
@@ -651,7 +651,7 @@ def __init__(
     def render(self, task: "Task") -> ProgressBar:
         """Gets a progress bar widget for a task."""
         return ProgressBar(
-            total=max(0, task.total),
+            total=max(0, task.total) if task.total is not None else None,
```
DrBenton, line 654: I had to add loads of `if task.total is not None else` here and there, now that this "total" can be `None`

willmcgugan:
Making total `None` may qualify as a breaking change, but the alternatives don't seem much better...

What does a progress bar with default settings display when `total` is `None` now? Can I have a screenshot please.

willmcgugan:
I think what we might have to do is replace that TextColumn with the percentage, with a column that displays either the percentage OR iterations per second. I'm pretty sure that is what tqdm does.

The ETA should probably not display anything if there is no total.

Basically, the user should get something that broadly does what they would expect, given the constraints. Suggest analysing what tqdm does here.

DrBenton:
@willmcgugan hopefully this commit addresses your comment :crossed_fingers:  :slightly_smiling_face: : [f41bda6](https://github.com/Textualize/rich/pull/2166/commits/f41bda6de0b3b62541d9c66ef3582b3d1d107c30)

The progress bar in action:
progress-bar-with-none-total-(4)

rich/progress.py
```diff
@@ -1585,16 +1647,16 @@ def remove_task(self, task_id: TaskID) -> None:
         *Progress.get_default_columns(),
         TimeElapsedColumn(),
         console=console,
-        transient=True,
+        transient=False,
```
willmcgugan, line 1650: I'm assuming the code below was for testing. Can you restore this so that `progress -m rich.progress` show a demonstration of multiple progress bars.

willmcgugan (changes requested):
Great! Couple of small requests.

DrBenton:
Wooops, sorry @willmcgugan  for the mess in the `if __name__ == "__main__":` part of the "progress.py" module - I should have reviewed myself before requesting your review again! :sweat: 

I amended the commit, hopefully addressing your 2 comments: d2830c89b7a91631a :crossed_fingers:

willmcgugan (approved):
LGTM

Issue #2147, opened 2022-04-03
embedding Rich SVGs leaks styles


willmcgugan:
When embedding SVGs within a web page, the SVG styles "leak" in to the HTML. For instance the span style changes in the rest of the page.

I'm not sure if there is some easy way to prevent that, but if not, perhaps we should generate a random ID for the SVG and update the styles to only use that UD.

PR #2170, opened 2022-04-05
[console] namespace SVG elements to avoid conflicts with surrounding HTML page

DrBenton:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

The name of the font and the ID of the wrapper we use for the generated SVG code are now dynamic, and their default value contain "rich" in order to avoid conflicts with people's own resources if they embed such generated SVGs in an HTML page.
All the CSS selectors are also prefixed with the ID of the wrapper now, which should isolate then from the other ones defined in a surrounding HTML page.

Issue: https://github.com/Textualize/rich/issues/2147

On this annotated screenshot we can see this namespacing in action:
Screenshot from 2022-04-04 16-23-13-annotated

### DIFF
--- a/rich/__init__.py
+++ b/rich/__init__.py
@@ -1,7 +1,7 @@
 """Rich text and beautiful formatting in the terminal."""
 
 import os
-from typing import Callable, IO, TYPE_CHECKING, Any, Optional, Union
+from typing import IO, TYPE_CHECKING, Any, Callable, Optional, Union
 
 from ._extension import load_ipython_extension  # noqa: F401
 
@@ -11,7 +11,7 @@ if TYPE_CHECKING:
     from .console import Console
 
 # Global console used by alternative print
-_console: Optional["Console"] = None
+_console_default_singleton: Optional["Console"] = None
 
 _IMPORT_CWD = os.path.abspath(os.getcwd())
 
@@ -23,13 +23,13 @@ def get_console() -> "Console":
     Returns:
         Console: A console instance.
     """
-    global _console
-    if _console is None:
+    global _console_default_singleton
+    if _console_default_singleton is None:
         from .console import Console
 
-        _console = Console()
+        _console_default_singleton = Console()
 
-    return _console
+    return _console_default_singleton
 
 
 def reconfigure(*args: Any, **kwargs: Any) -> None:
--- a/rich/__init__.py
+++ b/rich/__init__.py
@@ -11,7 +11,7 @@ if TYPE_CHECKING:
     from .console import Console
 
 # Global console used by alternative print
-_console_default_singleton: Optional["Console"] = None
+_console: Optional["Console"] = None
 
 _IMPORT_CWD = os.path.abspath(os.getcwd())
 
@@ -23,13 +23,13 @@ def get_console() -> "Console":
     Returns:
         Console: A console instance.
     """
-    global _console_default_singleton
-    if _console_default_singleton is None:
+    global _console
+    if _console is None:
         from .console import Console
 
-        _console_default_singleton = Console()
+        _console = Console()
 
-    return _console_default_singleton
+    return _console
 
 
 def reconfigure(*args: Any, **kwargs: Any) -> None:
--- a/rich/_console.py
+++ b/rich/_console.py
@@ -1,0 +1,141 @@
+CONSOLE_HTML_FORMAT = """\
+<!DOCTYPE html>
+<head>
+<meta charset="UTF-8">
+<style>
+{stylesheet}
+body {{
+    color: {foreground};
+    background-color: {background};
+}}
+</style>
+</head>
+<html>
+<body>
+    <code>
+        <pre style="font-family:Menlo,'DejaVu Sans Mono',consolas,'Courier New',monospace">{code}</pre>
+    </code>
+</body>
+</html>
+"""
+
+CONSOLE_SVG_FORMAT = """\
+<svg width="{total_width}" height="{total_height}" viewBox="0 0 {total_width} {total_height}"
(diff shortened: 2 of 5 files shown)


rich/__init__.py
```diff
@@ -1,7 +1,7 @@
 """Rich text and beautiful formatting in the terminal."""
 
 import os
-from typing import Callable, IO, TYPE_CHECKING, Any, Optional, Union
+from typing import IO, TYPE_CHECKING, Any, Callable, Optional, Union
```
DrBenton, line 4: Not a change I would have introduced myself, but since I had to update this file pre-commit triggered the `isort` check on it, which spotted that slight imports ordering issue :slightly_smiling_face:

rich/__init__.py
```diff
@@ -11,7 +11,7 @@
     from .console import Console
 
 # Global console used by alternative print
-_console: Optional["Console"] = None
+_console_default_singleton: Optional["Console"] = None
```
DrBenton, line 14: I stumbled across that Python behaviour I was not aware of: it seems that in this `__init__.py` file the fact that we have a variable named `_console` conflicts with a `_console.py` module if there is one.
So `global _console` will actually refer to the `_console` **module**, rather than this variable! :open_mouth: 

Renaming it avoids the problem.

Screenshot of the issue:
(not caused by the fact that the local variable is also named `_console` , but by the fact that `get_console()` will return the `_console.py` module instead of the global `_console` variable)
Screenshot from 2022-04-05 15-37-56

rich/_console.py
```diff
@@ -0,0 +1,141 @@
+CONSOLE_HTML_FORMAT = """\
```
DrBenton, line 1: As these 2 `CONSOLE_` variables take quite a lot of space in the `console` module (especially the SVG one) I extracted them to this private module to make the work on the `console` module easier (less scrolling to navigate in this rather long file is always good to have right? :smile: )

If anyone was using them in their own code, with `from rich.console import CONSOLE_SVG_FORMAT` for example, this will still work as these 2 consts are imported in `console.py`

rich/_console.py
```diff
@@ -0,0 +1,141 @@
...
+</body>
+</html>
+"""
+
+CONSOLE_SVG_FORMAT = """\
+<svg width="{total_width}" height="{total_height}" viewBox="0 0 {total_width} {total_height}"
+     xmlns="http://www.w3.org/2000/svg">
+    <style>
+        @font-face {{
+            font-family: "{font_family}";
```
DrBenton, line 27: The font family is now dynamic, and its default name is _"Rich Fira Code"_ - so if anyone uses the Fira Code font in a HTML page but loaded from another source, our own one won't conflict with it :slightly_smiling_face:

rich/_console.py
```diff
@@ -0,0 +1,141 @@
...
+        }}
+        @font-face {{
+            font-family: "{font_family}";
+            src: local("FiraCode-Bold"),
+                 url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Bold.woff2") format("woff2"),
+                 url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Bold.woff") format("woff");
+            font-style: bold;
+            font-weight: 700;
+        }}
+        #{wrapper_id} span {{
```
DrBenton, line 42: The hard-coded `#wrapper` element now has a dynamic ID - its default value, if not specified when creating the SVG, is `rich-svg-wrapper` - and all our CSS selectors are now prefixed with this ID.
This should prevent these selectors from conflicting with the ones defined in the HTML page that embeds this SVG.

People can also provide their own ID - which can be a random value - if they want to.

DrBenton:
Does that address your comments. @willmcgugan ? :slightly_smiling_face: 
[55c80fb](https://github.com/Textualize/rich/pull/2170/commits/55c80fbeb13f8870b00477a51302e5fccd00c34f)

rich/_console.py
```diff
@@ -0,0 +1,151 @@
...
+                    </div>
+                </div>
+            </div>
+        </body>
+    </foreignObject>
+</svg>
+"""
+
+
+def _svg_hash(svg_main_code: str) -> str:
```
willmcgugan, line 146: This probably doesn't belong in a file full of constants.

Perhaps it should be a closure? If we don't need to access it outside of the export svg function.

Can you please add a docstring for this. We use the google format for docstrings. I have a vscode plugin which can partially generate them.

willmcgugan (changes requested):

DrBenton:
@willmcgugan I reverted my renaming of the private `_console` singleton in the `__init__.py` file, renamed the `_console.py` following your (sensible :slightly_smiling_face: ) suggestion, and moved the `_svg_hash` function into the `console` module. This function is also now documented. [e89ec10](https://github.com/Textualize/rich/pull/2170/commits/e89ec1063e1cd13078d582892025dc920c77e66b)

I could put this `_svg_hash` function into the `export_svg` method of the Console class, as it's the only place where it's used, but as this method is already quite long (around 130 lines) I opted for a simple private function in the module itself.
Happy to move it into the `export_svg` method (or elsewhere) if you prefer, of course :slightly_smiling_face: 

We could also simply inline its single line of code into the `export_svg` method. 
I just thought that giving it a name would make better semantics, but it might not be worth its 9 lines of code (1 line of proper code + 8 of doctring) after all :sweat_smile:

willmcgugan (approved):

willmcgugan:
LGTM

PR #2182, opened 2022-04-10
Fix import of OrderedDict on Python <3.9

Dreamsorcerer:
We're getting a failure in our CI now where twine uses rich:
`ImportError: cannot import name 'OrderedDict' from 'typing_extensions'`

I'm guessing that typing_extensions have removed OrderedDict as it is already available in all supported versions of Python. This should instead be imported from the standard library.

### DIFF
--- a/rich/_lru_cache.py
+++ b/rich/_lru_cache.py
@@ -5,7 +5,7 @@ CacheKey = TypeVar("CacheKey")
 CacheValue = TypeVar("CacheValue")
 
 if sys.version_info < (3, 9):
-    from typing_extensions import OrderedDict
+    from typing import OrderedDict
 else:
     from collections import OrderedDict
 


willmcgugan:
Fixed in 12.2.0

Dreamsorcerer:
Just looked through our logs. It is already using 12.2. This is not fixed, you are still trying to import from typing_extensions, which doesn't appear to contain this class.

Dreamsorcerer:
Nevermind, just seen there's a conflict. It's got an old version of typing_extensions for some reason.

max-l:
I'm gerring the bug with 

rich==12.4.4
and python3.8.5

Dreamsorcerer:
What about the typing_extensions version? The package requires a sufficiently recent version, but in our setup we were somehow installing an old/incompatible version.

PR #2192, opened 2022-04-14
Add support for a "fuzzy" choice matcher

azeemba:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

When choices are pretty distinct, it is nice to let the user only type
part of the choice and automatically match the rest. For example
if the choices are "Strike" and "Ball", then if the user types "b" we
know they meant ball.

This commit adds a new Prompt subclass to support this fuzzy matching.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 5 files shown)


azeemba:
@willmcgugan I definitely understand the motivation. Is it okay for me to create a PR to add this to `examples` or maybe even the docs file? 

No worries if not.

PR #2200, opened 2022-04-19
Setting terminal window title

darrenburns:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Adds support for setting the title of the terminal window.

There is no means within Rich of "resetting" the window title to its
previous value, meaning the title you set will persist even after your application
exists.

`fish` shell resets the window title before and after each command by default,
negating this issue. Windows Terminal and command prompt will also reset the title for you.
Most other shells and terminals, however, do not do this.

Some terminals may require configuration changes before you can set the title.
Some terminals may not support setting the title at all.

Other software (including the terminal itself, the shell, custom prompts, plugins, etc.)
may also set the terminal window title. This could result in whatever value you write
using this method being overwritten.


rich/_windows_renderer.py
```diff
@@ -51,3 +51,6 @@ def legacy_windows_render(buffer: Iterable[Segment], term: LegacyWindowsTerm) ->
                         term.erase_start_of_line()
                     elif mode == 2:
                         term.erase_line()
+                elif control_type == ControlType.SET_WINDOW_TITLE:
```
willmcgugan, line 54: Great! Did you test this on the Windows laptop?

willmcgugan (approved):
LGTM

PR #2210, opened 2022-04-21
progress speed

willmcgugan:
Adds a `show_speed` parameter to `TaskProgressColumn` which displays speed in iterations per second when the total is not known in advance.

Screenshot 2022-04-21 at 16 32 58

### DIFF
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -717,7 +717,7 @@ class TaskProgressColumn(TextColumn):
         )
 
     @classmethod
-    def render_speed(cls, speed: float | None) -> Text:
+    def render_speed(cls, speed: Optional[float]) -> Text:
         """Render the speed in iterations per second.
 
         Args:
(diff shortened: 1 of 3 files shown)


rich/progress.py
```diff
@@ -676,7 +678,18 @@ def render(self, task: "Task") -> Text:
 
 
 class TaskProgressColumn(TextColumn):
-    """A column displaying the progress of a task."""
+    """Show task progress as a percentage.
```
DrBenton, line 681: ah, sorry, I should have documented that myself when I added it! :bow:

DrBenton (approved):
LGTM! :slightly_smiling_face:

darrenburns (changes requested):

rich/progress.py
```diff
@@ -115,6 +115,7 @@ def track(
     pulse_style: StyleType = "bar.pulse",
     update_period: float = 0.1,
     disable: bool = False,
+    show_speed: bool = True,
```
darrenburns, line 118: Default/constructor mismatch here too

darrenburns:
Two small docstring fixes needed but otherwise LGTM :)

PR #2211, opened 2022-04-22
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,5 +1,8 @@
 name: codespell
 on: [pull_request, push]
+permissions:
+  contents: read
+
 jobs:
   codespell:
     runs-on: ubuntu-latest
--- a/.github/workflows/readmechanged.yml
+++ b/.github/workflows/readmechanged.yml
@@ -7,6 +7,9 @@ on:
     paths:
       - 'README.md'
 
+permissions:
+  contents: read
+
 jobs:
   send_notification:
     runs-on: ubuntu-latest


tdung (approved):

henryiii:
Like in https://github.com/pypa/cibuildwheel/pull/1090 - I believe the only problem (and the only thing listed in the linked articles) is `pull_request_target`, normal `pull_requests` are safe. If they weren't, Dependabot or something else would likely be employed with warnings.

naveensrinivasan:
> Like in [pypa/cibuildwheel#1090](https://github.com/pypa/cibuildwheel/pull/1090) - I believe the only problem (and the only thing listed in the linked articles) is `pull_request_target`, normal `pull_requests` are safe. If they weren't, Dependabot or something else would likely be employed with warnings. (Happy to be corrected if wrong)

Sounds good. I can close this.

tdung (commented):
Xx

tdung (commented):
XN

tdung (commented):
H

PR #2217, opened 2022-04-24
optimization for Segment.divid

willmcgugan:
Micro-optimization for Segment.divide

~20% improvement for a test case (in profile_divide.py), although it is heavily dependant on source data


rich/segment.py
```diff
@@ -611,9 +611,14 @@ def divide(
             yield []
         pos = 0
 
+        _cell_len = cell_len
         for segment in segments:
-            while segment.text:
-                end_pos = pos + segment.cell_length
```
willmcgugan, line 616: This code is in the critical path in Textual. Small wins would have tangible improvements.

Segment.cell_length is a hidden function call that unpacks segment and calls `cell_len`.  We can avoid that little bit of work by unpacking and doing the job of `cell_length` here.

DrBenton (approved):
Nice one! :rocket:

PR #2225, opened 2022-04-26
[traceback] Fix "missing whitespace" bug

### LINKED ISSUE #2206: Missing whitespace in traceback
In tracebacks sometimes there is a missing blank line between traceback frames. Not sure what causes it.

DrBenton:
It turns out that we intentionally don't add such a whitespace at the beginning of the trace rendering, but we were actually doing so for the 2nd item of the stack rather than the 1st one

closes #2206 

## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

#### Before:

We can see a blank line before the 1st frame of the stack, and a missing blank line before the 2nd one:
Screenshot from 2022-04-26 11-37-56

#### After:

The 2nd frame is now properly displayed with a blank line before it, while the 1st frame starts right away at the beginning of the Panel:
Screenshot from 2022-04-26 11-37-35


willmcgugan (approved):
LGTM

PR #2226, opened 2022-04-26
[traceback] Add traceback frame opt-out via a `_rich_traceback_omit = True` machinery

### LINKED ISSUE #2207: Hide frame from tracebacks
Sometimes it is useful to completely omit a frame from a traceback. Suggest a local variable with a name like `_rich_traceback_omit`

DrBenton:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

closes #2207

### DIFF
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -367,6 +367,8 @@ class Traceback:
                 if filename and not filename.startswith("<"):
                     if not os.path.isabs(filename):
                         filename = os.path.join(_IMPORT_CWD, filename)
+                if frame_summary.f_locals.get("_rich_traceback_omit", False) is True:
+                    continue
                 frame = Frame(
                     filename=filename or "?",
                     lineno=line_no,
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -367,7 +367,7 @@ class Traceback:
                 if filename and not filename.startswith("<"):
                     if not os.path.isabs(filename):
                         filename = os.path.join(_IMPORT_CWD, filename)
-                if frame_summary.f_locals.get("_rich_traceback_omit", False) is True:
+                if frame_summary.f_locals.get("_rich_traceback_omit", False):
                     continue
                 frame = Frame(
                     filename=filename or "?",
@@ -385,7 +385,7 @@ class Traceback:
                     else None,
                 )
                 append(frame)
-                if "_rich_traceback_guard" in frame_summary.f_locals:
+                if frame_summary.f_locals.get("_rich_traceback_guard", False):
                     del stack.frames[:]
 
             cause = getattr(exc_value, "__cause__", None)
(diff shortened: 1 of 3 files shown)


tests/test_traceback.py
```diff
@@ -307,6 +308,41 @@ def test_suppress():
         assert "foo" in traceback.suppress[1]
 
 
+@pytest.mark.parametrize(
+    "rich_traceback_omit_for_level2,expected_frames_length,expected_frame_names",
+    (
+        # fmt: off
+        [True, 3, ["test_rich_traceback_omit_optional_local_flag", "level1", "level3"]],
+        [False, 4, ["test_rich_traceback_omit_optional_local_flag", "level1", "level2", "level3"]],
+        # fmt: on
```
DrBenton, line 317: I added these `# fmt: ` annotations just to avoid Black changing the second test case so that it spawns across multiple lines when the 1st one doesn't, as they are functionally the same thing
But happy to let Black format _everything_ and remove those if you prefer, of course :slightly_smiling_face:

darrenburns: I like this. Black completely ruins the readability of some parametrised tests

willmcgugan: Fine by me!

rich/traceback.py
```diff
@@ -367,6 +367,8 @@ def safe_str(_object: Any) -> str:
                 if filename and not filename.startswith("<"):
                     if not os.path.isabs(filename):
                         filename = os.path.join(_IMPORT_CWD, filename)
+                if frame_summary.f_locals.get("_rich_traceback_omit", False) is True:
```
willmcgugan, line 370: The `_rich_traceback_guard` mechanism is triggered by just the presence of the local (i.e. doesn't care about the value). I think this should be consistent.

On reflection I think we should look at the value to make it easier to add a condition. The default should be `False` if the local isn't defined.

Something like this:

```python
if getattr(frame_summary.f_locals, "_rich_traceback_omit", False):
   ...
```

The `"_rich_traceback_guard"` should also be updated in the same way.

willmcgugan (changes requested):

rich/traceback.py
DrBenton: I see! However, `f_locals` being a Python dictionary it doesn't support the `getattr` protocol if I'm not wrong?

But I updated the `_rich_traceback_guard` management so that both look the same. Also, they are both activated as soon as the value is defined in the locals and its value is _true-ish_ value, rather than strict `True` :slightly_smiling_face: 
--> done there: [f0ca6ff](https://github.com/Textualize/rich/pull/2226/commits/f0ca6ff2bc0dbca2ffcd2a5114d04b92679ecbb3)

willmcgugan: My bad, should of course be `.get`.

willmcgugan (approved):

PR #2227, opened 2022-04-26
Return a Task object from add_task, not an id

t-mart:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

(Missing items would be completed on final submission. Currently, I'm just exploring this idea.)

## Description

*Preface: This pull request is practically just a feature request (more open to discussion), but with a quick attempt to see how difficult the work would be. I don't mean for the code/design to be final. Just a proof-of-concept.*

I would like `add_task` to return a `Task` object instead of an opaque identifier `TaskId`. Further, `Task` objects should be able to update, advance, and reset themselves. My reasoning is as follows:

- It is easier to pass around a single reference to a `Task` instead of two references to a `TaskId` and `Progress` object. Having "rich-ified" a few personal projects, the tedium of passing both these things deep down a call stack adds up.
- `TaskId` (ints) aren't useful to users, but `Task` objects are. For example, I can't ask a `TaskId` what its total is, but a `Task` does know.
- There's no (public) way to translate `TaskId`s into `Task`s currently. Even if you know that the private-ish `self._tasks` dict is where they're kept, you further have to know that the `TaskId` is really just a key into that dict

## Implementation

The quick summary is:

-  Return a `Task` object from `Progress.add_task`, not an id
- Add a `progress` attribute to `Task` objects. On initialization, this gets set to the `Progress` object that is managing the task. 
  - *(Not attempted option)* Make the attribute private (`_progress`)?
- Add `start()`, `stop()`, `update()`, `reset()`, `advance()` methods to the Task class. Currently, these are just thin wrappers to the method on the progress object. For example `def start(...): self.progress.start(...)`.
  - *(Not attempted option)* Strip these methods from `Progress` altogether, and have the implementations only live under `Task`. This simplifies `Progress`'s interface and makes it less confusing to figure out the "right way" to do something.
- Update callers of `add_task` to use the `task.id` instead of just the old returned `task_id`. This includes in tests, which still pass 100% with this PR's code.

Because these changes change the interface of `add_task`, this change would be **backwards incompatible**.

Later work would include new tests and documentation updates.

### DIFF
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -924,6 +924,9 @@ class Task:
     id: TaskID
     """Task ID associated with this task (used in Progress methods)."""
 
+    progress: "Progress"
+    """Associated Progress object"""
+
     description: str
     """str: Description of the task."""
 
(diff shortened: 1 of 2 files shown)


willmcgugan:
Since this is a breaking change I couldn't accept it unless there was a very compelling reason to do so.

Exposing the Task object is probably a bad idea because there are some internal attributes there, that a user might expect to modify and have a meaningful effect.

What might work is having an object that wraps TaskID and adds public methods. But this object would have not break the current API. i.e. `progress.update(my_task, ...` should continue to work in addition to  `my_task.update(`.

I'm going to close this PR, but happy to continue the discussion on Discussions tab.

PR #2244, opened 2022-05-03
Respect dim style on SVG export

### LINKED ISSUE #2237: SVG export colors wrong
When exporting syntax to SVG the color for the numbers and guide lines were incorrect. In the terminal they are a mid grey, but in the export they are closer to pure white. If I was to guess I would say that the dim attribute isn't respected here.

terminal

svg

sbarrios93:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [x] I've updated ~CHANGELOG.md~ and CONTRIBUTORS.md where appropriate.
- [_Not sure if needed_ ] I've added tests for new code. 
- [x] I accept that @willmcgugan may be pedantic in the code review. (Please do!!)🥸

## Description

Fixes #2237 

`dim` style has property `self.color` as `None`, making [this block](https://github.com/Textualize/rich/blob/bb00fc96ee09525e5b5f2ffe7762138b5999e949/rich/console.py#L2300) append a new color style after the correct one. (Look at the `r2` style on a svg file where the dim style is not being formatted correctly. There is more than one foreground attribute being set)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -5,6 +5,7 @@ The following people have contributed to the development of Rich:
 <!-- Add your name below, sort alphabetically by surname. Link to Github profile / your home page. -->
 
 - [Patrick Arminio](https://github.com/patrick91)
+- [Sebastian Barrios](https://github.com/sbarrios93)
 - [Gregory Beauregard](https://github.com/GBeauregard/pyffstream)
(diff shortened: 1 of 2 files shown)


willmcgugan:
Thanks, be we're rewriting the SVG export functionality.

Issue #2251, opened 2022-05-04
[BUG] Rich tracebacks raise another exception when using pyinstaller


gormaniac:
**Describe the bug**

A Python `code` object when Python is run by pyinstaller, does not have a `co_filename` attribute that points to an absolute path, instead it is a relative path. For example it will be `rich/console.py` instead of `/abs/path/to/project/rich/console.py`.

Because the path is relative, we will hit [this branch](https://github.com/Textualize/rich/blob/master/rich/traceback.py#L369) of code in `rich.traceback.Traceback.extract`. This code prepends the current working directory to the relative path in the `code` object. This file will likely never exist. It is possible that the current working directory happens to be where the extracted pyinstaller bundle is, but it is unlikely.

Since the generated filename doesn't exist, rich will throw an exception stating `[Errno 2] No such file or directory:`. This muddies the relevant traceback and suggests to users that the issue in whatever code uses rich is related to a missing file rather than whatever raised the original exception.

It is worth noting that in tests I've run without pyinstaller, I've found that `code.co_filename` attributes point to absolute paths so this isn't a big issue for typical Python code.

I've come up with a pyinstaller specific fix but I don't know if you'd like to include a py2exe related fix as well (I'm also unsure if py2exe even has this issue). I can submit a PR for this if you'd like. We'd just need to change the appropriate bit in `rich.traceback.Traceback.extract` to the below:
```python
filename = frame_summary.f_code.co_filename
if filename and not filename.startswith("<"):
    if not os.path.isabs(filename):
        # Use has/getattr because `frozen` won't exist if not pyinstaller
        if hasattr(sys, 'frozen') and getattr(sys, 'frozen', False) == True:
            filename = os.path.join(sys._MEIPASS, filename)
        else:
            filename = os.path.join(_IMPORT_CWD, filename)
```

See the [pyinstaller docs](https://pyinstaller.org/en/stable/runtime-information.html#run-time-information) for info on `sys.frozen` and `sys._MEIPASS`.

Build a pyinstaller bundle with the below code and run it. Change directories and run it again. After both commands, try opening the file listed in the traceback, it shouldn't exist. You will also see an error that says the non-existent file doesn't exist, which isn't what we tried to display in our code.
```python
from rich.console import Console
console = Console()

try:
    raise Exception('A detailed error message')
except Exception:
    console.print_exception(show_locals=True)
```
```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
[...]

PR #2253, opened 2022-05-04
Support pyinstaller relative paths in Traceback (#2251)

gormaniac:
## Type of changes

- [X] Bug fix

## Checklist

## Description

In `rich.traceback.Traceback.extract`, if `frame_summary.f_code.co_filename` isn't an absolute path, we now check if we're running in a pyinstaller bundle, and will fabricate a file name to avoid errors. See #2251 for details.


gormaniac:
Struggling to mock a traceback for a test. We can't just use one from `sys.exc_info` as we wouldn't hit the new code, we need a relative path. All of the attributes we need on a traceback are read only.

willmcgugan:
Closing. Assumed stale. Feel free to reopen if you want to continue work on it.

PR #2259, opened 2022-05-06
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>


henryiii:
I would recommend ignoring minor updates in the official actions, otherwise this would become annoying. Something like this:

```
version: 2
updates:
  # Maintain dependencies for GitHub Actions
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    ignore:
      # Official actions have moving tags like v1
      - dependency-name: "actions/*"
        update-types: ["version-update:semver-minor", "version-update:semver-patch"]
```

See https://scikit-hep.org/developer/gha_basic#updating or various repositories like pypa/cibuildwheel.

naveensrinivasan:
> ```yaml
> ignore:
>       # Official actions have moving tags like v1
>       - dependency-name: "actions/*"
>         update-types: ["version-update:semver-minor", "version-update:semver-patch"]
> ```

Updated it! PTAL Thanks

PR #2268, opened 2022-05-10
Fix redundant statements

lqhuang:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other (cleanup)

## Checklist

  - [x] CONTRIBUTORS.md is updated.
  - [ ] There is no new feature or bugfix, so no need to update CHANGELOG.md.

## Description

I found some generally duplicated functional snippets.

I thought in `rich/console.py`, [line 662-664, line 676-687](https://github.com/Textualize/rich/blob/1772458f80924c78fa3a8d0260bca9762ad86e18/rich/console.py#L676-L687) and [line 986-999](https://github.com/Textualize/rich/blob/1772458f80924c78fa3a8d0260bca9762ad86e18/rich/console.py#L986-L999)(`size` property) seem have exactly same effects. I checked other places to find related invokes, no more special case, so I have removed one of them in this PR.

Please review and confirm it actually redundant.

Thanks!

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -659,9 +659,6 @@ class Console:
             width = width or 93
             height = height or 100
 
-        self._width = width
-        self._height = height
-        self.soft_wrap = soft_wrap
         self.tab_size = tab_size
         self.record = record
         self._markup = markup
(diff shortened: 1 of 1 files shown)


rich/console.py
```diff
@@ -673,18 +673,6 @@ def __init__(
             if legacy_windows is None
             else legacy_windows
         )
-        if width is None:
```
willmcgugan, line 676: This block is not redundant. It the size from being calculated ever time the size property is used.

lqhuang: Thanks for review! I check it again, it definitely will result in different values in some cases. Actually, I just want to clean up [Line 662-664](https://github.com/Textualize/rich/blob/1772458f80924c78fa3a8d0260bca9762ad86e18/rich/console.py#L662-L664) at first, then I glanced some codes in `def size` are very similar, that's my fault.

So how about current updates?

willmcgugan (approved):

willmcgugan:
Thanks

PR #2270, opened 2022-05-11
[console][typing] Extract color-system-as-a-string to a type literal

DrBenton:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

- [ ] I've added tests for new code. _(typing-only modification, so not really testable)_

## Description

While working on integration tests on Textual I found myself duplicating the `Literal["auto", "standard", "256", "truecolor", "windows"]` literal of the Console in my derived class. This is because we're using an inline literal type for the color-system-as-a-string argument we can pass to the Console constructor.
As it's inlined in the constructor signature, we cannot re-use this literal it elsewhere (in Textual for example).

With this tiny PR, Python code outside of Rich can now use this literal :slightly_smiling_face:

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -522,14 +522,19 @@ def _is_jupyter() -> bool:  # pragma: no cover
         return False  # Other type (?)
 
 
-COLOR_SYSTEMS = {
+ConsoleColorSystem = Literal["auto", "standard", "256", "truecolor", "windows"]
+
+
+COLOR_SYSTEMS: Dict[ConsoleColorSystem, ColorSystem] = {
     "standard": ColorSystem.STANDARD,
     "256": ColorSystem.EIGHT_BIT,
     "truecolor": ColorSystem.TRUECOLOR,
     "windows": ColorSystem.WINDOWS,
 }
 
-_COLOR_SYSTEMS_NAMES = {system: name for name, system in COLOR_SYSTEMS.items()}
+_COLOR_SYSTEMS_NAMES: Dict[ColorSystem, ConsoleColorSystem] = {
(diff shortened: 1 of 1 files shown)


rich/console.py
```diff
@@ -522,14 +522,19 @@ def _is_jupyter() -> bool:  # pragma: no cover
         return False  # Other type (?)
 
 
-COLOR_SYSTEMS = {
+ConsoleColorSystem = Literal["auto", "standard", "256", "truecolor", "windows"]
+
+
+COLOR_SYSTEMS: Dict[ConsoleColorSystem, ColorSystem] = {
```
DrBenton, line 528: just adding more specific typing to these 2 dicts while I'm there - we can do this now that the literal type has its own existence :slightly_smiling_face:

rich/console.py
```diff
@@ -522,14 +522,19 @@ def _is_jupyter() -> bool:  # pragma: no cover
         return False  # Other type (?)
 
 
-COLOR_SYSTEMS = {
+ConsoleColorSystem = Literal["auto", "standard", "256", "truecolor", "windows"]
```
willmcgugan, line 525: This name isn't too descriptive. Perhaps `ColorSystemName` ?

It's also not too accurate, given that `auto` is not a color system, it tells the Console to auto detect one of the other values.

willmcgugan (changes requested):
I'm not convinced this is worthwhile (see comments)

I'd suggest moving the burden back to Textual. Defining a literal there, but bear in mind it is for internal use. I.e. The user will be unlikely to see this.

DrBenton:
Alright! :slightly_smiling_face:

PR #2296, opened 2022-05-25
Text wrapping edgecase

### LINKED ISSUE #2273: [BUG] Text wrapping edge case
Discovered a possible edge case with text wrapping, probably connected to when a "word" is greater than the available width.

Consider the following:

```python
console.print("Hello, World...123 ", width=10)
```

This produces:

```
Hello, Wor
ld...123
```

I think this occurs because "World...123" doesn't fit in to 10 characters, so folds on to the next line.

I would expect there to be an additional new line after the Hello, like this:

```
Hello,
World...12
3
```

darrenburns:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Fixes #2273


DrBenton (approved):
Well done! :detective:

rich/cells.py
```diff
@@ -115,7 +115,7 @@ def chop_cells(text: str, max_size: int, position: int = 0) -> List[str]:
     characters = [
         (character, _get_character_cell_size(character)) for character in text
     ]
-    total_size = position + 1
+    total_size = position
```
DrBenton, line 118: Always feels good to remove such occurrences of `+ 1` - they always feel a bit too magical, right? :smile: 
(and often not the right kind of magic :crystal_ball: :mage: :japanese_ogre: )

DrBenton (approved):

willmcgugan (approved):
LGTM

willmcgugan:
The system failed for this PR. I'm pretty I merged a failing test.

PR #2302, opened 2022-05-26
Add support to control pytest test, verbosity and coverage

tomers:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate. --> I've added it in earlier PR
- [] I've added tests for new code. --> Infra change. No need for added tests

## Description

Added ability to control which tests are executed (all by default), in what verbosity level to run (`-vv` by default), and whether to run code coverage (run coverage by default) when running `make test`.

Example:
```
$ make test VERBOSE=1 COVERAGE=0 TEST_KW=test_highlight_iso8601_regex
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,13 @@
+TEST_KW?=
+VERBOSE?=1
+COVERAGE?=1
+
 test:
-	TERM=unknown pytest --cov-report term-missing --cov=rich tests/ -vv
+	TERM=unknown pytest \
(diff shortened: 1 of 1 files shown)


willmcgugan:
Doesn't seem like much of an improvement over copying the command. And it's not very discoverable. Thanks anyway.

PR #2330, opened 2022-06-07
Substituting plain headed box when show_header=False on Table

### LINKED ISSUE #2312: [BUG] If `show_header=False` the next row will have the header border style
img

darrenburns:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Ensure that if the user passes `show_header=False` to `Table`, then we don't use a `Box` which assumes the first row is a header row.

(Note the examples below aren't of the _exact_ same table, only the first row of the box is relevant anyway)

### After
The first line is smooth because `show_header=False`, so the box used was updated to one that doesn't use different header edges.

image

### Before
The first line uses thicker lines, because the HEAVY_HEAD box is used by default, even though there's no header.

image

Fixes #2312


.pre-commit-config.yaml
```diff
@@ -14,7 +14,6 @@ repos:
       - id: check-yaml
       - id: end-of-file-fixer
       - id: mixed-line-ending
-      - id: trailing-whitespace
```
darrenburns, line 17: This incorrectly removed some whitespace inside a multiline string (inside a `Box`), which led to errors.

Issue #504, opened 2020-12-16
Line width of logs is highly truncated on Jupyter Lab


Tzeusy:
Running with the RichHandler() in Jupyter Lab runs into issues where the output does not seem to be able to access the full width of the Jupyter terminal, while the original logging stream handler performs fine.

I've tried messing with the available params for RichHandler() but none of them seem to solve the issue (in this screenshot you can see I've tried `traceback_width`.

![image](https://user-images.githubusercontent.com/21179283/102292266-c7c67280-3f7f-11eb-9801-d351178ce7f9.png)

**To Reproduce** 
Sample code:
```
import sys
import logging
from rich.logging import RichHandler

logger = logging.getLogger()

formatter = logging.Formatter('%(asctime)s - %(name)s:%(lineno)d - %(levelname)s - %(message)s')
shandler = logging.StreamHandler(sys.stdout)
shandler.setFormatter(formatter)
logger.addHandler(shandler)
logger.addHandler(RichHandler(highlighter=None, rich_tracebacks=True, tracebacks_show_locals=True, tracebacks_width=300, locals_max_string=100))
logger.setLevel(logging.INFO)

logger.info("This is a line that is 168 characters long"*4)

# Traceback box also gets truncated
try:
    1/0
except:
    logger.exception("Error! A really long error message that describes the problem...")
```

**Platform**
Chrome

willmcgugan:
Rich has no way to know how wide the browser is and has to render content with a fixed width. In a terminal it will extend to the full available width.

If you really want to use the logging handler in Jupyter, maybe set the time format to "%X" and disable the path display.

PR #2332, opened 2022-06-09
Allow Jupyter to vary width/height using COLUMNS/LINES

AntonyMilneQB:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

I'm opening this is a very small PR just to get initial feedback on whether it sounds like a good idea. If the response is positive then I will update tests as required.

`COLUMNS` and `LINES` environment-variables can already be used to set `Console._width` and `Console._height`: https://github.com/Textualize/rich/blob/51889e2e663bbda8d59b2a68f23322f578681cea/rich/console.py#L676-L683

However, these environment variables are ignored when using Jupyter. If `width` or `height` are not explicitly set in `__init__`  then rich assumes default values: https://github.com/Textualize/rich/blob/51889e2e663bbda8d59b2a68f23322f578681cea/rich/console.py#L658-L660

This PR makes it so that `COLUMNS` and `LINES` can be used to set `Console._width` and `Console._height` even when  `self.is_jupyter` is `True`. `width`/`height` arguments still take precedence over the environment variables, exactly the same as when outside Jupyter.

## Further context

We are very excited to start usubg this beautiful library in kedro, an open source data science pipeline framework (see https://github.com/kedro-org/kedro/issues/1464 if you're interested). A core part of that is to use the rich logging handler, which we love.

When the rich logging handler is used in Jupyter, the fixed width of 93 characters wraps messages too narrowly for our needs. This has been brought up before (https://github.com/Textualize/rich/issues/504 https://bytemeta.vip/repo/willmcgugan/rich/issues/1870 https://github.com/Textualize/rich/discussions/344). From these I understand that rich doesn't know the width of the screen in Jupyter and why another custom logger will not be added to the package. (Although I'm not sure exactly where the default value of `width = 93` comes from, and looking through git history couldn't answer this? Would be curious to know.) We also do not want users to need to supply a custom logging handler just to accommodate this case.

By making `COLUMNS` and `LINES` recognised in Jupyter, I think it would immediately address several people's requirements without any real extra complexity being added to rich. So this seems like an easy win to me, but I'm very happy to receive an explanation of what I've missed here if it's not a good idea!

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -656,8 +656,8 @@ class Console:
 
         self.is_jupyter = _is_jupyter() if force_jupyter is None else force_jupyter
         if self.is_jupyter:
-            width = width or 93
-            height = height or 100
+            width = width or int(self._environ.get("COLUMNS", 93))
+            height = height or int(self._environ.get("LINES", 100))
 
         self.soft_wrap = soft_wrap
         self._width = width
--- a/rich/console.py
+++ b/rich/console.py
@@ -656,8 +656,10 @@ class Console:
 
         self.is_jupyter = _is_jupyter() if force_jupyter is None else force_jupyter
         if self.is_jupyter:
-            width = width or int(self._environ.get("COLUMNS")) or 93
-            height = height or int(self._environ.get("LINES")) or 100
+            jupyter_columns = self._environ.get("JUPYTER_COLUMNS", "")
+            width = width or (int(jupyter_columns) if jupyter_columns.isdigit() else 93)
+            jupyter_lines = self._environ.get("JUPYTER_LINES", "")
+            height = height or (int(jupyter_lines) if jupyter_lines.isdigit() else 100)
 
         self.soft_wrap = soft_wrap
         self._width = width
--- a/rich/console.py
+++ b/rich/console.py
@@ -72,6 +72,8 @@ if TYPE_CHECKING:
     from .live import Live
     from .status import Status
 
+JUPYTER_DEFAULT_COLUMNS = 115
+JUPYTER_DEFAULT_LINES = 100
 WINDOWS = platform.system() == "Windows"
 
 HighlighterType = Callable[[Union[str, "Text"]], "Text"]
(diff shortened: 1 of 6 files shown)


rich/console.py
```diff
@@ -656,8 +656,8 @@ def __init__(
 
         self.is_jupyter = _is_jupyter() if force_jupyter is None else force_jupyter
         if self.is_jupyter:
```
AntonyMilneQB, line 658: For anyone else puzzling over `self.legacy_windows`, which is subtracted from `width` in other places: this isn't needed at all here because `self.legacy_window = 0` when `self.is_jupyter == False`.

willmcgugan:
The default of `93` was chosen by experimenting with Jupyter Notebook. If I set it too high, it will scroll, too low and it wastes horizontal space. 93 was a compromise that fitted well on a mid-sized screen at the time.

The issue with using `COLUMNS` and `LINES` is that those env vars effect the terminal. You may not want to have the same value for used both terminal and notebook. Perhaps a dedicate `JUPYTER_COLUMNS` env var to target Jupyter specifically?

rich/console.py
```diff
@@ -656,8 +656,12 @@ def __init__(
...
-            width = width or 93
-            height = height or 100
+            jupyter_columns = self._environ.get("JUPYTER_COLUMNS", "")
```
AntonyMilneQB, line 659: N.B. this works the same way, but is written slightly differently, from how `COLUMNS`/`LINES` are handled in a couple of other places in the file, e.g.
```
columns = self._environ.get("COLUMNS")
        if columns is not None and columns.isdigit():
            width = int(columns)
```
The reason for this was to not overcomplicate the `if` with another conditional, given it's already following an `or`. I could change the other `COLUMNS`/`LINES` parts in this file to match what I've done here or vice versa.

tests/test_jupyter.py
```diff
@@ -3,5 +3,30 @@
 
 def test_jupyter():
     console = Console(force_jupyter=True)
-    assert console.width == 93
+    assert console.width == 115
+    assert console.height == 100
     assert console.color_system == "truecolor"
+
+
+def test_jupyter_columns_env():
```
AntonyMilneQB, line 11: These are heavily based on the console tests.

AntonyMilneQB:
Thanks for the comments @willmcgugan. This should be ready to review now. I played around for a while with both Jupyter notebook and lab and found that a width 115 seemed to work well even on a fairly small screen.

Two small questions on this file unrelated to this change:
1. I'm curious why you do "Copy of os.environ allows us to replace it for testing" rather than monkeypatching environment  variables in tests?
2. Currently you require both `COLUMNS` and `LINES` to be set for either one to have an effect. Is this intentional or would you accept a PR that made them independent? I want to set `COLUMNS` to something relatively high in CI (which is a dumb terminal so quite narrow by default) but also have to set `LINES` for this to have an effect currently. https://github.com/Textualize/rich/blob/51889e2e663bbda8d59b2a68f23322f578681cea/rich/console.py#L963

rich/console.py
```diff
@@ -656,8 +656,12 @@ def __init__(
...
+            width = width or (
+                int(jupyter_columns) if jupyter_columns.isdigit() else 115
```
willmcgugan, line 661: let's move these numbers to the module level as constants. Suggest `JUPYTER_DEFAULT_COLUMNS=115`

BTW Have you tested this on lab and notebook, and could you send screenshots so I can see it fits comfortably...

rich/console.py
willmcgugan: We're looking up `JUPYTER_COLUMNS` every time, even when `width` is non zero. 

Suggest structuring it so that it is clear we only use it when `width` is not set. i.e.

`if not width:`

willmcgugan (changes requested):

AntonyMilneQB:
@willmcgugan Thanks for the review. I've made the changes as requested.

Here's what a width of 115 looks like in Jupyter notebook:
image

And in Jupyter lab:
image

willmcgugan (approved):

AntonyMilneQB:
Thank you @willmcgugan! Do you have a rough idea of when this might be merged and when a new rich release would be? Thank you!

willmcgugan:
Should have a new release within a week.

willmcgugan:
Should have a new release within a week.

PR #2341, opened 2022-06-15
[syntax] add a `stylize_ranges` method to the Syntax class

### LINKED ISSUE #2329: Add stylised range support to Syntax
In order to support https://github.com/Textualize/rich/issues/2286 we will need to be able to add arbitrary ranges to Syntax.

A new API method to apply a style from (LINE A, COLUMN A) to (LINE B, COLUMN B).

DrBenton:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

With this new API we can apply a style from (LINE A, COLUMN A) to (LINE B, COLUMN B) - which is something we will need to be able to add arbitrary ranges to Syntax.

Fixes #2329


rich/syntax.py
```diff
@@ -193,6 +205,24 @@ def get_background_style(self) -> Style:
         return self._background_style
 
 
+class SyntaxPosition(NamedTuple):
```
DrBenton, line 208: I could have used a basic `Tuple[int, int]` for that, but I thought that some additional semantics (and reminding the user that line numbers start at 1 rather than 0 while we're here) could be nice?

rich/syntax.py
```diff
@@ -193,6 +205,24 @@ def get_background_style(self) -> Style:
...
+
+class SyntaxHighlightRange(NamedTuple):
+    """
+    A range to highlight in a Syntax object.
+    The default style, if none is provided, is to highlight the range in bold.
+    """
+
+    start: SyntaxPosition
+    end: SyntaxPosition
+    style: Style = Style(bold=True)
```
DrBenton, line 223: Not sure about that one - rather than having such an opinionated (and kinda abstract) default Style, maybe the Style should always be specified? :thinking:

DrBenton:
@willmcgugan @darrenburns I haven't updated the CHANGELOG yet, as I have some doubts about some parts of this PR - that's why I'm summoning you folks for a first review, so we can arbitrate together about some of these doubts :slightly_smiling_face:

DrBenton:
As discussed with @willmcgugan :
 - The feature should not be part of the constructor, but exposed as a dedicated `stylize_ranges` method
 - We should try to get better performance by avoiding the loop that goes through every character, by applying the ranges after the code has been split into several lines instead

DrBenton:
@willmcgugan I ended up opting for another approach: instead of operating on the code split in lines (which is a state we get only in some cases, such as displaying the line numbers) I kept my previous approach, but improved its performance: :fast_forward:  [75315fa](https://github.com/Textualize/rich/pull/2341/commits/75315fabc5fedbbd441078f9e8c93d75c96f4ccc)

The improvement comes from the function that returns a "code index" from a "code position" : e.g. for this code snippet the position _"line 3 / character 15"_ must be translated to index _"168"_:
```py
def loop_first_last(values: Iterable[T]) -> Iterable[Tuple[bool, bool, T]]:
    """Iterate and generate a tuple with a flag for first and last value."""
    iter_values = iter(values)
```
With the new implementation of this function, instead of iterating on every character of the code to be able to do this conversion I first store the length of each line of code in a list (which is rather cheap to do, as it's simple calls to `str.find()`), and then use that information to do the conversion.
(we need to do this conversion to be able to pass indexes to the `Text.stylize()` method)

By doing so, in order to stylise some parts of a 200 lines of code snippet (the content of _"rich/columns.py"_) I get a performance penalty that is rather low.
Performance with `%timeit` on a laptop that is doing other things in parallel is not perfect measurement of course :shrug: , but in this test rendering it with several stylised ranges takes 22.3ms in average, versus 21.6ms without stylised ranges:
Screenshot from 2022-06-16 16-20-46

DrBenton:
@willmcgugan I thought of a better way to calculate the index in the string from a line number and column index - and it seems to be a little bit more efficient as a result - the overhead of stylising ranges is now pretty minimal :slightly_smiling_face: [4d80b3a](https://github.com/Textualize/rich/pull/2341/commits/4d80b3a558239780d824ea69bb8718b81335f302)
Screenshot from 2022-06-17 09-26-23

rich/syntax.py
```diff
@@ -265,6 +295,7 @@ def __init__(
         self.padding = padding
 
         self._theme = self.get_theme(theme)
+        self._stylized_ranges: Sequence[SyntaxHighlightRange] = ()
```
willmcgugan, line 298: This can be a `List` now can't it?

rich/syntax.py
willmcgugan: I'd probably default to a plain tuple in `SyntaxHighlightRange` and name it when you unpack.

```python
start_col, start_row = syntax_highlight_range.start
```

rich/syntax.py
```diff
@@ -574,11 +665,8 @@ def _get_syntax(
             else self.code_width
         )
 
-        ends_on_nl = self.code.endswith("\n")
-        code = self.code if ends_on_nl else self.code + "\n"
-        code = textwrap.dedent(code) if self.dedent else code
-        code = code.expandtabs(self.tab_size)
-        text = self.highlight(code, self.line_range)
+        ends_on_nl, processed_code = self._get_processed_code(self.code)
```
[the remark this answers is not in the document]
willmcgugan: Agree, but I'd probably rename: you aren't just getting the processed code, you are doing the processing. Suggest `process_code`

willmcgugan (changes requested):
Clever solution!

However, I think we can do better at finding the line locations with a regex.

Consider the following:

```python
>>> import re
>>> [match.start() for match in re.finditer("^", "\nfoobar\nbazegg\n", flags=re.MULTILINE)]
[0, 1, 8, 15]
``` 

Otherwise, a few API requests. And would you mind posting a few examples and screenshots on the PR.

DrBenton:
@willmcgugan Thanks for your review! I'm goignto address your comments  :slightly_smiling_face:  

Regarding `re.finditer` I though of using it indeed, instead of implementing this stuff myself,;but I was afraid of a potential performance hit by using regular expressions instead of simple-and-stupid iterative string lookup.
I just did a quick comparison using `%timeit` (which the usual flaws of such a comparison), out of curiosity, and it seems that using `re` is a lot slower?
Screenshot from 2022-06-17 11-55-11

DrBenton:
> I don’t think that’s quite apples to apples. The regex will get the offsets within the code, not the lengths of the lines. I think it will replace 2 functions.

Sorry, I wasn't clear! `_get_lines_length` actually finds the offsets within the code too, and only does a quick additional step on top of it in order to convert this to "line lengths" (so offsets `[3, 8, 30]` become lengths `[3, 5, 22]`). 
But as this additional step is so trivial, `_get_lines_length` can be considered as really doing the same thing than the `re.finditer` approach - i.e. finding the positions of newline characters in the code :slightly_smiling_face:

DrBenton:
@willmcgugan I think I addressed your comments! If it looks good to you I'll add some documentation for this new feature next week :slightly_smiling_face: 
As a side effect, the code is more compact now that we extract the newline characters offsets with a one-liner:
Screenshot from 2022-06-17 16-42-39

rich/syntax.py
```diff
@@ -484,8 +520,28 @@ def tokens_to_spans() -> Iterable[Tuple[str, Optional[Style]]]:
...
+        Adds a custom style on a part of the code, that will be applied to the syntax display when it's rendered.
+        Line numbers are 1-based, while column indexes are 0-based.
+
+        Args:
+            style (Style): The style to apply.
+            start (Tuple[int, int]): The start of the range, in the form `[line number, column index]`.
+            end (Tuple[int, int]): The end of the range, in the form `[line number, column index]`.
+        """
+        self._stylized_ranges.append(
+            SyntaxHighlightRange(style=style, start=start, end=end)
```
willmcgugan, line 542: I think it is better to use positional args here. Granted keyword args are better for self-documenting code, but consistent ordering helps.

rich/syntax.py
```diff
@@ -484,8 +520,28 @@ def tokens_to_spans() -> Iterable[Tuple[str, Optional[Style]]]:
...
             if self.background_color is not None:
                 text.stylize(f"on {self.background_color}")
+
+        if self._stylized_ranges:
+            self._apply_stylized_ranges(code=code, text=text)
+
         return text
 
+    def stylize_range(
+        self, style: Style, start: SyntaxPosition, end: SyntaxPosition
```
willmcgugan, line 530: `style` should be a `StyleType` which is a union of a str and `Style`. The idea is that I can pass either `"bold red"` *or* a Style object.

You can convert a `StyleType` in to a `Style` by calling `console.get_style(style_type)`

rich/syntax.py
```diff
@@ -690,6 +743,81 @@ def _get_syntax(
...
+        Args:
+            code (str): Code to highlight.
+            text (Text): Text instance to apply the style to.
+        """
+        # N.B. using "\n" here is much faster than using metacharacters such as "^" or "\Z":
+        newlines_offsets = [
+            match.start() for match in re.finditer("\n", code, flags=re.MULTILINE)
+        ]
+        # Let's add outer boundaries at each side of the list:
+        newlines_offsets.insert(0, -1)
```
willmcgugan, line 760: Inserts at 0 are inefficient. Could you create the list in a single expression?

```python
newlines_offsets = [0, *(match.start ...), len(code)]
```

rich/syntax.py
```diff
@@ -690,6 +743,81 @@ def _get_syntax(
...
+        processed_code = (
+            textwrap.dedent(processed_code) if self.dedent else processed_code
+        )
+        processed_code = processed_code.expandtabs(self.tab_size)
+        return ends_on_nl, processed_code
+
+
+def _get_code_indexes_for_syntax_positions(
+    newlines_offsets: Sequence[int], positions: Sequence[SyntaxPosition]
+) -> Sequence[Optional[int]]:
```
willmcgugan, line 793: It's generally better to be strict with the return type. You're returning a list, so the return type should probably reflect that.

rich/syntax.py
```diff
@@ -690,6 +743,81 @@ def _get_syntax(
...
+        processed_code = code if ends_on_nl else code + "\n"
+        processed_code = (
+            textwrap.dedent(processed_code) if self.dedent else processed_code
+        )
+        processed_code = processed_code.expandtabs(self.tab_size)
+        return ends_on_nl, processed_code
+
+
+def _get_code_indexes_for_syntax_positions(
+    newlines_offsets: Sequence[int], positions: Sequence[SyntaxPosition]
```
willmcgugan, line 792: Does `positions` need to be a sequence?

AFAICT we only ever call this for 2 positions. I think it may make sense for this function to take a single `SyntaxPosition`, and the caller can call it twice.

rich/syntax.py
```diff
@@ -448,7 +484,7 @@ def highlight(
 
                 def line_tokenize() -> Iterable[Tuple[Any, str]]:
                     """Split tokens to one per line."""
-                    assert lexer
+                    assert lexer  # required to make MyPy happy
```
willmcgugan, line 487: For future devs, the comment should probably explain why. Suggest "we know lexer is not None at this point" or words to that effect.

tests/test_syntax.py
```diff
@@ -1,4 +1,3 @@
-# coding=utf-8
```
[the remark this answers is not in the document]
willmcgugan: It's Python 3 as well (https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations)

But the default is utf-8 so I don't know why this is there!

willmcgugan (changes requested):
Some API changes requested.

DrBenton:
@willmcgugan Thanks for your review! Hopefully I addressed all your comments there: [8e58adc](https://github.com/Textualize/rich/pull/2341/commits/8e58adcab0750406dac30fc22f3183d415c5a559) :slightly_smiling_face:  :crossed_fingers:

DrBenton:
@willmcgugan I removed the odd `-1` from the algorithm, and tested that it also passes successfully on empty code while I was there :slightly_smiling_face:  [b6a4cfa](https://github.com/Textualize/rich/pull/2341/commits/b6a4cfab5a707358fab31e995cc5024aeb76bee0)

willmcgugan (changes requested):
One request...

rich/syntax.py
```diff
@@ -690,6 +740,82 @@ def _get_syntax(
...
+                for match in re.finditer("\n", code, flags=re.MULTILINE)
+            ],
+            len(code) + 1,
+        ]
+
+        for stylized_range in self._stylized_ranges:
+            start = _get_code_index_for_syntax_position(
+                newlines_offsets, stylized_range.start
+            )
+            end = _get_code_index_for_syntax_position(
```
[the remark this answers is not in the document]
DrBenton: Right, we now clamp the column value to the end of the line :slightly_smiling_face: :  [6e63470](https://github.com/Textualize/rich/pull/2341/commits/6e634705b7ff493918080da7053b3127677440d8)

Issue #2344, opened 2022-06-15
[REQUEST] Get elapsed time from Task in Progress


staehle:
Hello! Rich has been super useful, to the point that I'm replacing a very similar home-built multi-progress-bar display solution with it. Now, perhaps this is something that already exists, but it wasn't obvious how to do it.

When using Progress, how do you get the final Task's elapsed time, in order to print that out to a log file?

I'm currently doing this, which while it works, doesn't seem right:

```py
with rich.progress.Progress(*STD_PROGRESS_COLS, expand=True) as progress:
    tasks = []
    for i, thing in enumerate(things_to_do):
        tasks.append(progress.add_task(f"[blue]Thing {i}:[/] {thing}"))

        <threaded_stuff>(thing, tasks[i], progress)

    <threaded_stuff>.join()

    with path_log.open(mode='a') as lfp:
        lfp.write("Final times per task:\n")
        for i, thing in enumerate(things_to_do):
            # pylint:disable=protected-access                         # <-- Feels dirty
            f_time = progress._tasks[tasks[i]].elapsed                # <-- Probably shouldn't do this
            s_time = time.strftime("%Hh:%Mm:%Ss", time.gmtime(f_time))
            lfp.write(f" Task {i}, {thing}: {s_time}")
```

Is there a better way?  I would just add an extra timer, but it seems like I should be able to just get the existing timer values from Task.

If directly accessing the private `_tasks` list is currently the only option, Here's a PR that adds a getter for that value from the Progress object, if this is a better way of doing it: #2345

cceyda:
To show elapsed time you can use `TimeElapsedColumn`
```py
from rich.progress import Progress,TimeElapsedColumn

columns = [
  *Progress.get_default_columns(),
   TimeElapsedColumn()
]
progress =Progress(*columns)
```
which shows the value calculate by `elapsed = task.finished_time if task.finished else task.elapsed` as you have discovered. you can use the `progress.tasks` property instead of `progress._tasks` to feel less dirty 😄

PR #2345, opened 2022-06-15
Progress: Add .get_task_elapsed(TaskID)

staehle:
Issue #2344

## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Hello! Related #2344. I didn't think adding a test for this was specifically necessary, since it's just adding a getter for an existing value.

It's also entirely possible you already have a better way of doing this!

### DIFF
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -1633,6 +1633,18 @@ class Progress(JupyterMixin):
         with self._lock:
             del self._tasks[task_id]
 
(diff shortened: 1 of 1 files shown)


willmcgugan:
A test is *always* necessary, I'm afraid!

staehle:
Fair enough! Before I do that though, just wanted to confirm that this would be an appropriate solution for #2344 ?

PR #2346, opened 2022-06-15
Fixed Style hash

willmcgugan:
Fixes a performance bug in the Style class.

Essentially the hash calculation was inconsistent. Things that were equal would sometimes have different hashes. I noticed this when adding `lru_cache` to `Style.__add__`. What should have been an easy win caused dramatically reduced performance, because it broken the mechanics of storing Styles in a dict.

The hashing is now in once place. It is also lazy, so the hash is stored the first time it is used.


DrBenton (approved):
LGTM :slightly_smiling_face:

PR #2347, opened 2022-06-16
Add `echo` param to `Console.capture`

darrenburns:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Adds `echo` parameter to the `capture` method of `Console`. When `True`, captured output will also be written to the output file.

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -27,6 +27,7 @@ from typing import (
     Mapping,
     NamedTuple,
     Optional,
+    Set,
     TextIO,
     Tuple,
     Type,
@@ -317,14 +318,17 @@ class Capture:
 
     Args:
         console (Console): A console instance to capture output.
+        echo (bool): True if captured content should also be written to the
+            terminal. False otherwise.
     """
 
-    def __init__(self, console: "Console") -> None:
+    def __init__(self, console: "Console", echo: bool = False) -> None:
         self._console = console
+        self._echo = echo
         self._result: Optional[str] = None
 
     def __enter__(self) -> "Capture":
-        self._console.begin_capture()
+        self._console.begin_capture(self._echo)
         return self
 
     def __exit__(
@@ -723,6 +727,7 @@ class Console:
         self._thread_locals = ConsoleThreadLocals(
             theme_stack=ThemeStack(themes.DEFAULT if theme is None else theme)
         )
+        self._buffer_indices_to_echo: Set[int] = {self._thread_locals.buffer_index}
         self._record_buffer: List[Segment] = []
         self._render_hooks: List[RenderHook] = []
         self._live: Optional["Live"] = None
@@ -837,9 +842,16 @@ class Console:
         """Exit buffer context."""
         self._exit_buffer()
 
-    def begin_capture(self) -> None:
-        """Begin capturing console output. Call :meth:`end_capture` to exit capture mode and return output."""
+    def begin_capture(self, echo: bool = False) -> None:
+        """Begin capturing console output. Call :meth:`end_capture` to exit capture mode and return output.
+
+        Args:
+            echo (bool): True if captured content should also be written to the
+                terminal. False otherwise.
+        """
         self._enter_buffer()
+        if echo:
+            self._buffer_indices_to_echo.add(self._buffer_index)
 
     def end_capture(self) -> str:
         """End capture mode and return captured string.
@@ -849,6 +861,8 @@ class Console:
         """
         render_result = self._render_buffer(self._buffer)
         del self._buffer[:]
+        if self._buffer_index in self._buffer_indices_to_echo:
+            self._buffer_indices_to_echo.remove(self._buffer_index)
         self._exit_buffer()
         return render_result
 
@@ -1049,7 +1063,7 @@ class Console:
         """Play a 'bell' sound (if supported by the terminal)."""
         self.control(Control.bell())
 
-    def capture(self) -> Capture:
+    def capture(self, echo: bool = False) -> Capture:
         """A context manager to *capture* the result of print() or log() in a string,
         rather than writing it to the console.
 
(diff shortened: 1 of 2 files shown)


rich/console.py
```diff
@@ -1951,6 +1965,7 @@ def log(
 
     def _check_buffer(self) -> None:
         """Check if the buffer may be rendered. Render it if it can (e.g. Console.quiet is False)
+        be rendered, then clear it.
```
willmcgugan, line 1968: Docstring doesn't read too well. Can you rephrase?

willmcgugan (changes requested):
Can update the docs to describe the new `echo` param please.

rich/console.py
```diff
@@ -723,6 +727,7 @@ def __init__(
         self._thread_locals = ConsoleThreadLocals(
             theme_stack=ThemeStack(themes.DEFAULT if theme is None else theme)
         )
+        self._buffer_indices_to_render: Set[int] = {self._thread_locals.buffer_index}
```
willmcgugan, line 730: Don't quite follow what the intent of this is. Handling nested captures? Can you explain to me on Tuesday please.

willmcgugan (changes requested):

rich/console.py
darrenburns: Previously we only rendered content if the buffer index was 0. This was a hard-coded check. Now we also want to do this if we're inside a capture context (the buffer index is not 0) and where the `echo` arg is `True`, so we record the indices where that's the case.

rich/console.py
```diff
@@ -849,6 +861,8 @@ def end_capture(self) -> str:
         """
         render_result = self._render_buffer(self._buffer)
         del self._buffer[:]
+        if self._buffer_index in self._buffer_indices_to_render:
+            self._buffer_indices_to_render.remove(self._buffer_index)
```
willmcgugan, line 865:
```suggestion
        self._buffer_indices_to_render.discard(self._buffer_index)
```

rich/console.py
```diff
@@ -1049,7 +1063,7 @@ def bell(self) -> None:
         """Play a 'bell' sound (if supported by the terminal)."""
         self.control(Control.bell())
 
-    def capture(self) -> Capture:
+    def capture(self, echo: bool = False) -> Capture:
```
willmcgugan, line 1066: Docstring for new arg por favor

tests/test_console.py
```diff
@@ -368,6 +368,19 @@ def test_capture_and_record(capsys):
     assert out == "ABC\n"
 
 
+def test_capture_echo_outputs_captured_content_to_terminal(capsys):
```
willmcgugan, line 371: Could you add a test with more nesting. e..g. a capture inside a capture. Or a simple `with console:` block within a capture.

rich/console.py
willmcgugan, line 730: I think this will break with multiple threads.

You may need to add `_buffer_indices_to_render` to the thread locals.

willmcgugan (changes requested):
A few requests.

tests/test_console.py
darrenburns: Tried this test passes on `master` (it shouldn't), so I'm not sure nested captures work at all:

```python
def test_capture_echo_nested_capture():
    console = Console()
    console.print(1)
    with console.capture() as capture1:
        console.print(2)
        with console.capture() as capture2:
            console.print(3)
    console.print(4)

    assert capture1.get() == ""
    assert capture2.get() == ""
```

willmcgugan: I'm wondering what the expected behaviour should be.

The outer capture prevents content from getting to the terminal. So the inner capture has nothing to capture? So maybe this is the expected behaviour.

Could you also test this:

```python
with console.capture():
    with console:
        console.print("foo")
```

The with console should make the block atomic, i.e. it saves up all content until it exists the block.

nathanrpage97:
Just wanted to see if `echo` is the right term or would [`tee`](https://en.wikipedia.org/wiki/Tee_(command)) be better?

tests/test_console.py
darrenburns: > The outer capture prevents content from getting to the terminal. So the inner capture has nothing to capture? So maybe this is the expected behaviour.

I'd still expect the `2` to be captured inside `capture1`.

This test passes:

```
def test_capture_echo_nested_capture_console():
    console = Console()
    with console.capture() as capture1:
        console.print(1)
        with console:
            console.print(2)

    assert capture1.get() == "1\n2\n"
```

darrenburns: Ok, now that I've refreshed my memory with this stuff 😄 ...

This is the behaviour on `master`, which looks wrong:

```python
def test_capture_echo_nested_capture():
    console = Console()
    console.print(1)
    with console.capture() as capture1:
        console.print(2)
        with console.capture() as capture2:
            console.print(3)
    console.print(4)

    assert capture1.get() == ""
    assert capture2.get() == "2\n3\n"
```

My expectation would be that `capture1` captures `2\n3\n` and `capture2` captures `3\n`.

darrenburns:
@nathanrpage97 I agree, I don't think `echo` is the right term either 🤷‍♂️

nathanrpage97:
Yea, I guess it may be a bit esoteric for the general users of the library

ssbarnea:
@darrenburns Any chance to revive this? The changes introduced by 12.5.1 broke capture support in enrich, https://github.com/pycontribs/enrich/issues/40 and temporary I had to pin-down rich as I would need the echo/tee functionality. Thanks.

ssbarnea:
@willmcgugan That PR happens to be on xmas wishlist for quite some time. I wonder if Santa is feeling generous these days ;)

ssbarnea:
@darrenburns any chance to update it? It would be quite cool to see it in.

PR #2357, opened 2022-06-21
Prepend "async def" to Inspect coroutine functions

zephyrkul:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [X] I accept that @\willmcgugan may be pedantic in the code review.

## Description

Adds a feature to prepend `async def` when `rich.inspect` is used on a 3.5+ style coroutine function, as demonstrated in the below screenshot.
3.4-style generator-based coroutines are left as-is.

Terminal screenshot demonstrating new feature

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Environment variables `JUPYTER_COLUMNS` and `JUPYTER_LINES` to control width and height of console in Jupyter
+- `inspect` will prefix coroutine functions with `async def`
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


willmcgugan:
Good idea. Would I be correct in thinking this will do the same for methods?

zephyrkul:
It does, yes, stdlib's `inspect.iscoroutinefunction()` unwraps methods and partials for its check.
And just to double-check, I tested that as well:
<details><summary>Screenshot</summary>

A screenshot demonstrating the new rich.inspect for methods
</details>

willmcgugan:
And if you do `inspect(Foo, methods=True)` Does it show `async def` for each method?

willmcgugan:
Perfect. Thanks.

PR #2359, opened 2022-06-22
Update asv instructions and benchmarks

darrenburns:
* Removes the GitHub Action to copy ASV results over to `rich-benchmarks` repo. This is now manual, but it's required so rarely that the GitHub Action is probably overkill anyway.
* Pins setuptools version in ASV config - this was due to an issue I'd seen reported by a few projects and this was the suggested solution.
* Updates the benchmarks data to match what's currently on the site.
* Excludes benchmarks folder from pre-commit hooks since it contains generated data.
* Added instructions on how to update the benchmarks site.


DrBenton (approved):
I have a pretty approximate understanding of what's going there, as I haven't used that tool before, but I saw the shiny results which looks very good so I'm happy to approve that PR :smile:

willmcgugan (approved):

PR #2364, opened 2022-06-27
logging render_message: Use message as-is if already Text

jevinskie:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

If logging a message that is already rich Text, use it as is. This allows styling your log messages with rich which was required for my [rainbow hexdumps](https://twitter.com/jevinskie/status/1541193014785110018).

### DIFF
--- a/rich/logging.py
+++ b/rich/logging.py
@@ -173,7 +173,10 @@ class RichHandler(Handler):
             ConsoleRenderable: Renderable to display log message.
         """
         use_markup = getattr(record, "markup", self.markup)
-        message_text = Text.from_markup(message) if use_markup else Text(message)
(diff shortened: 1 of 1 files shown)


willmcgugan:
The hexdump is *very* cool, but I'm afraid this is an abuse of logging. Logging may be configured to go to multiple destinations. If you log a Text object it will break with anything other than RichHandler.

Consider using `console.log` which will give you similar results, and will already work with Text objects, but doesn't have the same limitations as logging

jevinskie:
Ack, thanks for the suggestion to use console.log instead!

PR #2365, opened 2022-06-28
added cached_cell_len

willmcgugan:
Makes `cached_cell_len` public.

This is primarily for Textual which will need to get the `cell_len` of generally small segments.


DrBenton (approved):
:+1: ! :-)

PR #2366, opened 2022-06-28
[jupyter] bypass the pretty-rendering of some specific graphics-related classes

DrBenton:
Matplotlib "Artists", for example, manage their own rendering in a Jupyter notebook, and we should not try to render them too.

## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [ ] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate. `(N/A)`

## Description

[STILL WIP]


rich/pretty.py
```diff
@@ -148,6 +148,18 @@ def _ipy_display_hook(
                 if repr_result is not None:
                     return  # Delegate rendering to IPython
 
+        # When in a Jupyter notebook let's avoid the display of some specific classes,
+        # as they result in the rendering of useless and noisy lines such as `<Figure size 432x288 with 1 Axes>`.
+        # What does this do?
+        # --> if the class has "matplotlib.artist.Artist" in its hierarchy for example, we don't render it.
+        classes_hierarchy_fully_qualified_names = {
+            # With this `[8:-2]` we convert  "<class 'pkg.Something'>" to  "pkg.Something":
+            str(class_)[8:-2]
```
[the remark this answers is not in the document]
willmcgugan: Sorry, that's not a reliable mechanism. It may well work most of the time, but there is no guarantee that will be the case for for all classes.

Consider using `__module__` which I *think* gets the fully qualified path to the module.

We also do not want to make any assumptions about the object being printed. There is no guarantee it is a class instance, and therefore has a `__class__`

rich/jupyter.py
```diff
@@ -14,6 +14,12 @@
 <pre style="white-space:pre;overflow-x:auto;line-height:normal;font-family:Menlo,'DejaVu Sans Mono',consolas,'Courier New',monospace">{code}</pre>
 """
 
+JUPYTER_CLASSES_TO_NOT_RENDER = {
```
willmcgugan, line 17: AFAICT this is only used in a single context. It doesn't need to be publicly exposed so we could move it to the file where it is referenced.

willmcgugan (changes requested):
Bear in mind the a user could display any object.

rich/pretty.py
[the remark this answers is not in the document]
DrBenton: Off the top of your head, would you have an example of a object that doesn't have a `__class__`, so I can test another way to manage this with various types of exotic objects?
(because if an object doesn't have such an attribute I'll likely struggle to get the part of its fully qualified name that comes after its module, so I would like to see what's possible for such objects :slightly_smiling_face: )

We would also have to find a way to get qualified names of objects that don't have a `__module__`: 
Screenshot from 2022-06-29 15-31-10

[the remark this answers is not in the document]
willmcgugan: Anything with `__getattribute__` defined. Which happens more than you might think. Usually from ORMs or other magic classes.

You're absolutely correct that `__module__` may not exist. You can't assume anything about objects, you will probably have to resort to `getattr`

DrBenton:
@willmcgugan Here is a new ~challenger~ implementation of the detection of the "_classes' FQDN denylist for Jupyter_" :slightly_smiling_face: [8b17bc6](https://github.com/Textualize/rich/pull/2366/commits/8b17bc61f826422b1fbd6aeb610631868ff47c35)

DrBenton:
I just did a quick check with a class that doesn't have the usual `__module__` / `__class__` attributes (which I cannot do in the tests since I haven't found any cryptic objects I could use for this usign Rich's dependencies and the Python stdlib), and it seems to work: :slightly_smiling_face: 
Screenshot from 2022-07-01 12-52-27

rich/_inspect.py
```diff
@@ -233,3 +233,29 @@ def _get_formatted_doc(self, object_: Any) -> Optional[str]:
...
+def object_types_tree_as_strings(obj: object) -> Tuple[str, ...]:
+    return tuple(
+        [
+            f'{getattr(type_, "__module__", "")}.{getattr(type_, "__qualname__", "")}'
+            for type_ in object_types_tree(obj)
+        ]
+    )
+
+
+def object_is_one_of_types(
```
DrBenton, line 255: note sure about that function's name - but I couldn't find better for the moment :sweat_smile:

rich/_inspect.py
```diff
@@ -233,3 +233,29 @@ def _get_formatted_doc(self, object_: Any) -> Optional[str]:
         if not self.help:
             docs = _first_paragraph(docs)
         return escape_control_codes(docs)
+
+
+def object_types_tree(obj: Union[object, Type[Any]]) -> Tuple[type, ...]:
```
willmcgugan, line 238: doc strings por favor

This doesn't seem to return a tree per se. Could it be `get_mro` ?

rich/_inspect.py
```diff
@@ -233,3 +233,29 @@ def _get_formatted_doc(self, object_: Any) -> Optional[str]:
...
+    obj_type: Type[Any] = type(obj)
+    if obj_type is type:
+        # the given object was already a type: let's use this instead:
+        obj_type = cast(Type[Any], obj)  # I wish MyPy could infer this itself :-/
+    return getattr(obj_type, "__mro__", ())
+
+
+def object_types_tree_as_strings(obj: object) -> Tuple[str, ...]:
```
willmcgugan, line 246: doc string por favor. Why convert to a tuple? Looks like it could be an iterable.

willmcgugan (changes requested):
Small requests

DrBenton:
@willmcgugan Done! :slightly_smiling_face:  [c9888a3](https://github.com/Textualize/rich/pull/2366/commits/c9888a320954d9425eb411b30125d8c081ce1ec9)

rich/_inspect.py
```diff
@@ -233,3 +233,38 @@ def _get_formatted_doc(self, object_: Any) -> Optional[str]:
         if not self.help:
             docs = _first_paragraph(docs)
         return escape_control_codes(docs)
+
+
+def object_types_mro(obj: Union[object, Type[Any]]) -> Tuple[type, ...]:
+    """Returns the MRO of an object's class, or of the object itself if it's a class."""
+    obj_type: Type[Any] = type(obj)
```
willmcgugan, line 240: Is this function strictly necessary? The call to `getattr` pretty much negates any benefit from typing. I'd be tempted to fold this function in to the function below.

rich/_inspect.py
```diff
@@ -233,3 +233,38 @@ def _get_formatted_doc(self, object_: Any) -> Optional[str]:
...
+    if obj_type is type:
+        # the given object was already a type: let's use this instead:
+        obj_type = cast(Type[Any], obj)  # I wish MyPy could infer this itself :-/
+    return getattr(obj_type, "__mro__", ())
+
+
+def object_types_mro_as_strings(obj: object) -> Collection[str]:
```
willmcgugan, line 247: `get_object_imports` ?

willmcgugan (approved):
Just some small suggestions

rich/_inspect.py
DrBenton: These are not imports: it's the list of an object's class (or the object itself it's a class) hierarchy in terms of OOP :slightly_smiling_face: 

So `get_object_types_mro_as_strings(TotalFileSizeColumn())` for example returns `['rich.progress.TotalFileSizeColumn', 'rich.progress.ProgressColumn', 'abc.ABC', 'builtins.object']`

Maybe it could be named `get_object_hierarchy_as_strings`, or something like this?

PR #2379, opened 2022-07-05
Fix using tuples as positional arguments in rich repr

saulshanabrook:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Currently, it isn't possible to use tuples as positional arguments from `__rich_repr__`, because they are special cased to support adding a key and a default value.

This adds support for them, by allowing `None` as a key.


willmcgugan:
This already works:

```python
class Foo:
    def __rich_repr__(self):
        yield None, (1,)

f=Foo()
from rich import print
print(f)
```

If its not documented, that should certainly be rectified.

willmcgugan (approved):

PR #2383, opened 2022-07-09
🐛 Fix type annotations for `@group` decorator

tiangolo:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other (type annotations)

## Checklist

## Description

The current type annotations for `rich.console.group` discard any type information in the original function.

This means that calling the function decorated with `@group()` doesn't get autocompletion nor inline errors for the arguments.

This is how it looks right now, notice that it should be showing an error:

Selection_167

And there's no autocompletion for the function parameters:

Selection_168

This change fixes the type annotations of the resulting function decorated with `@group()`.

Now it shows errors:

Selection_169

And it shows autocompletion:

Selection_170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix edges used in first row of tables when `show_header=False` https://github.com/Textualize/rich/pull/2330
 - Fix interaction between `Capture` contexts and `Console(record=True)` https://github.com/Textualize/rich/pull/2343
 - Fixed hash issue in Styles class https://github.com/Textualize/rich/pull/2346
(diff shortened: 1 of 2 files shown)


tiangolo:
Hehe, it seems there's a discrepancy in the types, not sure. I suspect the new types caught something not completely right. It probably needs an update in the types for `group()` or in the types for `traceback._render_syntax_error()`, but I'm not sure which one of the two is wrong...

willmcgugan:
Thanks. I think we might tackle this as part of a larger review of typing. https://github.com/Textualize/rich/issues/2846

nilsso:
Sorry to necro instead of opening a new issue, but why not the following?
```python
def group[**P](fit: bool = True) -> Callable[
    [Callable[P, Iterator[RenderableType]]],
    Callable[P, Group],
]: ...
```

PR #2394, opened 2022-07-14
Add Python 3.11.0-beta4 to CI

darrenburns:
* Tests Rich against Python 3.11.0-beta4
* There are issues with `pytest-cov` on 3.11 only. I'm going to minimally reproduce them and, if I can, report it to the pytest team. For now, coverage doesn't run on 3.11, but will continue to run on other versions.
* There were 5 test failures that have been resolved.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -35,10 +35,16 @@ jobs:
         run: |
           source $VENV
           make typecheck
-      - name: Test with pytest
+      - if: ${{ matrix.version }} != '3.11.0-beta.4'
+        name: Test with pytest (with coverage)
         run: |
           source $VENV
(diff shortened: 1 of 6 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -35,10 +35,16 @@ jobs:
...
+      - name: Test with pytest (with coverage)
+        if: matrix.python-version != '3.11.0-beta.4'
         run: |
           source $VENV
           pytest tests -v --cov=./rich --cov-report=xml:./coverage.xml --cov-report term-missing
+      - name: Test with pytest (no coverage)
+        if: matrix.python-version == '3.11.0-beta.4'
+        run: |
+          source $VENV
+          pytest tests -v
```
darrenburns, line 47: Using `--cov` causes a silent failure on 3.11, so we run without coverage on that version for now.

willmcgugan (approved):

Issue #2395, opened 2022-07-14
[BUG] Inconsistency across spinners


brechtm:
Some spinners, notably the ones built using emojis, include a trailing space. An extra space is inserted between the spinner and the spinner's text. I would argue that this is one space too much.

This difference is easy to see when running `python -m rich.spinner`.

willmcgugan:
Can you cut and paste it, and count the spaces?

brechtm:
Here are two from `python -m rich.spinner`:
```
 ◵ 'circleQuarters'
 🕘  'clock'
```

The frames in _spinners.py clearly include the spaces.

willmcgugan:
The spinners are inherited from another project, and I suspect they are there to compensate for differences in how emojis are rendered on different platforms.

willmcgugan:
I’d consider a PR but since you’re the first to notice, it’s very low priority.

PR #2396, opened 2022-07-14
Add optional 'spacer' argument to Spinner

brechtm:
Addresses #2395

## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description


brechtm:
@willmcgugan I've completed the missing things on the checklist, so this one is ready for review.

willmcgugan:
Not convinced this is necessary. If the spacing is not consistent, then perhaps that should be addressed. But I don't think it needs to be an option.

Closing because I'm clearing the backlog. Re-open if you want to discuss further.

brechtm:
We discussed this before; please see your comment in #2395:

> The spinners are inherited from another project, and I suspect they are there to compensate for differences in how emojis are rendered on different platforms.

willmcgugan:
If your spacer argument makes it look a little better on your platform, at the expense of it breaking on another, then I would't want to add it.

If you can show me that it doesn't visually break on Mac, Windows, and Linux, I'll reconsider...

brechtm:
As far as I can see, this is independent of platform.

The default value for spacer makes it so that the behavior is identical to what it was before.

willmcgugan:
And if you change it from the default you break another platform. Which will be the case if the spinners were crafted with the space in mind.

brechtm:
I don't understand how spaces act differently on the various platforms. In any case, the original issue is easy to work around.

PR #2397, opened 2022-07-16
Print cause of exceptions even when they have no traceback

nilsvu:
Exceptions from subprocesses raised by 'multiprocessing' don't include a traceback, so rich doesn't currently print them. They contain important information though, so this PR enables printing the exception without the traceback. This works because `walk_tb` doesn't do anything if the input is `None`.

Minimal example:

```py
import rich.traceback
import multiprocessing


def fail():
    raise ValueError


if __name__ == "__main__":
    rich.traceback.install()
    with multiprocessing.Pool() as pool:
        # The traceback printed by rich doesn't point to the function `fail`,
        # but only to code in the `multiprocessing` module. This PR fixes that.
        pool.apply(fail)
```

Note that exceptions from multiprocessing include the traceback formatted as a string in the description of the exception (see this code in multiprocessing.pool: https://github.com/python/cpython/blob/main/Lib/multiprocessing/pool.py#L57). This will be printed by rich after this PR.

## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -43,6 +43,7 @@ The following people have contributed to the development of Rich:
 - [Nicolas Simonds](https://github.com/0xDEC0DE)
 - [Aaron Stephens](https://github.com/aaronst)
 - [Gabriele N. Tornetta](https://github.com/p403n1x87)
+- [Nils Vu](https://github.com/nilsvu)
 - [Arian Mollik Wasi](https://github.com/wasi-master)
 - [Handhika Yanuar Pratama](https://github.com/theDreamer911)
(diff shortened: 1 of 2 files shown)


willmcgugan:
Can I see an example of how multiprocessing exceptions are printed?

willmcgugan:
Can you look at the CI fails please.

nilsvu:
Should be fixed, just needs your approval to run CI

willmcgugan (approved):

willmcgugan:
Thanks

Issue #2408, opened 2022-07-20
Provide an ability to mask/exclude certain keys when printing `show_locals=True`


datajoely:
Hello! We've just started our journey of adopting rich on [kedro](http://github.com/kedro-org/kedro), the first part of this journey introduced the rich logging handler + pretty tracebacks.

We've just had a user, quite rightly, highlight that `show_locals` has printed out some [credentials](https://github.com/kedro-org/kedro/issues/1712#issuecomment-1190071071) in plain text. The immediate fix for this is to set `show_locals=False`, which we will fix quickly. 

My question is as follows, would you accept a PR where we allow the user to exclude certain keys when printing out the stack trace?  I think it would go somewhere in this dictionary comprehension:

https://github.com/Textualize/rich/blob/aea574a215823a7f05e088399745cee00f959324/rich/traceback.py#L376

Thanks!

willmcgugan:
Agree in principal. Another (possibly complimentary) option might be to have a specially crafted local attribute with a set of keys to exclude from locals.

Something like:

```python
def get_password(self):
    _rich_exclude_locals = {"password"}
```

This would be more granular than a global black list.

If we were to add a blacklist, perhaps it should accept a list of regexes?

willmcgugan:
That would be possible with atomic data structures. For other objects you can't really know what attributes will be displayed in the `repr`. It would also be unwise to iterate over the attributes on an object as Rich may end up invoking arbitrary code via properties etc.

It would be possible for Rich to omit arbitrary `__repr__` calls to prevent user code from leaking credentials. You may miss out on some useful debugging information, but that might be a reasonable compromise.

danielgafni:
+1, this seems like a big security issue for production

willmcgugan:
I'd still accept a PR if anyone proposes an elegant solution, but frankly showing locals is a debug aid, and should always be disabled in production.

PR #2414, opened 2022-07-21
Provide `exclude_locals` mechanism to exclude certain regex patterns from Traceback output 

datajoely:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Provide `exclude_locals` mechanism to exclude certain regex patterns from Traceback output. Refers to #2408 .

### DIFF
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -1,13 +1,24 @@
 from __future__ import absolute_import
 
 import os
-import re
 import platform
+import re
 import sys
 from dataclasses import dataclass, field
 from traceback import walk_tb
 from types import ModuleType, TracebackType
-from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Type, Union, Tuple
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+)
 
 from pygments.lexers import guess_lexer_for_filename
 from pygments.token import Comment, Keyword, Name, Number, Operator, String
@@ -42,7 +53,7 @@ def install(
     theme: Optional[str] = None,
     word_wrap: bool = False,
     show_locals: bool = False,
-    exclude_locals: Optional[Tuple[str]],
+    exclude_locals: Optional[Tuple[str]] = None,
     indent_guides: bool = True,
     suppress: Iterable[Union[str, ModuleType]] = (),
     max_frames: int = 100,
@@ -60,7 +71,7 @@ def install(
             a theme appropriate for the platform.
         word_wrap (bool, optional): Enable word wrapping of long lines. Defaults to False.
         show_locals (bool, optional): Enable display of local variables. Defaults to False.
-        exclude_locals (Optional[Tuple[str]], optional): List patterns to exclude from local variables output. 
+        exclude_locals (Optional[Tuple[str]], optional): List patterns to exclude from local variables output.
             Defaults to None.
         indent_guides (bool, optional): Enable indent guides in code and locals. Defaults to True.
         suppress (Sequence[Union[str, ModuleType]]): Optional sequence of modules or paths to exclude from traceback.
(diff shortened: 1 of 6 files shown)


rich/traceback.py
```diff
@@ -362,27 +395,35 @@ def safe_str(_object: Any) -> str:
...
             for frame_summary, line_no in walk_tb(traceback):
                 filename = frame_summary.f_code.co_filename
                 if filename and not filename.startswith("<"):
                     if not os.path.isabs(filename):
                         filename = os.path.join(_IMPORT_CWD, filename)
                 if frame_summary.f_locals.get("_rich_traceback_omit", False):
                     continue
+
+                if show_locals:
+                    local_vars = {
```
willmcgugan, line 409: AFAICT This will entirely omit matching regexes, which might leave the reader wondering why some values are missing. It would be nicer if the value is replaced by something to indicate it was removed. So the user would see something along these lines:

```
     foo = "bar"
password = <redacted>
```

The best place to do this is probably in the `Pretty` class. This would also allow you to filter recursively so that a "password" key in a dict would also be redacted.

P.S. We have a policy of no abbreviations, in particular single letters.

rich/traceback.py
```diff
@@ -41,6 +53,7 @@ def install(
     theme: Optional[str] = None,
     word_wrap: bool = False,
     show_locals: bool = False,
+    exclude_locals: Optional[Tuple[str]] = None,
```
willmcgugan, line 56: The typing could probably be relaxed a little. I usually go for `Iterable` so the caller could pass in any container with strings.

willmcgugan (changes requested):

rich/traceback.py
datajoely: Gotcha! I'll have another go!

datajoely: So I didn't do the recursive implementation, I've taken quite a conservative approach. The output you get looks something like this:

```
    ╭────────────────────── locals ──────────────────────╮
    │         password = <redacted variable name>        │
    │          my_list = [1, 2, 3]                       │
    │ credentials_dict = <redacted variable content>     │
    ╰────────────────────────────────────────────────────╯
```

What are your thoughts?

rich/traceback.py
```diff
@@ -343,6 +393,27 @@ def safe_str(_object: Any) -> str:
...
 
+        def mask_local_variable(
+            key: Any, value: Any, patterns: Optional[Iterable[str]]
+        ) -> Any:
+            """Mask locals output if key or value match regular expressions passed into `mask_locals`."""
+
+            exclude_patterns = patterns if patterns else ()
+
+            excluded_variable_name = any(
+                re.search(x, key.__repr__()) for x in exclude_patterns
```
willmcgugan, line 404: `__repr__` can fail. It can throw exceptions or return non-strings. When dealing with exceptions Rich as to be very defensive. See the `safe_repr` method.

rich/traceback.py
```diff
@@ -369,20 +440,21 @@ def safe_str(_object: Any) -> str:
                         filename = os.path.join(_IMPORT_CWD, filename)
                 if frame_summary.f_locals.get("_rich_traceback_omit", False):
                     continue
+
+                if show_locals:
+                    local_variables = {
+                        k: mask_local_variable(key=k, value=v, patterns=mask_locals)
```
willmcgugan, line 446: Excluding `pretty.traverse` will break the advanced formatting of data structures, and a number of other Rich features. I think the solution has to involve `pretty.transverse`.

willmcgugan (changes requested):
I think the redacted locals has to be recursive, which means it should be a feature of the `rich.retty` module.

This is tricker code than it might appear because Rich has to be extremely defensive. We can't risk the exception output generating an error of its own.

I think this is a worthwhile feature, so I'd probably put it on the todo list regardless. But you are welcome to persevere!

peterromfeldhk:
:eyes:

datajoely:
I do plan to pick this up again later in the month, but if anyone wants to take it over please go ahead

rich/traceback.py
peterromfeldhk: not sure how someone would take over an open PR
```
diff --git a/rich/traceback.py b/rich/traceback.py
index 3ff09c15..f84c45e4 100644
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -4,6 +4,7 @@ import os
 import platform
 import re
 import sys
+from pprint import saferepr
 from dataclasses import dataclass, field
 from traceback import walk_tb
 from types import ModuleType, TracebackType
@@ -401,10 +402,10 @@ class Traceback:
             exclude_patterns = patterns if patterns else ()
 
             excluded_variable_name = any(
-                re.search(x, key.__repr__()) for x in exclude_patterns
+                re.search(x, saferepr(key)) for x in exclude_patterns
             )
             excluded_variable_value = any(
-                re.search(x, value.__repr__()) for x in exclude_patterns
+                re.search(x, saferepr(value)) for x in exclude_patterns
             )
 
             if excluded_variable_name or excluded_variable_value:
```

willmcgugan:
Closing, assumed stale.

PR #2439, opened 2022-08-02
New markdown parser

darrenburns:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

Work in progress.

Switches over to `markdown-it-py` instead of `commonmark` for Markdown parsing.

new_markdown


rich/markdown.py
```diff
@@ -405,16 +413,24 @@ class Markdown(JupyterMixin):
...
+        "list_item_open": ListItem,
         "image": ImageItem,
     }
-    inlines = {"emph", "strong", "code", "strike"}
+
+    # Maps tag names to Rich style keys, if tag and key differ
+    tag_to_style_name = {
+        "em": "emph",
+    }
+
```
darrenburns, line 432: I could probably just add a `markdown.em` style and have it match `markdown.emph`, then get rid of this. When I made it I thought there would be more mismatches 🤷

rich/markdown.py
```diff
@@ -449,83 +478,109 @@ def __rich_console__(
...
+                    # TODO: I want to encapsulate the stuff below into the Link renderable
+                    #  but can't work out how to do it.
+                    context.enter_style(link_style)
+                    context.on_text(element.text.plain, node_type)
+                    context.leave_style()
+                    context.on_text(" (", node_type)
+                    context.enter_style(link_style)
+                    context.on_text(element.href, node_type)
+                    context.leave_style()
+                    context.on_text(")", node_type)
```
darrenburns, line 523: This was the part I'm very "meh" about. It's basically a hack. As the TODO comment suggests, I wanted to throw the data into the `Link` renderable when the tag opens, and have it deal with this stuff. However, I can't get the information to render in the right place. Right now, we create a `Link`, and put it on the stack. Then we pop it off here and "manually" render the data - it's misusing the stack really.

rich/markdown.py
```diff
@@ -302,23 +299,33 @@ def render_number(
...
         Args:
             markdown (Markdown): The parent Markdown object.
-            node (Any): A node from Pygments.
+            token (Any): A token from markdown-it.
 
         Returns:
             MarkdownElement: A new markdown element
         """
-        return cls(node.destination, markdown.hyperlinks)
+        return cls(token.attrs.get("src"), markdown.hyperlinks)
```
willmcgugan, line 328: I'm a bit confused here. Looks like this may potentially call `cls` with `None`, which doesn't look allowed from the `__init__` below. Surprised typing didn't pick that up.

rich/markdown.py
```diff
@@ -449,83 +470,107 @@ def __rich_console__(
...
-                if current.is_container():
-                    if entering:
-                        context.enter_style(f"markdown.{node_type}")
-                    else:
-                        context.leave_style()
                 else:
-                    context.enter_style(f"markdown.{node_type}")
-                    if current.literal:
-                        context.on_text(current.literal, node_type)
+                    element: Any = context.stack.pop()
```
willmcgugan, line 505: Can we not type this as the base class?

rich/markdown.py
willmcgugan: I think I understand what's going on here. You're inserting content that isn't technically in the document, when hyperlinks are disabled?

I doesn't look all *that* dirty to me. But I think you're right that is should probably be the Link object's responsibility.

We can revisit that tomorrow, and try to figure out a cleaner solution. But if this is what we go for, it's not going to keep me up at night ;)

willmcgugan (approved):
Nice work. It's quite a small footprint for such a significant change. Love that the public interface doesn't change.

Approving, but let's not merge just yet. Would like to do one last Python3.6 release.

rich/markdown.py
darrenburns: It doesn't work because access attributes below which don't exist on the base class, only on `Link` (a subclass of the base MarkdownElement). This is most of the reason why I'd like to encapsulate all the stuff below inside `Link.__rich_console__`, it's cleaner _and_ would resolve this typing issue as we wouldn't have those attribute accesses :)

rich/text.py
```diff
@@ -903,7 +903,7 @@ def append(
         """
 
         if not isinstance(text, (str, Text)):
-            raise TypeError("Only str or Text can be appended to Text")
+            raise TypeError(f"Only str or Text can be appended to Text")
```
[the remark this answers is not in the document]
darrenburns: Think this is a comment on an outdated commit, I fixed this this morning here: https://github.com/Textualize/rich/pull/2439/commits/15c41e7d7acc6a9c5546a56d73fe0e9971aefae1

rich/markdown.py
darrenburns: I ended up making this change

rich/markdown.py
darrenburns: It's because the signatures were typed as `Any`, and I never changed them.  I've updated them to ensure we get more help from `mypy` (and fixed all the corresponding issues, including the one you mentioned).

rich/markdown.py
darrenburns: Good point, forgot about that - have added that in

synrg:
I realize this might be slightly out of scope of this PR, as it does not enable strikethrough extension, but I thought I'd mention it here unless it indicates an issue that is better to address now.

I found that merely enabling the extension was not enough to make `~~strike~~` work. That resulted in an empty string when rendered. What I found in my debugging is that strikethrough markup is being tokenized as tag `s`, not `strike`. In order to make it work, I had to both enable the strikethrough extension and modify the tag to process, as per:

```diff
diff --git a/rich/markdown.py b/rich/markdown.py
index 36d6491e..3b75c579 100644
--- a/rich/markdown.py
+++ b/rich/markdown.py
@@ -427,7 +427,7 @@ class Markdown(JupyterMixin):
         "image": ImageItem,
     }

-    inlines = {"em", "strong", "code", "strike"}
+    inlines = {"em", "strong", "code", "s"}

     def __init__(
         self,
@@ -439,7 +439,7 @@ class Markdown(JupyterMixin):
         inline_code_lexer: Optional[str] = None,
         inline_code_theme: Optional[str] = None,
     ) -> None:
-        parser = MarkdownIt().enable("table")
+        parser = MarkdownIt().enable(["table", "strikethrough"])
         self.markup = markup
         self.parsed = parser.parse(markup)
         self.code_theme = code_theme
```

Furthermore, since the default theme has `"markdown.strike": Style(strike=True)`, I had to use a custom theme using the `s` tag:

```py
from rich.theme import Theme
from rich.style import Style
our_theme = Theme({
    "markdown.s": Style(strike=True),
})
console = Console(theme=our_theme)
```

daniloegea:
Hi all,

considering commonmark is not being maintained anymore, are there plans to merge this PR in the near future?

darrenburns:
@synrg That's great, thanks! I've added that to this PR so that we support strikethrough (cc @willmcgugan).

@daniloegea Just polishing this up now, so should be getting merged relatively soon.

.github/workflows/pythonpackage.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
darrenburns, line 11: We were missing 3.11.

willmcgugan (approved):

willmcgugan:
Nice. This will make some folks happy.

daniloegea:
That's brilliant. Thanks so much! :)

synrg:
superb! many thanks

slyon:
Awesome, thank you! Will there be a new release including this change, soon?

chrisjsewell:
FYI as the maintainer of markdown-it-py, I approve 😄 was wondering why there was a sudden spike in pypi downloads 😆 

Love rich, keep up the good work!

PR #2446, opened 2022-08-04
Add crosshairs feature

rodrigogiraoserrao:
## Type of changes

- [x] New feature

## Checklist

## Description

This introduces a `Crosshairs` renderable that draws a cross in the terminal composed of box characters.

E.g.,

```py
from rich import print
from rich.crosshairs import Crosshairs

print(Crosshairs(1, 3))
```

would produce output similar to

```txt
 │
 │
 │
─┼──────────...
 │
...
```


rodrigogiraoserrao:
I have 2 questions regarding this:

 - the test cases I wrote include some edge cases that were written out in a big string that is then split. I think this is a very reasonable way to test the 9 edge cases for the `Crosshairs` because the output is meant to be very visual; does anyone oppose to having the test cases generated like this? (See [this string](https://github.com/rodrigogiraoserrao/rich/blob/4c77c497608d88993957c0397a764e11cc0ca1cb/tests/test_crosshairs.py#L8) and [the slight parsing](https://github.com/rodrigogiraoserrao/rich/blob/4c77c497608d88993957c0397a764e11cc0ca1cb/tests/test_crosshairs.py#L62-L63).)

 - for the lines that only contain the vertical bar `│`, should they contain trailing whitespace to match the length of the horizontal bar of the cross, or do those lines end at `│`?

willmcgugan:
That seems a very good way of expressing the test expectations. It's nice to be able to visually compare.

I think, yes the crosshairs should have trailing whitespace.

BTW, sorry to add this late in the day, but the crosshair class should take a `StyleType` so you can tweak the colors of the background / foreground.

rodrigogiraoserrao:
> BTW, sorry to add this late in the day, but the crosshair class should take a `StyleType` so you can tweak the colors of the background / foreground.

Alright, on it.

rodrigogiraoserrao:
At this point, I am just uncomfortable with something: my `__rich_console__` does not use the arguments `options`, and I realise that _might_ be OK, but I do not have enough experience with the console protocol to know for sure.
I only need to have access to the console size and to build the style, but I can do that with the argument `console`.

Finally, I have to write a test that makes sure `Crosshairs` instances can be styled. I saw some style tests in other files that encode the result with the explicit ANSI codes that are expected. Is this the only way to go about doing this?

willmcgugan:
It's fine not to use `options`. Not all renderables require it.

Re testing the ANSI codes. You could test at a slightly higher level. Perhaps test the Segment objects have the expected Style. Converting Styles to ANSI codes has been so thoroughly tested there probably isn't any benefit in test that again.

rodrigogiraoserrao:
> Re testing the ANSI codes. You could test at a slightly higher level. Perhaps test the Segment objects have the expected Style. [...]

Meanwhile, I already wrote a test with the explicit ANSI codes, which I'll assume was too much trouble, but also fine now that it is done. Correct me if I'm wrong.

Other than that, this PR is ready for review.

willmcgugan:
Thanks! Will do a review shortly.

rich/crosshairs.py
```diff
@@ -0,0 +1,88 @@
...
+class Crosshairs(JupyterMixin):
+    """Renders a crosshairs that spans the whole console window.
+
+    Args:
+        x (int): The x position of the cross centre.
+        y (int): The y position of the cross centre.
+        style (StyleType, optional): Style to render the crosshairs with. Defaults to "".
+    """
+
+    def __init__(self, x: int, y: int, style: "StyleType" = ""):
```
willmcgugan, line 22: Nitpick, I like to add `-> None:` on constructors for consistency (even though its kind of redundant).

willmcgugan (approved):
LGTM. Nice efficient solution!

rich/crosshairs.py
rodrigogiraoserrao: My bad, fixed.

rodrigogiraoserrao:
The build error looks like it is the same build error as in #2450, again with Py 3.11...

Also, not sure how I managed to _decrease_ the coverage of the file `rich/cells.py`. I didn't add code to that file and I didn't change any tests that already existed.

PR #2450, opened 2022-08-06
Fix CHANGELOG.md

rodrigogiraoserrao:
## Type of changes

- [x] Documentation

## Checklist

- [x] I accept that @willmcgugan may be pedantic in the ~~code~~ review.

## Description

The changelog had missing/wrong compare links for most recent and oldest versions.
I see some compare links are present even if the underlying tags were not created, so I followed this style and created all links as if the tags existed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1807,9 +1807,13 @@ Major version bump for a breaking change to `Text.stylize signature`, which corr
 
 - First official release, API still to be stabilized
 
-[unreleased]: https://github.com/willmcgugan/rich/compare/v12.4.3...HEAD
-[12.4.1]: https://github.com/willmcgugan/rich/compare/v12.4.2...v12.4.3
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao:
Checking why the MacOS on Py3.11 failed, and it doesn't look like it's this PR's fault, but maybe I misunderstood the error.

willmcgugan:
Yeah, looks environment related. Will try a rerun.

rodrigogiraoserrao:
The Mac build passed after your rerun, so it does look like it was env related.
Who's at fault, here..?

willmcgugan:
Beats me. We do have some integration tests which may suffer from race conditions. One of the pains of concurrent programming.

Unrelated to this PR though!

willmcgugan:
obrigado

rodrigogiraoserrao:
De nada.
I think you need to approve before this gets auto-merged 🙃

willmcgugan (approved):

PR #2489, opened 2022-08-25
Add delimiters to elided pretty prints

### LINKED ISSUE #2485: [REQUEST] Allow delimiters when using `pprint(..., max_depth=...)`

Currently, when pretty printing a nested object, we can set the max depth to simplify the object display:
```python
>>> from rich.pretty import pprint
>>> pprint({"a": {"b": {"c": "d"}}, "e": ["f", "g"]}, max_depth=1)
{'a': ..., 'e': ...}
```

However, by using ellipsis like this, we obscure the information about the data type of "a". In contrast, Python's built-in `pprint` module adds delimiters around the ellipsis, to differentiate between a list and a dict:
```python
>>> import pprint
>>> pprint.pprint({"a": {"b": {"c": "d"}}, "e": ["f", "g"]}, depth=1)
{'a': {...}, 'e': [...]}
```

Notice how, with only 4 extra characters, we have significantly more information about the eluded structure.

I would like to propose that we add a new argument to `pprint()` such as `max_depth_delims: bool`, which when True, would enable the behaviour above.

When I use `max_depth` in tutorial or example, it's because I don't want unnecessary detail taking up a large amount of screen real estate. However I still want my users to understand that there *is* additional detail available, and to give them a hint as to what kind of data is nested further in the tree, and how to access it. Adding these delimiters nicely enable this compromise.

multimeric:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Closes #2485

One thing not discussed in the issue was how to deal with more complex types like dataclasses and namedtuples. So far I'm treating them the same as dicts and lists. In other words, the delimiters stay the same as if we weren't abridging anything, but the children are replaced with `...`. For example: `"{'unit': StockKeepingUnit(...)}"`

### DIFF
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -676,7 +676,10 @@ def traverse(
                 append = children.append
 
                 if reached_max_depth:
-                    node = Node(value_repr=f"...")
+                    if angular:
+                        node = Node(value_repr=f"<{class_name}...>")
+                    else:
+                        node = Node(value_repr=f"{class_name}(...)")
                 else:
                     if angular:
                         node = Node(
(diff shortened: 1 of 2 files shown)


willmcgugan (approved):

willmcgugan (approved):

willmcgugan:
THanks

Issue #1810, opened 2022-01-06
[REQUEST] Ability to suppress the length of local variables or completely filter them in Rich tracebacks.


pbsds:
**Have you checked the issues for a similar suggestions?**
#318 and #1378 is related

**How would you improve Rich?**

Showing the local variables in Rich tracebacks is incredibly helpful, but this does at times become too verbose prompting me to disable it. Especially when dealing with types such as `torch.nn.Module`, which feature _very_ long `__repr__` and `__str__` outputs (example of a single frame featured below). My feature request is twofold: the ability to _filter_ certain local variables entierly, and the ability to _suppress_/_truncate_ the length of others.

Variable names that start with a double underscore prefix (e.g. `__init__`) is a prime target for filtering.
I imagine parametrizing such behavior in `rich.traceback.install` with an optional lambda function (e.g. `locals_filterer=lambda x: not x.startswith("__")`).

For suppressing/truncating the length, one could either fallback to the default `object.__repr__` format, or simply clip all `__repr__`() output to `n` lines with trailing ellipsis (as dictionaries already seem to do).
For deciding which locals to suppress, if not all, one could either provide a list of qualified class names (`type(obj).__qualname__`) and types (kinda like the existing `suppress` argument for modules), or provide a lambda which determines if it should be supressed (e.g. `locals_suppress=lambda x: any(isinstance(x, i) for i in list_of_types_to_suppress)`)

<details>
<summary>Example output (note the `module = MyMachineLearningModel`):</summary>

```
│ /home/pbsds/ntnu/xxxxxxxxxxxxxxxxxxxxxxxxxx/yyyyyy/cli.py:332 in log_training_setup                                          │
│                                                                                                                              │
│   329 │   │   if logger is not None:                                                                                         │
│   330 │   │   │   print(f"{logger.__class__.__qualname__} hparams:")                                                         │
│   331 │   │   │   print_dict_twocolumn(config.logging)                                                                       │
│ ❱ 332 │   │   │   logging.log_config(logger, logger = {"_class": logger.__class__.__name__} |                                │
│   333 │   │   │                                                                                                              │
│   334 │   │   │   # host info                                                                                                │
│   335 │   │   │   def cmd(cmd: Union[str, list[str]]) -> str:                                                                │
│                                                                                                                              │
│ ╭───────────────────────────────────────────────────────── locals ─────────────────────────────────────────────────────────╮ │
[...]

PR #2500, opened 2022-08-31
Add option to suppress provided types in tracebacks

cmckain:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

As described in #1810, this will allow for certain types to be excluded content-wise from Rich tracebacks, including if they are contained within certain hardcoded (list, tuple, and dict) arrays, where they will be replaced by their type (or array[type]). This argument may be used whenever tracebacks are initialized, including the general install() handler. It takes an array, instead of a callable as originally suggested[^1], because I think the problem this is intended to fix is more around the content of a variable, rather than the name of a variable (#2414 is probably a better mechanism for implementing that).

Although @willmcgugan had suggested a similar argument be included for limiting the length of local variables, I think that you would have to implement additional checks for the content of nested arrays (that's why the original locals_max_string and locals_max_length arguments didn't work for my purposes) and that should be for an additional PR.

[^1]: It could probably be rewritten to accept both arrays and functions if that would make this more useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Parse ANSI escape sequences in pretty repr https://github.com/Textualize/rich/pull/2470
 - Add support for `FORCE_COLOR` env var https://github.com/Textualize/rich/pull/2449
 - Allow a `max_depth` argument to be passed to the `install()` hook https://github.com/Textualize/rich/issues/2486
+- Add a 'locals_suppress' argument to the Traceback handler https://github.com/Textualize/rich/issues/1810
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


willmcgugan:
Name (suppress_locals) and type (i.e. Any) isn't great. Also not convinced of the utility of omitting particular types.

Don't think this is the solution. Willing to be convinved.

cmckain:
The name was just the first thing to come to mind (and I agree suppress really isn't the right word in hindsight) and the type has to be Any to identify it as accepting any type (although, if there is a type in Typing that specifically counts as any type, that would work too).

As for the utility, I completely understand that my use case isn't applicable to everyone but I implemented it like this to deal with BS4's PageElements (in arrays and individuallly) specifically where they would add a couple of zeroes to the size of my trackbacks in my logging setup because Rich would just print an entire HTML page/element and where monkey patching every BS4 function that returns one to include a proper __rich_rept___ wouldn't be feasible.

Regardless, I appreciate you taking the time to read this PR.

callzhang:
I think it is exactly what I need. I am using rich in deep learning and I need to suppress locals of `torch.nn` type, otherwise the model weights will be printed for thousands of pages.

Bit0r:
@willmcgugan May I ask how to install your package? I think I need this feature.

PR #2513, opened 2022-09-07
Supporting no stdout, fix force color bug

### LINKED ISSUE #2400: [BUG] `AttributeError` using `Console.capture` when run with `pythonw`
**Describe the bug**

This code, when run using `pythonw`:

```py
table = Table(box=DOUBLE_EDGE)
printer = Console(force_terminal=False)
with printer.capture() as capture:
    printer.print(table)
```

fails with this stack trace:

```text
  File "<MY CODE>", line ...
    with printer.capture() as capture:
  File "rich\console.py", line 336, in __exit__
    self._result = self._console.end_capture()
  File "rich\console.py", line 852, in end_capture
    self._exit_buffer()
  File "rich\console.py", line 796, in _exit_buffer
    self._check_buffer()
  File "rich\console.py", line 1979, in _check_buffer
    self.file.fileno() in _STD_STREAMS_OUTPUT
AttributeError: 'NoneType' object has no attribute 'fileno'
```

My guess is that `pythonw` causes `Console` to get some attributes that are usually not null, but in this case, are.

I can probably immediately work around this by instantiating the `Console` like so:

```py
table = Table(box=DOUBLE_EDGE)
capture = StringIO()
printer = Console(file=capture)
printer.print(table)
```

but Rich could probably cater for the `Console` instance having no `STDOUT` in some cases (such as above).

**Platform**

* Windows 10
* Python v3.10.0

I've avoided including details generated by `python -m rich.diagnose` because more diagnostic information would be provided using `pythonw ...`—but of course, since it disables `STDOUT`, there's no obvious way to grab that info for you. 🤣

darrenburns:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Fixes #2400.

Ensures RIch can run when stdout is not available (for example from pythonw).

I've tested on a Windows machine and ensured that you can use a console (and capture from it) as expected when using python.

Also fixes the issue with FORCE_COLOR.

image

image


willmcgugan (changes requested):
Might want to add a test for the correct behaviour when FORCE_COLOR is empty string

willmcgugan (approved):

PR #2520, opened 2022-09-12
Added console named argument to rich.reconfigure.

AndreasBackx:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

The documentation for `rich.reconfigure` says there is an option for a `console` argument. That however was never implemented and I cannot find any changes around that piece of the code so assume it was forgotten and not removed? 🤔 Instead of making an issue about it, I thought the fastest thing was to add it. I'm open to any changes that you would propose.

I am interested in passing my own `console` so I can pass a custom subclass. The reason I would want this, is so I can change the `LogRender` being used by the `Console` which would require a custom subclass afaik. https://github.com/Textualize/rich/discussions/344 is relevant, though it only does it for the `logging` handler, this would allow a custom `LogRender` for other messages as well.

I was thinking of making it a `Type[Console]`, though that would require the documentation to change as well. Personally I am not very fussed on how this is implemented so let me know what you think/prefer. :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Handle stdout/stderr being null https://github.com/Textualize/rich/pull/2513
 - Fix NO_COLOR support on legacy Windows https://github.com/Textualize/rich/pull/2458
+- `rich.reconfigure` accepts `console` argument as was in the documentation but was not implemented
 
 ## [12.5.2] - 2022-07-18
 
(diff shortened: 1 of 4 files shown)


willmcgugan:
The `console` arg was a mistake and was removed.

I don't think you'll need it. You might want to open a discussion with your log render issues. There are probably better ways of solving it.

AndreasBackx:
@willmcgugan thanks for getting back! I've created https://github.com/Textualize/rich/discussions/2534, let me know what we could do to move this forward. 😊

PR #2537, opened 2022-09-21
Add support for specifying (persistent) id for SVG exports

ashb:
As part of Airflow's CI we save SVG images of our command's help output and commit them to git.

Without this change a single character change in the help text results in every single class and id in the SVG changing, which makes the diffs unreadable.

This adds an option for us to be able to specify a fixed prefix so that our git diffs are minimized.

(I also tided up the help docstring for export_svg and save_svg which didn't match signature any more)

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -2262,12 +2262,12 @@ class Console:
         clear: bool = True,
         code_format: str = CONSOLE_SVG_FORMAT,
         font_aspect_ratio: float = 0.61,
+        unique_id: Optional[str] = None,
     ) -> str:
         """
         Generate an SVG from the console contents (requires record=True in Console constructor).
 
         Args:
-            path (str): The path to write the SVG to.
             title (str, optional): The title of the tab in the output image
             theme (TerminalTheme, optional): The ``TerminalTheme`` object to use to style the terminal
             clear (bool, optional): Clear record buffer after exporting. Defaults to ``True``
@@ -2277,6 +2277,8 @@ class Console:
             font_aspect_ratio (float, optional): The width to height ratio of the font used in the ``code_format``
                 string. Defaults to 0.61, which is the width to height ratio of Fira Code (the default font).
                 If you aren't specifying a different font inside ``code_format``, you probably don't need this.
+            unique_id (str, optional): unique id that is used as the prefix for various elements (CSS styles, node
+                ids). If not set, this defaults to a computed value based on the recorded content.
         """
 
         from rich.cells import cell_len
@@ -2372,14 +2374,16 @@ class Console:
             if clear:
                 self._record_buffer.clear()
 
-        unique_id = "terminal-" + str(
-            zlib.adler32(
-                ("".join(repr(segment) for segment in segments)).encode(
-                    "utf-8", "ignore"
+        if unique_id is None:
+            unique_id = "terminal-" + str(
+                zlib.adler32(
+                    ("".join(repr(segment) for segment in segments)).encode(
+                        "utf-8",
+                        "ignore",
+                    )
+                    + title.encode("utf-8", "ignore")
                 )
-                + title.encode("utf-8", "ignore")
             )
-        )
         y = 0
         for y, line in enumerate(Segment.split_and_crop_lines(segments, length=width)):
             x = 0
@@ -2510,6 +2514,7 @@ class Console:
         clear: bool = True,
         code_format: str = CONSOLE_SVG_FORMAT,
         font_aspect_ratio: float = 0.61,
+        unique_id: Optional[str] = None,
     ) -> None:
         """Generate an SVG file from the console contents (requires record=True in Console constructor).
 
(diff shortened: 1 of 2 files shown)


ashb (commented):
As part of Airflow's CI we save SVG images of our command's help output and commit them to git.

Without this change a single character change in the help text results in every single class and id in the SVG changing, which makes the diffs unreadable.

This adds an option for us to be able to specify a fixed prefix so that our git diffs are minimized.

(I also tided up the help docstring for `export_svg` and `save_svg` which didn't match signature any more)

potiuk (approved):
Nice one. It will be helpful for us :)

willmcgugan:
Are you aware of the reasoning behind the unique id? It is to allow embedding of SVGs without conflicting styles. If you make the unique id non unique you will not be able to embed more than one SVG.

ashb:
Yes, we can give each image it's own unique (but stable name) based on the sub command name (we use rich-click). I can expand on the docs to make point to others?

rich/console.py
```diff
@@ -2261,18 +2261,20 @@ def export_svg(
...
     ) -> str:
         """
         Generate an SVG from the console contents (requires record=True in Console constructor).
 
         Args:
-            path (str): The path to write the SVG to.
             title (str): The title of the tab in the output image
             theme (TerminalTheme, optional): The ``TerminalTheme`` object to use to style the terminal
-            clear (bool, optional): Clear record buffer after exporting. Defaults to ``True``
+            clear (bool): Clear record buffer after exporting. Defaults to ``True``
```
willmcgugan, line 2272: `clear` is optional...

rich/console.py
```diff
@@ -2505,23 +2508,27 @@ def save_svg(
...
+        unique_id: Optional[str] = None,
     ) -> None:
         """Generate an SVG file from the console contents (requires record=True in Console constructor).
 
         Args:
             path (str): The path to write the SVG to.
             title (str): The title of the tab in the output image
             theme (TerminalTheme, optional): The ``TerminalTheme`` object to use to style the terminal
-            clear (bool, optional): Clear record buffer after exporting. Defaults to ``True``
+            clear (bool): Clear record buffer after exporting. Defaults to ``True``
```
willmcgugan, line 2519: clear is optional.

rich/console.py
```diff
@@ -2368,14 +2370,15 @@ def stringify(value: object) -> str:
             if clear:
                 self._record_buffer.clear()
 
-        unique_id = "terminal-" + str(
-            zlib.adler32(
-                ("".join(segment.text for segment in segments)).encode(
-                    "utf-8", "ignore"
+        if not unique_id:
```
willmcgugan, line 2373: Suggest an explicit check for None here.

willmcgugan (changes requested):

rich/console.py
ashb, line 2524:
```suggestion
            clear (bool, optional): Clear record buffer after exporting. Defaults to ``True``
```

rich/console.py
ashb, line 2273:
```suggestion
            clear (bool, optional): Clear record buffer after exporting. Defaults to ``True``
```

rich/console.py
ashb, line 2377:
```suggestion
        if unique_id is not None:
```

ashb:
Whoops, that'll teach me for making changes in PR suggestion. Fixing...

ashb:
Fixing tests now. One moment please

willmcgugan:
Thanks, probably a result of the merge.

ashb:
I added tests too for good measure. (used `EXPECTED_SVG.replace()` instead of duplicating the otherwise identical string)

willmcgugan:
Thanks

rich/console.py
```diff
@@ -2262,12 +2262,12 @@ def export_svg(
         clear: bool = True,
         code_format: str = CONSOLE_SVG_FORMAT,
         font_aspect_ratio: float = 0.61,
+        unique_id: Optional[str] = None,
```
jayaddison, line 2265: Hi - apologies in advance: maybe some unnecessarily verbose commentary.

I'm wondering whether an (replacement?  additional?) argument, `isolate: bool = True` could make sense here.

With `isolate=False`, the caller would be saying that they don't want/need unique identifiers to be included on the exported SVG.  That could match and perhaps improve upon the Airflow usecase (https://github.com/apache/airflow/pull/27960), producing smaller SVG artifacts from the [Breeze CI tooling](https://github.com/apache/airflow/blob/b5084bcef9c4b0304a1643f891e1b8c80c8b0621/BREEZE.rst), without requiring a [magic constant](https://github.com/apache/airflow/pull/27960/commits/4b56d3d7a56b34701ddcaeed921b55eab6244a5e#diff-c3cfdecea722cd31d4d30a978fb7d0ddd1c28538abee29fcfaf0040a5e2e3501R453).

As I understand it, the reason to generate per-tag identifiers was to limit the scope of the CSS rules emitted by the SVG export (necessary because [textually-embedded SVG elements in XML](https://www.w3.org/TR/SVG11/styling.html#Scope) share their stylesheets with the document).

[In SVG2, shadow-trees may provide a way to ensure that stylesheets are encapsulated](https://www.w3.org/TR/SVG2/struct.html#UseStyleInheritance) -- opening a migration path towards removal of the `unique_id` value (and maybe the `zlib` dependency).

So far, I'm unaware of any other codebases on GitHub that are passing custom values to the `unique_id` argument.

ashb: Svg2 is a dead spec with (next to?) no browser support to my knowledge

ashb: And zlib is part of the standard library

jayaddison: With those responses noted, are there elements of the suggestion that you do agree with or do you consider it a no-go?

PR #2538, opened 2022-09-22
fix default value of ensure_ascii to match comments

onlyacat:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Make the behaviour of `ensure_ascii` parameter match the comments and the documents.

The default value of `ensure_ascii` is supposed to be **False**, but it is **True** in the master branch, according to the comments or the documents, like:

https://github.com/Textualize/rich/blob/bfea226da7baddd3c5e31c9f0c0d2f2ca45e4c7c/rich/__init__.py#L84

and

https://github.com/Textualize/rich/blob/bfea226da7baddd3c5e31c9f0c0d2f2ca45e4c7c/rich/__init__.py#L98

---

I am not sure of the original preference, but I think the escaping result may be confusing since it looks like an additional unexpected process.

Thus I change the default value to **False**.


onlyacat:
PTAL /cc @willmcgugan

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix NO_COLOR support on legacy Windows https://github.com/Textualize/rich/pull/2458
 - Fix pretty printer handling of cyclic references https://github.com/Textualize/rich/pull/2524
 - Fix missing `mode` property on file wrapper breaking uploads via `requests` https://github.com/Textualize/rich/pull/2495
+- Fix mismatching default value of prarmeter `ensure_ascii` https://github.com/Textualize/rich/pull/2538
```
willmcgugan, line 24: typo

willmcgugan (approved):

PR #2554, opened 2022-10-01
Implement Text comparator methods; fix Text equality checking; implement Text hashing for use as dict keys

michelcrypt4d4mus:
## Type of changes

- [x] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

1. Implement `Text.__hash__()` to enable `Text` object as `dict` keys. Given that the markup version of any `Text` object is basically a unique string it seemed reasonable to just call the built in `hash()` method on the markup.
2. "Fix" the equality operator to take the style of a `Text` object into account in addition to the underlying string and the spans in that `Text` object. Seemed pretty odd to check the spans but not the overall style. (Hopefully you regard this as an actual fix.) Had to update a few old tests in the process.
3. Implement `Text.__lt__` and `Text.__le__`, which enables `Text` objects to be compared with `<` and `>` in addition to `==`, and thus also allows lists/tuples of `Text` objects to be sorted with the `sorted()` function.

The overall goal here was to move `Text` closer to a world where it can actually be used interchangeably with `str` as the documentation seems to imply it would like to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Document using `None` as name in `__rich_repr__` for tuple positional args https://github.com/Textualize/rich/pull/2379
 - Add `font_aspect_ratio` parameter in SVG export https://github.com/Textualize/rich/pull/2539/files
 - Added `Table.add_section` method. https://github.com/Textualize/rich/pull/2544
-- Implemented `Text.__hash__` method (enables `Text` objects to be used as `dict` keys).
-- Implemented `Text.__le__` and `Text.__lt__` to enable comparison (and thus also sorting) of `Text` objects in a reproducible way based on the string, the style, and the `Span`s.
+- Implemented `Text.__hash__` method (enables `Text` objects to be used as `dict` keys). https://github.com/Textualize/rich/pull/2554
+- Implemented `Text.__le__` and `Text.__lt__` to enable comparison (and thus also sorting) of `Text` objects in a reproducible way based on the string, the style, and the `Span`s. https://github.com/Textualize/rich/pull/2554
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


willmcgugan:
Text objects are mutable, which makes them unsuitable for use as dictionary keys.

I don't know if the `__eq__` was a fix or not. It's not clear if anything was broken. There are a number of attributes not included in the comparison, which suggests the current version is fine. Without a clear bug to fix, this is a somewhat arbitrary change, more likely to break other projects.

The `lt` change allows you to sort things. But why would you expect that to be the order? What problem is being solved?

Appreciate the effort, but I can't merge this. If you would like to contribute, I'd suggesting picking a bug and focusing your efforts on fixing that.

michelcrypt4d4mus:
1. **re: `lt`**  - the goal was to be able to sort `Text` objects like strings (alphabetically). this was the motivating reason for the PR. the other elements of the ordering (style, etc) were just to get sorting closer to reproducible so that sorting the same group of `Text` objects would not yield different results at different times (and of course it still could, given `justify` etc, but it would be an extreme edge case).
2. **re: `eq`** - i hear you about possibly breaking other projects; in fact originally i had confined this new definition of `eq` to the `le` method.  current implementation seems kind of weird though, given that a `Text` object's style is arguably a lot more significant than its spans.  _/shrug_
3. **re: `hash`** my bad. i threw that in there bc it was an annoyance i ran into; didn't think it through.

the only change i actually value is the `lt` change but if you can't merge it no harm no foul; thanks for the great library.

michelcrypt4d4mus:
@willmcgugan is there any kind of `Text` sorting functionality you'd be amenable to? happy to open a PR for it if so, even if it's just sorting by the underlying string¹ . 

i've been using `rich` to handle [large-ish aggregations of data](https://github.com/michelcrypt4d4mus/pdfalyzer#rich-tree-view) prior to rendering it and while there are obviously uncomplicated solutions it would be nice if i could just sort lists of `Text` without needing a workaround.

¹ obviously sorting just by the string would leave `Text` in the sort of uncomfortable situation where two `Text` objects could be neither greater than, less than, or equal to each other if they have the same string but different `Span` properties, but maybe that's fine.

willmcgugan:
Neat project! Are you familiar with the key argument to sort and sorted functions?

michelcrypt4d4mus:
yes i'm familiar (that's what i mean by an uncomplicated workaround). not having a default sort is definitely not the end of the world but it's mildly annoying from time to time, hence the PR.  a concrete example: if you have prepped a bunch of `Text` objects for entry into a `Table` you can't just call `sorted` to alphabetize them.

(my other annoyance i'd be happy to open a PR about is the fact that you can't instantiate a `Table` with a `Text` header, only with a `str` or `Column`... but `Column` is happy to take a `Text` obj as a header)

Issue #2468, opened 2022-08-16
[BUG] rich.pretty.pprint is too eager to highlight a closing >


davep:
**Describe the bug**

Given this code:

```python
from rich.pretty import pprint as print

class TestClass:

    def __repr__( self ) -> str:
        return '<Outer: some: "value", plus: <Some: other: "child object">, and: 42>'

print( TestClass() )
```

`pprint` applies bold to the first instance of `<` and the first instance of '>', resulting in an unbalanced pair of highlights if there is another instance of a '</>'-wrapped object in the `__repr__` return value.

![Screenshot 2022-08-16 at 11 28 31](https://user-images.githubusercontent.com/28237/184858224-97ea6d39-0e7d-400a-bae8-5d4dc1633eaf.png)

**Platform**
macOS, iTerm2.

PR #2565, opened 2022-10-10
Make repr tag highlighting greedy

davep:
Addressing #2468; before this change the tag-matching code was non-greedy, resulting in an unbalanced match if there were tags within tags. This change makes this greedy to ensure that there's a better chance of the match being balanced.

## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

### DIFF
--- a/rich/highlighter.py
+++ b/rich/highlighter.py
@@ -82,7 +82,7 @@ class ReprHighlighter(RegexHighlighter):
 
     base_style = "repr."
     highlights = [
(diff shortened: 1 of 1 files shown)


davep:
Does this deserve an entry in the `CHANGELOG` and, if so, what's the best method of adding it knowing that the major version will bump due to https://github.com/Textualize/rich/pull/2567?

willmcgugan:
Great. Please add a test for this.

PR #2603, opened 2022-10-20
chore: apply some refurb suggestions

henryiii:
Signed-off-by: Henry Schreiner <henryschreineriii@gmail.com>

## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

After the mention on PythonBytes, I thought I'd try running refurb on rich - and I actually had already tried this earlier! These were the changes that looked pretty reasonable; I didn't make the ones that didn't seem better to me, and I also skipped the "else" ones, which I do think are better but can be a bit polarizing.

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -758,7 +758,7 @@ class Console:
         self._is_alt_screen = False
 
     def __repr__(self) -> str:
-        return f"<console width={self.width} {str(self._color_system)}>"
+        return f"<console width={self.width} {self._color_system}>"
 
     @property
     def file(self) -> IO[str]:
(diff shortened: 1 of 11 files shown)


rich/console.py
```diff
@@ -758,7 +758,7 @@ def __init__(
         self._is_alt_screen = False
 
     def __repr__(self) -> str:
-        return f"<console width={self.width} {str(self._color_system)}>"
+        return f"<console width={self.width} {self._color_system}>"
```
henryiii, line 761:
```suggestion
        return f"<console width={self.width} {self._color_system!s}>"
```
Oops, `format(num, "")` is the int for an IntEnum.

willmcgugan (approved):

PR #2605, opened 2022-10-22
Add information about styleze range

Bernardoow:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

**`TLDR`** MR that add information to how use the `stylize_range` with `syntax object`.

I used the rich lib to build a little script that search code at Gitlab.

To do this, I searched at the source code information to syntax tool whether is possible to highlight part from the text. 
I found that this is possible. I want help the community to avoid this search and be able to get this knowledge just reading the documentation.


willmcgugan:
Thanks. +1 for the docs but I think we need to word it in a similar style to the rest of the doc.

https://github.com/Textualize/rich/issues/2842

PR #2621, opened 2022-10-28
Use `operator.length_hint` to estimate progress total

ionite34:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [ ] (N/A) I've added tests for new code.

## Description

This PR changes `progress.track` to use `operator.length_hint` (which calls `__len__` if exists, or `__length_hint__`) instead of `len()` when total is not provided. Allowing most built-in iterators and external library objects implementing the protocol with known length to have accurate progress tracking (such as `range` and `reversed`).

`operator.length_hint` was added in Python 3.4 as [PEP 424](https://peps.python.org/pep-0424/)

Other progress libraries, such as `tqdm`, already use this method:
https://github.com/tqdm/tqdm/blob/6791e8c5b3d6c30bdd2060c346996bfb5a6f10d1/tqdm/contrib/concurrent.py#L61


willmcgugan (approved):

willmcgugan:
Nice

PR #2635, opened 2022-11-09
Fix syntax error when building with nuitka

cilki:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

[Nuitka](https://github.com/Nuitka/Nuitka) throws a syntax error when building the `rich.pretty` module because of the missing symbol `get_ipython`. Instead of using try/except as control flow, it's better to explicitly check for the existence of this global function. 

Otherwise, rich works well when converted into C with Nuitka. Thanks!


willmcgugan (approved):

willmcgugan:
Interesting. Did you notice any difference with Rich as compiled C?

rich/pretty.py
```diff
@@ -272,7 +272,7 @@ def __call__(self, value: Any) -> Any:
         # replace plain text formatter with rich formatter
         rich_formatter = RichFormatter()
         ip.display_formatter.formatters["text/plain"] = rich_formatter
-    except Exception:
```
ssbarnea, line 275: While I never used nuitka, I know enough to solute removal of "except Exception" ;)

PR #2659, opened 2022-11-18
Show elapsed time when finished in rich.progress.track()

jankatins:
Before it was showing

```
Inserting nodes... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
```

now it will show:

```
Inserting nodes... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:12:03
```

-> Now one can see the elapsed time without adding another way to measure that time...

## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [x] I've adjusted the test for new code.

## Description

Just added the appropriate parameter: `TimeRemainingColumn(elapsed_when_finished=True),` to the created columns.


SertugF (commented):
Normally it shows, estimated time to complete the progress. Making it show another kind of value(time spent for progress) after completion can be confusing for majority of users. That is the main reason its(elapsed_when_finished) default value is set to false. I find making it true is not logical.

jankatins:
The problem is that the `100% 0:00:00` final output is useless -> it's done, but how long did it take?. The best would be to change it to `100% (took xx:xx:xx)`. But that would mean a (slightly) bigger change.

willmcgugan (approved):

willmcgugan:
Does seem like a better default.

PR #2664, opened 2022-11-22
Fix typing of Progress.add_task signature.

rodrigogiraoserrao:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [ ] ~~I've added tests for new code.~~ Does not apply

## Description

The parameter `start` of `Progress.add_task` can be set to `None` for an indeterminate progress bar but the typing of the parameter says it must be a float. This fixes the typing so that static checkers like mypy don't complain about otherwise perfectly sane code.

### DIFF
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -1590,7 +1590,7 @@ class Progress(JupyterMixin):
         self,
         description: str,
         start: bool = True,
-        total: Optional[float] = 100.0,
+        total: Optional[Union[float, None]] = 100.0,
         completed: int = 0,
         visible: bool = True,
         **fields: Any,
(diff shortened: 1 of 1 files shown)


rich/progress.py
```diff
@@ -1590,7 +1590,7 @@ def add_task(
         self,
         description: str,
         start: bool = True,
-        total: Optional[float] = 100.0,
+        total: Optional[Union[float, None]] = 100.0,
```
willmcgugan, line 1593: This isn’t right. Optional means the same as a union with None.

willmcgugan (changes requested):

Issue #2338, opened 2022-06-14
Added `echo` bool to capture.


willmcgugan:
The `capture` method should have an `echo` parameter. When `echo=True` (the default) text should be captured and written to the terminal. When `echo=False` the text should be captured but not written to the terminal.

See https://github.com/Textualize/rich/issues/2172 for context.

SertugF:
Hello, i have been working on this. If echo is  true by default it breaks the test below. Since new capture method prints all captured strings 'out' changes.

```
def test_capture_and_record(capsys):
    recorder = Console(record=True)
    recorder.print("ABC")

    with recorder.capture() as capture:
        recorder.print("Hello")

    assert capture.get() == "Hello\n"

    recorded_text = recorder.export_text()
    out, err = capsys.readouterr()

    assert recorded_text == "ABC\nHello\n"
    assert capture.get() == "Hello\n"
    assert out == "ABC\n"
```

terminal output : 
![image](https://user-images.githubusercontent.com/45711819/205930521-71f2ac58-febd-483a-9762-9fddec3b92da.png)

In order to pass previous tests should i just change default 'echo' value to 'false' ? Or does this test needs to be replaced ?

SertugF:
Okay, below is what you recommend then ?

```
def test_capture_and_record(capsys):
    recorder = Console(record=True)
    recorder.print("ABC")

    with recorder.capture(echo=False) as capture:
        recorder.print("Hello")

    assert capture.get() == "Hello\n"

    recorded_text = recorder.export_text()
    out, err = capsys.readouterr()

    assert recorded_text == "ABC\nHello\n"
    assert capture.get() == "Hello\n"
    assert out == "ABC\n"
```

While trying solve the problem, i have seen that even tho my code works for single recorder.print() call, it actually breaks when there is multiple of them. Since there is no test for that.  It passes. So i have made test below in order to future-proof it. Can somebody give some feedback about it.

```
def test_capture_and_record_multiple(capsys):
    recorder = Console(record=True)

    with recorder.capture(echo=False) as capture:
        recorder.print("Hello")
        recorder.print("World")

    assert capture.get() == "Hello\nWorld\n"

    recorded_text_echo_false = recorder.export_text()
    out, err = capsys.readouterr()

    assert recorded_text_echo_false == "Hello\nWorld\n"
    assert out == ""

    with recorder.capture(echo=True) as capture:
        recorder.print("Foo")
        recorder.print("Bar")

[...]

PR #2697, opened 2022-12-09
added a new parameter for method capture(echo)

SertugF:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [X] Tests
- [ ] Other

## Checklist

## Description

Added echo parameter for capture method. If value of echo is true, text is captured and showed in the terminal. If value of echo is false, text is captured but not showed in the terminal. Fixed earlier test for new argument and added new test. Changed documentation for capture. https://github.com/Textualize/rich/issues/2338

### DIFF
--- a/docs/source/console.rst
+++ b/docs/source/console.rst
@@ -313,10 +313,10 @@ Note that when writing to a file you may want to explicitly set the ``width`` ar
 Capturing output
 ----------------
 
-There may be situations where you want to *capture* the output from a Console rather than writing it directly to the terminal. You can do this with the :meth:`~rich.console.Console.capture` method which returns a context manager. On exit from this context manager, call :meth:`~rich.console.Capture.get` to return the string that would have been written to the terminal. Here's an example::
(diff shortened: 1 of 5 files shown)


ssbarnea:
@willmcgugan I do not think that requiring approval to even run CI is making rich very welcoming to new folks. There is the risk the might lose interest before even having the first successful contribution.

ssbarnea (approved):

docs/source/console.rst
```diff
@@ -313,10 +313,10 @@ Note that when writing to a file you may want to explicitly set the ``width`` ar
 Capturing output
 ----------------
 
-There may be situations where you want to *capture* the output from a Console rather than writing it directly to the terminal. You can do this with the :meth:`~rich.console.Console.capture` method which returns a context manager. On exit from this context manager, call :meth:`~rich.console.Capture.get` to return the string that would have been written to the terminal. Here's an example::
+There may be situations where you want to *capture* the output from a Console without writing it directly to the terminal. You can do this with the :meth:`~rich.console.Console.capture` method using echo parameter which returns a context manager. If echo value is set to True(default) text is captured and written to the terminal. When echo value is set to False text captured but not written to the terminal.  On exit from this context manager, call :meth:`~rich.console.Capture.get` to return the string that would have been written to the terminal. Here's an example::
```
jankatins, line 316: default should be False, given that this was the default before

SertugF: Check #2338  "The capture method should have an echo parameter. When echo=True (the default) text should be captured and written to the terminal. When echo=False the text should be captured but not written to the terminal." @willmcgugan wants it true default.

PR #2699, opened 2022-12-12
Add 'encapsulated' boolean parameter (default: true) to rich.console.export_svg method

jayaddison:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

- Adds a (filterable) warning when the `unique_id` parameter added by #2537 is used.
- Adds an `encapsulated` parameter that is intended to provide a migration path towards the removal of the `unique_id` variable in future

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -2381,9 +2381,9 @@ class Console:
             import warnings
 
             msg = (
-                "Using non-checksum prefix for SVG elements.  Non-unique prefixes have "
-                "the potential to cause namespace collisions when an exported SVG is "
-                "embedded into another document."
+                "Using non-checksum prefix for SVG element export.  Non-unique "
+                "prefixes have the potential to cause namespace collisions when an "
+                "exported SVG is embedded into another document."
             )
             warnings.warn(msg)
 
--- a/rich/console.py
+++ b/rich/console.py
@@ -2377,7 +2377,10 @@ class Console:
             if clear:
                 self._record_buffer.clear()
 
-        if unique_id is not None or encapsulated is False:
+        if encapsulated is False:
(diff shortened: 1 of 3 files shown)


rich/console.py
```diff
@@ -2373,6 +2377,18 @@ def stringify(value: object) -> str:
...
+            import warnings
+
+            msg = (
+                "'unique_id' parameter passed to SVG export.  This has the potential "
+                "to cause namespace collisions if the SVG is embedded in a document."
+            )
+            warnings.warn(msg)
+
+        if encapsulated is False:
+            unique_id = "terminal"
```
jayaddison, line 2390: Hmm.. this probably has the potential to cause as-many, if not more namespace collisions than (most?) user-selected `unique_id` parameters.

jayaddison: (the idea is that `encapsulated=False` would only be set when the caller knows they intend to place the SVG in a context where collisions shouldn't occur.  is that a reasonable expectation, though?)

jayaddison: I don't think that the original caller is the problem, to be honest - if anyone discovers this parameter and provides a specific value, then _they_ probably knew what they were doing.

The problem is that the document that is emitted may end up being re-used and embedded in other contexts that the caller themselves may not be aware of.

To that extent, encapsulation should really be near-required for formats that _don't_ encapsulate by default.  So we should raise the same warning here -- unless the output format is _known_ to be self-encapsulating.

jayaddison:
(closing to clean up some stale branches and pull requests; commits recoverable with some `git` magic if-so-desired; I consent to these commits being merged if later considered valuable)

PR #2723, opened 2022-12-29
Add support for `pipx run rich`

cjolowicz:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Allow users to quickly try out rich in their terminal by typing `pipx run rich`. Running the demo does not require an explicit and permanent installation of rich. It's equivalent to running `python -m rich` from a temporary virtual environment.

This change declares an entry-point for `pipx run`. It does _not_ add a console script for rich, nor does it enable installation using `pipx install`.

There is some potential for confusion with `pipx install rich-cli`, which adds a `rich` command to the system. This needs to be weighed against the convenience of getting a quick demo of rich's capabilities. Personally, I like the convenience--your call though.

## Manual tests

Test the basic feature using the following commands:

```
$ poetry build
$ pipx run --spec dist/rich-13.0.0-py3-none-any.whl rich
[output from python -m rich]
```

Test that you still can't install rich as an app like this:

```
$ poetry build
$ pipx install dist/rich-13.0.0-py3-none-any.whl
No apps associated with package rich. [...]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add type annotation for key_separator of pretty.Node https://github.com/Textualize/rich/issues/2625
-
+- Add support for `pipx run rich` to run a demo without a separate installation step
 
 ## [12.6.0] - 2022-10-02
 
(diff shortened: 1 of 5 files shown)


willmcgugan:
I fear it might cause a little confusion. And I’m not aware of folk have much difficulty in running the demo, so I’m going to pass.

But thanks for educating me on these pips entry points!

Issue #2172, opened 2022-04-06
[BUG] Console object failing to capture in a context manager


GeoWill:
**Describe the bug**

I might have been holding it wrong, but up until version  `12.0.1` the following minimal example gave the expected output. 

Code:
```python 
from rich.console import Console
recorder = Console(record=True)
with recorder.capture():
     recorder.print("foo")
recorder.export_text()
```

Output:
`''`

Expected output:
`'foo\n'`

First appears in version `12.0.1` for me. 

**Platform**
<details>
<summary>Click to expand</summary>

Python 3.6.12 on Ubuntu 20.04 

I may ask you to copy and paste the output of the following commands. It may save some time if you do it now.

If you're using Rich in a terminal:

```
python -m rich.diagnose
```

```
╭─────────────────────── <class 'rich.console.Console'> ────────────────────────╮
│ A high level console interface.                                               │
│                                                                               │
│ ╭───────────────────────────────────────────────────────────────────────────╮ │
│ │ <console width=81 ColorSystem.EIGHT_BIT>                                  │ │
│ ╰───────────────────────────────────────────────────────────────────────────╯ │
│                                                                               │
│     color_system = '256'                                                      │
│         encoding = 'utf-8'                                                    │
│             file = <_io.TextIOWrapper name='<stdout>' mode='w'                │
│                    encoding='utf-8'>                                          │
│           height = 41                                                         │
│    is_alt_screen = False                                                      │
│ is_dumb_terminal = False                                                      │
│   is_interactive = True                                                       │
│       is_jupyter = False                                                      │
│      is_terminal = True                                                       │
│   legacy_windows = False                                                      │
│         no_color = False                                                      │
│          options = ConsoleOptions(                                            │
│                        size=ConsoleDimensions(width=81, height=41),           │
│                        legacy_windows=False,                                  │
│                        min_width=1,                                           │
│                        max_width=81,                                          │
│                        is_terminal=True,                                      │
│                        encoding='utf-8',                                      │
│                        max_height=41,                                         │
│                        justify=None,                                          │
│                        overflow=None,                                         │
│                        no_wrap=False,                                         │
│                        highlight=None,                                        │
│                        markup=None,                                           │
│                        height=None                                            │
│                    )                                                          │
│            quiet = False                                                      │
│           record = False                                                      │
│         safe_box = True                                                       │
│             size = ConsoleDimensions(width=81, height=41)                     │
│        soft_wrap = False                                                      │
│           stderr = False                                                      │
[...]

PR #2741, opened 2023-01-10
Capture output goes to terminal by default

tinvaan:
Capture and console.capture() accepts a kwargs, where you may specify echo=False to silence output being sent to terminal.

## Type of changes

- [x] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Attempts a fix for https://github.com/Textualize/rich/issues/2172 by implementing @willmcgugan's [suggestion](https://github.com/Textualize/rich/issues/2172#issuecomment-1154942255)

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -327,10 +327,10 @@ class Capture:
         console (Console): A console instance to capture output.
     """
 
-    def __init__(self, console: "Console", **kwargs: Any) -> None:
+    def __init__(self, console: "Console", echo: bool=True) -> None:
+        self._echo = echo
         self._console = console
         self._result: Optional[str] = None
-        self._echo = kwargs.get("echo", True)
 
     def __enter__(self) -> "Capture":
         self._console.begin_capture()
@@ -344,7 +344,7 @@ class Capture:
     ) -> None:
         self._result = self._console.end_capture()
         if self._echo:
-            self._console.out(self._result)
+            self._console.print(self._result)
 
     def get(self) -> str:
         """Get the result of the capture."""
@@ -1083,7 +1083,7 @@ class Console:
         """Play a 'bell' sound (if supported by the terminal)."""
         self.control(Control.bell())
 
-    def capture(self, **kwargs: Any) -> Capture:
+    def capture(self, echo: bool=True) -> Capture:
         """A context manager to *capture* the result of print() or log() in a string,
         rather than writing it to the console.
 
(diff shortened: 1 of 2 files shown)


rich/console.py
```diff
@@ -327,9 +327,10 @@ class Capture:
         console (Console): A console instance to capture output.
     """
 
-    def __init__(self, console: "Console") -> None:
+    def __init__(self, console: "Console", **kwargs) -> None:
```
willmcgugan, line 330: Don't see any reason to use kwargs here. It breaks typing.

rich/console.py
```diff
@@ -1080,7 +1083,7 @@ def bell(self) -> None:
         """Play a 'bell' sound (if supported by the terminal)."""
         self.control(Control.bell())
 
-    def capture(self) -> Capture:
+    def capture(self, **kwargs) -> Capture:
```
willmcgugan, line 1086: No need for `kwargs`.

rich/console.py
```diff
@@ -342,6 +343,8 @@ def __exit__(
         exc_tb: Optional[TracebackType],
     ) -> None:
         self._result = self._console.end_capture()
+        if self._echo:
+            self._console.out(self._result)
```
willmcgugan, line 347: This is going to run captured input through `print` again, which may modify the result.

tests/test_console.py
```diff
@@ -379,22 +379,33 @@ def test_capture():
         console.print("Hello")
     assert capture.get() == "Hello\n"
 
+    with console.capture(echo=False) as capture:
```
willmcgugan, line 382: This doesn't test that `echo` is working.

rich/console.py
tinvaan: right, guess I'll invoke `print` directly then.

rich/console.py
tinvaan: hmm guessing you're suggesting a default arg instead - `def __init__(..., echo: bool=True)`?

tinvaan:
@willmcgugan I've addressed the review comments, but I see locally that my change breaks the `pager` tests(`tests/test_console.py::test_pager`). 

However, printing the test variable to console shows the text styled as expected(**Hello World**) but we're asserting for a plain string in the tests. Shouldn't it be the other way round? What am I missing?

Screenshot 2023-01-13 at 8 51 28 AM

tinvaan:
@willmcgugan *poke*?

willmcgugan:
There are other considerations you may not be aware of. You might want to leave this one to us.

tinvaan:
> There are other considerations you may not be aware of. You might want to leave this one to us.

Ok, I'll keep the PR open for anyone who might want to attempt a fix. Thanks!

Issue #2745, opened 2023-01-11
[BUG] Panel title/subtitle styles are overriden by border style


noprobelm:
I'll echo the others here by first expressing my thanks for making such a great tool. I'm an enthusiast, and `rich` has become an integral part of my daily code. I greatly admire your work.

To the problem...

- [x] I've checked [docs](https://rich.readthedocs.io/en/latest/introduction.html) and [closed issues](https://github.com/Textualize/rich/issues?q=is%3Aissue+is%3Aclosed) for possible solutions.
- [x] I can't find my issue in the [FAQ](https://github.com/Textualize/rich/blob/master/FAQ.md).

## Problem Description

I'm having an issue with panels. If a `Panel` instance has stylized `title` and `subtitle` attributes, their styles will invariably be overwritten by whatever the `border_style` is at the time of rendering (if a border style is present). I would expect this kind of behavior if `title` and `subtitle` were without styles, but not when they've been explicitly defined in the `Panel` instance.

## How to Recreate

This code will recreate the problem.

```
from rich.panel import Panel
from rich.text import Text
from rich import print

title = Text("Wake up!", "bold #ffb703")
subtitle = Text("The world says hello!", "italic #f64740")
renderable = Text(f"A panel with 'title' and 'subtitle' elements")

panel = Panel(
    renderable=renderable,
    title=title,
    subtitle=subtitle,
    border_style="#5EAD7A",
    style='bold #81adc8'
)

print(panel)

panel = Panel(
    renderable=renderable,
    title=title,
    subtitle=subtitle,
)
panel.style='bold #81adc8'
print(panel)

```

Which produces

![unexpected](https://user-images.githubusercontent.com/7587293/211927243-10498cc2-aa07-4e77-969b-d3643e16459a.png)

I would expect the title/subtitles in the resulting panels would be stylized according to what's specified in the `Panel` instance, like this

![expected](https://user-images.githubusercontent.com/7587293/211927234-f6febe4b-f8c8-453a-9f4d-705ab4272843.png)

## Root Cause
I think this problem is due to the way the  `__rich_console__` method for `Panel` overwrites the styling for `title_text` and `subtitle_text` with `border_style` (if `Panel.border_style != "none"`), irrespective of whether styling for `Panel.title` or `Panel.subtitle` were defined already. The code in question:

[rich.panel.Panel; 198-199](https://github.com/Textualize/rich/blob/master/rich/panel.py#L198-L199)
[...]

PR #2746, opened 2023-01-11
Fix Panel logic for selecting priority of title/subtitle styling 

noprobelm:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This is a proposed fix for the issue described in #2745.

- The existing behavior of the `Panel.__rich_console__` method is to check if `title_text` and `subtitle_text` are present, and if so, apply the Panel's border style indicated by `border_style`.
- My proposed change is to extend the existing conditional statements to refrain from setting the styles for `title_text` and `subtitle_text` if they were already defined in the `Panel` instance when the console protocol was initiated.


willmcgugan:
There is a fix for this in main.

PR #2752, opened 2023-01-14
Fix splitting color definitions when parsing style

IsaacBreen:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Problem

The `rich` library's `Style` class had an issue where it would fail to parse style definitions that contained a color definition with spaces, such as `"rgb(0, 0, 0)"`. This would cause exceptions to be raised and caught silently, leading to unexpected behavior in places like `Text.stylize` and `Syntax`.

Style definitions that contain a color definition with a space fail to parse. For example:

```python
>>> from rich.style import Style
>>> Style.parse("rgb(0,0,0)")  # OK
Style(bgcolor=Color('rgb(0,0,0)', ColorType.TRUECOLOR, triplet=ColorTriplet(red=0, green=0, blue=0)))
>>> Style.parse("rgb(0,0,0)")
...
StyleSyntaxError: unable to parse 'rgb(0,' as background color; 'rgb(0,' is not a valid color
```

Furthermore, this exception gets caught silently when `get_style` is called with a `default` argument, as it is in `Text.render`, causing weird behaviour like this:

Screenshot 2023-01-14 at 4 00 50 pm

The issue is that style definitions are tokenized with `style_definition.split()`. For example, `"on rgb(0, 0, 0)"` gets becomes `["on", "rgb(0,", "0,", "0)"]` rather than `["on", "rgb(0, 0, 0)"]` as it should be.

Color definitions, on the other hand, are parsed with `rich.color.RE_COLOR`, which ignores whitespaces.

This PR basically replaces `style_definition.split()` with `RE_STYLE_DEFINITION.finditer(style_definition)`, where `RE_STYLE_DEFINITION` embeds the color regex:

```python
RE_STYLE_DEFINITION = re.compile(f"({RE_COLOR.pattern})|\S+", re.VERBOSE)
```

Screenshot 2023-01-14 at 4 14 03 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Tokenize style definitions using a regex that ignores whitespaces in color definitions.
+
 ## [13.0.1] - 2023-01-06
 
 ### Fixed
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -60,3 +60,4 @@ The following people have contributed to the development of Rich:
 - [Serkan UYSAL](https://github.com/uysalserkan)
 - [Zhe Huang](https://github.com/onlyacat)
 - [Ke Sun](https://github.com/ksun212)
+- [Isaac Breen](https://github.com/IsaacBreen)
--- a/rich/style.py
+++ b/rich/style.py
@@ -1,3 +1,4 @@
+import re
 import sys
 from functools import lru_cache
 from marshal import dumps, loads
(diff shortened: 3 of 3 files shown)


IsaacBreen:
Fixed a typing issue. Should pass now.

willmcgugan:
Seems reasonable, and tests are passing. But there may be unintended consequences.

Consider this:

```python
Color.parse("rgb(0,0,0)") == Color.parse("rgb(0,0, 0)")
```

That is False with this update. Suggest we need to normalise the color name.

rich/color.py
```diff
@@ -289,7 +289,7 @@ class ColorParseError(Exception):
     """The color could not be parsed."""
 
 
-RE_COLOR = re.compile(
+RE_COLOR: Pattern[str] = re.compile(
```
willmcgugan, line 292: The annotation is not required here, as Mypy can deduce it quite easily.

willmcgugan:
I'll need some tests to merge this.

@IsaacBreen I know its been a while. Let me know if you still want to work on this.

willmcgugan:
Closing, assumed stale. Feel free to reopen if you still want to work on this.

PR #2806, opened 2023-02-13
match expected IPython formatter spec in rich.pretty

### LINKED ISSUE #2803: duplicated plots in jupyter
How can we avoid having duplicated plots on jupyter?

When I run

```
import gdsfactory as gf
gf.components.bend_circular()
```

I get one plot in jupyter, thanks to  defining `def _repr_html_(self)`

However when I run 
```
from rich import pretty

pretty.install()
gf.components.bend_circular()
```

I get two plots of the same component.

minrk:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

More context [in #2803](https://github.com/Textualize/rich/issues/2803#issuecomment-1427886362), but the IPython formatter spec expects slightly different behavior than the hook currently in `rich.pretty`. In particular:

1. formatters should _return_ their output, not publish it on a side channel, and
2. registered formatters should not gate their behavior on what other formatters may be registered

From looking through the code, I think these two are related, because when multiple formats are produced in a mimebundle, a Jupyter frontend displays only one. But because the IPython hook _printed_ output instead of returning it, it was always _in addition to_ other reprs - making it necessary to check if it was probably going to be the only one. Following the spec and returning output removes the need to check if it's likely to be the only output.

This PR changes the registered formatter to match IPython's expectations - _return_ the text representation instead of publishing it separately, and remove checks for possible other formatters. This will fix all cases where duplicate output was produced because the checks were incomplete (not accounting for `Formatter.for_type`, for instance). The only downside is one of size/performance if rich text output is expensive and/or large, but might not be seen because. This is standard for IPython and Jupyter output, however, and Rich shouldn't be blamed for it.

closes #2803

### DIFF
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -133,31 +133,6 @@ def _ipy_display_hook(
 
     console = console or get_console()
     if console.is_jupyter:
-        # Delegate rendering to IPython if the object (and IPython) supports it
-        #  https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display
-        ipython_repr_methods = [
-            "_repr_html_",
-            "_repr_markdown_",
-            "_repr_json_",
-            "_repr_latex_",
-            "_repr_jpeg_",
-            "_repr_png_",
-            "_repr_svg_",
-            "_repr_mimebundle_",
-        ]
-        for repr_method in ipython_repr_methods:
-            method = getattr(value, repr_method, None)
-            if inspect.ismethod(method):
-                # Calling the method ourselves isn't ideal. The interface for the `_repr_*_` methods
-                #  specifies that if they return None, then they should not be rendered
-                #  by the notebook.
-                try:
-                    repr_result = method()
-                except Exception:
-                    continue  # If the method raises, treat it as if it doesn't exist, try any others
-                if repr_result is not None:
-                    return  # Delegate rendering to IPython
-
         # When in a Jupyter notebook let's avoid the display of some specific classes,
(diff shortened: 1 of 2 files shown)


rich/pretty.py
```diff
@@ -133,58 +133,37 @@ def _ipy_display_hook(
...
-                except Exception:
-                    continue  # If the method raises, treat it as if it doesn't exist, try any others
-                if repr_result is not None:
-                    return  # Delegate rendering to IPython
-
         # When in a Jupyter notebook let's avoid the display of some specific classes,
         # as they result in the rendering of useless and noisy lines such as `<Figure size 432x288 with 1 Axes>`.
         # What does this do?
         # --> if the class has "matplotlib.artist.Artist" in its hierarchy for example, we don't render it.
         if is_object_one_of_types(value, JUPYTER_CLASSES_TO_NOT_RENDER):
```
minrk, line 140: Having fixed the duplicate output, I'm not sure if this behavior is desired anymore - this was caused because matplotlib formatters were registered via [formatter.for_type](https://github.com/ipython/ipython/blob/9663a9adc4c87490f1dc59c8b6f32cdfd0c5094a/IPython/core/pylabtools.py#L284), so the check above didn't catch that a formatter was registered.

It would make sense to me to either remove this altogether, or `return repr(value)`, because as it is, if inline backend is not enabled, displaying or returning a figure will show _nothing_, rather than the simple repr of the object.

willmcgugan: I'm happy to remove that. It was a clearly a hack anyway.

willmcgugan (approved):

PR #2811, opened 2023-02-15
Improve Traceback Frame Suppression

debanjum:
## Type of changes

- [X] New feature

## Checklist

### Code Change 
- 266337ad Allow setting max frames in traceback logging
- 5051b47f Allow using regex paths to suppress frame filenames in traceback
  - Generalize suppressing frame filenames by supporting regex matching
    Example: `traceback_suppress=[r"/.*fastapi*"]` will match all frame
    filenames matching a fastapi library in stacktrace
  
  - Maintain backward compatibility to check if frame filename starts
    with path, if plain string passed
    Example: `traceback_suppress=["fastapi"]` will continue to match frame
    filenames starting with fastapi in stacktrace
  
    *Note: Not sure how practically useful this previous behavior is*

### Examples
- To suppress fastapi, uvicorn related frames in traceback
  ``` python
  RichHandler(rich_tracebacks=True, tracebacks_suppress=['/.*fastapi.*', '/.*uvicorn.*'])
  ```
- To only show a maximum of 7 frames in traceback
   ``` python
  RichHandler(rich_tracebacks=True, traceback_max_frames=7)
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [13.3.2] - Unreleased
+
+### Added
+
+- Added `traceback_max_frames` parameter to `logging.RichHandler`
+
+### Changed
+
+- Can now suppress frames using regexes passed via `suppress` parameter to `traceback.Traceback`
+
 ## [13.3.1] - 2023-01-28
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


willmcgugan:
I'm not keen on the regex approach. The chances are too high that you end up excluding something accidentally. I'd consider callbacks or some other protocol to do this.

debanjum:
That's unfortunate. The current traceback filtering seems overtly limited. I was hoping extending Rich with regex and allowing to limit the `traceback_max_frames` would be a welcome change as it'll provide more control to developers.

jessrosenfield:
hey there! following up here - i don't have a strong opinion on approach but want to emphasize that the regex approach would've satisfied an unmet need of mine:

would love to have an explicitly allowed list of tracebacks to display, rather than an exhaustive list of what to suppress! with `tracebacks_show_locals` on I *only* want to view tracebacks on my own project's packages - any external library should be suppressed for my purposes

PR #2873, opened 2023-03-11
use object.__new__(Style) consistently

The-Alchemist:
use `object.__new__(Style)` instead of `(self|cls).__new__(Style)`

* `__new__` appears to be a class method (https://docs.python.org/3/reference/datamodel.html#object.__new__)
* usage is now consistent, whereas before both `self.__new__` and `cls.__new__` were used
* I believe this usage of `__new__` is more "official" Python

Discovered while trying to use `rich` with GraalPython (https://github.com/oracle/graalpython).  Works great with this fix. :)

## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

### DIFF
--- a/rich/style.py
+++ b/rich/style.py
@@ -210,7 +210,7 @@ class Style:
             color (Optional[Color]): A (foreground) color, or None for no color. Defaults to None.
             bgcolor (Optional[Color]): A (background) color, or None for no color. Defaults to None.
         """
-        style: Style = cls.__new__(Style)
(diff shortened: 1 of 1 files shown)


The-Alchemist:
I believe `self.__new__(cls)` will soon be available in GraalPython, so not sure if this PR is still relevant...

willmcgugan:
Doesn't seem relevant now.

PR #2884, opened 2023-03-20
expand requirements

ucodery:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This is a small issue, but some packages that rich imports it does not also directly depend on.
The first one is the highest risk; since rich doesn't even require setuptools to build, if it isn't preinstalled pkg_resources will never be there.
The last commit was mostly about convenience as there was obviously an acknowledgement that emoji might not be present, but it seemed odd to promote mixing poetry and pip in a dev environment.
The versions are all arbitrary, as the features used from each package are all quite old.


willmcgugan:
I don't think this is needed. Requirements for one-off tools used mostly internally, don't warrant inclusion in dev dependancies. It increases the risk of conflicts.

PR #2891, opened 2023-03-22
fix: bug: python REPL evaluated result ignores custom highlighter

### LINKED ISSUE #2714: [BUG] iPython ignores custom highlighter

**Describe the bug**

Not sure if it's a bug per se, but it was unexpected behavior. I wanted to add some custom highlighting rules to iPython, so I made a custom highlighter object and added the following to `~/.ipython/profile_default/startup/`:

```python
from rich import reconfigure
from rich.highlighter import RegexHighlighter
from rich.pretty import install

class CustomReprHighlighter(RegexHighlighter):
    base_style = "repr."
    highlights = [
        # ... my custom highlights ...
    ]

reconfigure(highlighter=CustomReprHighlighter())
install()
```

The custom highlighting shows up in `rich.print`, but not in the `Out[]` cells. It seems like the iPython display hook is using it's own instance of `ReprHighlighter` and not the one from `get_console()`. This is the workaround I found:

```diff
--- a/pretty.py
+++ b/pretty.py
@@ -174,6 +174,7 @@ def _ipy_display_hook(
         if _safe_isinstance(value, RichRenderable)
         else Pretty(
             value,
+            highlighter=console.highlighter,
             overflow=overflow,
             indent_guides=indent_guides,
             max_length=max_length,
```

**Platform**

* Rich 12.6.0
* Python 3.10.2
* Ubuntu 22.04.1

xsjk:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

The output format of REPL evaluated result does not follow the customed options of console because the `display_hook` prints the `value` without setting the options to the`console`'s. Specify the params of `Pretty` to fix #2714

### DIFF
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -134,7 +134,11 @@ def _ipy_display_hook(
             if _safe_isinstance(value, RichRenderable)
             else Pretty(
                 value,
+                highlighter=console.highlighter,
+                indent_size=console.tab_size,
+                justify=console.options.justify,
                 overflow=overflow,
+                no_wrap=console.options.no_wrap,
                 indent_guides=indent_guides,
                 max_length=max_length,
                 max_string=max_string,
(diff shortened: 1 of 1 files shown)


willmcgugan:
I don't think it makes sense for pretty prints to respect these parameters. If you do need it, then you can construct a Pretty instance manually.

PR #2905, opened 2023-03-27
Text.format() feature

pom11:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [X] Tests
- [ ] Other

## Checklist

## Description
Added `format` method for `Text`. Currently works just for string placeholders like `{name}`. 

Supported placeholders: `{string}`
Unsupported placeholders: `{0}` `{}`

TODO:
Support all types of placeholders

### DIFF
--- a/rich/text.py
+++ b/rich/text.py
@@ -433,6 +433,26 @@ class Text(JupyterMixin):
         )
         copy_self._spans[:] = self._spans
         return copy_self
+        
+    def format(self, *args, **kwargs) -> "Text":
+        """Return a new Text instance with formatted specified value(s) and insert them inside the string's placeholder."""
+        formatted = ""
+        if len(args)!=0:
+            formatted = self.plain.format(*args, **kwargs)
+        elif len(kwargs)!=0:
+            formatted = self.plain.format(**kwargs)
+        if formatted != self.plain and formatted!="":
+            sanitized_text = strip_control_codes(formatted)
+            return Text(
+                sanitized_text,
+                style=self.style,
+                justify=self.justify,
+                overflow=self.overflow,
+                no_wrap=self.no_wrap,
+                end=self.end,
+                tab_size=self.tab_size
+                )
+        return self.copy()
 
     def stylize(
         self,
--- a/rich/text.py
+++ b/rich/text.py
@@ -1,4 +1,5 @@
 import re
+from string import Formatter
 from functools import partial, reduce
 from math import gcd
 from operator import itemgetter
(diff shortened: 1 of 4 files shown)


willmcgugan:
I'm afraid it doesn't work as expected.

Screenshot 2023-03-27 at 13 56 06

It's a worthwhile feature, but it would require full understanding of the Text internals. We would also need comprehensive unit tests.

pom11:
will take a look asap

pom11:
Screenshot 2023-03-27 at 16 14 11
Screenshot 2023-03-27 at 16 14 42
seems to be working for my machine

pom11:
Screenshot 2023-03-27 at 16 46 28
found the problem. 
```
>>> t
``` messes everything up

**Comment by pom11** [comment]
solved.

File: rich/text.py [hunk 1]
```diff
@@ -434,11 +435,94 @@ def copy(self) -> "Text":
...
+        initial_parsing = [x for x in Formatter().parse(self.plain)]
+        if len(initial_parsing) == 1 and initial_parsing[0][1] == None:
+            return self.copy()
+        formatted = ""
+        parsed_len = 0
+        _args = list(args)
+        _kwargs = kwargs
+        _spans = [[s.start, s.end, s.style] for s in self._spans]
+        if len(_args) != 0 or len(kwargs) != 0:
+            _field_numbers = [
```
willmcgugan, line 450: Not clear what is going on here.

tests/test_text.py
```diff
@@ -123,6 +123,13 @@ def test_copy():
     assert text == test_copy
     assert text is not test_copy
 
+def test_format():
```
willmcgugan, line 126: This needs more tests.

Check plain strings, strings with a single substitution, no styles, several styles, overlapping styles, etc. Try to cover as many edge cases as you can.

willmcgugan (changes requested):
The efficiency of this method could be improved.

Readability could use some work. Better named variables etc. `str_index` says what it is, but doesn't say what it is used for. We have a no abbreviation and no single letter variables (with a few exceptions).

A few  well-placed comments would also help to follow the code.

We will need a lot more tests before this could be merged.

rich/text.py
```diff
@@ -434,11 +435,94 @@ def copy(self) -> "Text":
...
+                "{"
+                + x[1]
+                + f'{":"+x[2] if x[2]!="" and x[2]!=None else ""}'
+                + f'{"!"+x[3] if x[3]!="" and x[3]!=None else ""}'
+                + "}"
+                for x in initial_parsing
+                if x[1] != None and not x[1].isdigit() and x[1] == ""
+            ]
+            if len(_field_numbers) != 0 and len(_field_brackets) != 0:
+                raise TypeError(
```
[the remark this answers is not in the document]
pom11: yes

rich/text.py
pom11: converted Span to a list cause can't set attributes to them

pom11:
will refactor as discussed on discord

pom11:
LGTM

pom11:
updates?

willmcgugan:
Sorry, I'm afraid I can't merge this PR. I have a feeling the implementation is doing more work than it needs to be, and the overall code quality isn't quite high enough.

The tests don't give me any confidence in the code. Test should be treated almost like documentation, with a function for each aspect of the code being tested. It's not enough just to get full converage.

If you need this, you might want to factor it out in to your own code. Keep it as an external function rather than a method in the core lib.

pom11:
the code is pretty much straight forward, covering all aspects from [docs](https://docs.python.org/3/library/string.html#format-string-syntax). at first glance the complexity comes from treating mixed arguments such as Text with string. i admit the missing comments from test could mislead what feature is testing.

do you plan to add this feature in the nearby future? would love to see a better approach

tried my best.

PR #2916, opened 2023-04-09
Fix auto detection of terminal size on Windows

lewis-yeung:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Currently, the automatic detection of terminal size on Windows is only performed for `_STDOUT_FILENO` which is the default value passed to `os.get_terminal_size`. So if we redirect the standard output (`stdout`) to a file and create a `Console` on the standard error (`stderr`) with `width` and `height` left `None`, the auto-detected terminal size will always be a fallback value `(80, 25)` (supposing that `legacy_windows` is `False`).

A simple example to illustrate this problem: I ran the following test script with only `stdout` redirected to a file (`stderr` was still connected to a terminal with width = 110 and height = 45):

```python
import sys

console = Console(stderr=True, legacy_windows=False)
print(console.size, file=sys.stderr)
```

and I got following output:

```
ConsoleDimensions(width=80, height=25)
```

which is not what I expected:

```
ConsoleDimensions(width=110, height=45)
```

On Linux, it works as expected.

To resolve the problem on Windows, we should perform the automatic detection for both `_STDOUT_FILENO` and `_STDERR_FILENO`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Fix auto detection of terminal size on Windows https://github.com/Textualize/rich/pull/2916
+
 ## [13.3.4] - 2023-04-12
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


lewis-yeung:
@willmcgugan Anything stopping us from continuing?

willmcgugan:
Closing, assumed stale. If it is still relevant, please open an issue with code to reproduce.

PR #2934, opened 2023-04-25
ensure threads are stopped in Progress().track

ThunderKey:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

If `Progress().track` is used with a generator and an exception is raised within this generator the `_TrackThread` is not stopped and the program does not finish. This change ensures that the `_TrackThread`s are stopped correctly.

A simple example would be:

```python

with Progress() as progress:
    tuple(i * object() for i in progress.track(range(10)))
```

This raises an error, since there is no `*` operator for `int` and `object`. The error is shown correctly, but the program does not complete, because a `_TrackThread` is still running.

The problem is, that generators are not aware of exceptions thrown. Therefore, `track` does not realize that the iteration was aborted and keeps the `_TrackThread` alive.

My workaround is to store all `_TrackThread` instances in a set and stop them in the `Progress().stop` method.

Unfortunately, this does not work for `rich.progress.track`, because this function is also not informed about exceptions. Therefore, the `with` statement is not aborted and continues to run.

I've prepared the changelog, but since I don't know your release schedule I've omitted the date of the release.


NicoWeio:
For reference: #3402

PR #3016, opened 2023-07-02
Added __lt__ function to Text class to enable sort functionality

sharkusk:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Basic change to enable sort functionality in Textual DataTable.  Copied structure of __contains__ function.  Textualize/textual/issues/2261

Please hold... I need to add the __gt__ function as well to cover both "str < Text" and "Text < str". -- Done.

### DIFF
--- a/rich/text.py
+++ b/rich/text.py
@@ -176,14 +176,14 @@ class Text(JupyterMixin):
             return self.plain < other
         elif isinstance(other, Text):
             return self.plain < other.plain
-        return False
+        return NotImplemented
 
     def __gt__(self, other: object) -> bool:
         if isinstance(other, str):
             return self.plain > other
         elif isinstance(other, Text):
             return self.plain > other.plain
-        return False
+        return NotImplemented
 
     def __contains__(self, other: object) -> bool:
         if isinstance(other, str):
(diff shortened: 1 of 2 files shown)


rich/text.py
```diff
@@ -171,6 +171,20 @@ def __eq__(self, other: object) -> bool:
             return NotImplemented
         return self.plain == other.plain and self._spans == other._spans
 
+    def __lt__(self, other: object) -> bool:
+        if isinstance(other, str):
+            return self.plain < other
+        elif isinstance(other, Text):
+            return self.plain < other.plain
+        return False
```
willmcgugan, line 179: This should return `NotImplemeted` rather than False.

willmcgugan (changes requested):

rich/text.py
sharkusk: Done.  I originally did this as a shortcut for comparing with None, but that is bad form.  I also removed the test cases accordingly.

willmcgugan:
Probably not needed now. The DataTable has grown more sorting functionality.

PR #3033, opened 2023-07-13
ExceptionGroup PEP654 proof of concept.

### LINKED ISSUE #1859: [REQUEST] Support for PEP 654 tracebacks (ExceptionGroups)
Hello,

let me provide some context first. [PEP 654](https://www.python.org/dev/peps/pep-0654/) introduces a new type of exception - an ExceptionGroup. (Also a BaseExceptionGroup, which is like the BaseException version of it). An ExceptionGroup is essentially a container for a list of inner exceptions. (It also introduces other things, like the except* clause, but I don't think that's relevant to Rich tracebacks.)

The hope is this ExceptionGroup will become a standard way for libraries to raise groups of exceptions. For example, this is a common use case for the Hypothesis library. There will be a backport package on PyPI to provide ExceptionGroups to older Python versions, so third party libraries will be able to use them without issue.

There is another PEP, [PEP 678](https://www.python.org/dev/peps/pep-0678/), that builds on ExceptionGroups to reify the concept of an exception note. This PEP has not yet been accepted, and is currently being discussed. (Due to a fluke, the CPython 3.11a4 release contains an implementation of it.) The idea of this PEP is for libraries to be able to attach a string to any exception under the `__note__` attribute, and for traceback machinery to simply print it out if it is present. The notes can and will be multiline in practice, I think.

So ExceptionGroups are an official thing, notes will maybe become a thing.

I'm working on a couple of libraries that can benefit from ExceptionGroups so I can give an example. Imagine you're structuring a JSON payload into a class, and there are errors. Setup code:

```python
from attrs import define
from cattrs import structure

@define
class Test:
    a: int
    b: int
    c: float


@define
class Outer:
    inner: Test
    c: int

structure({"inner": {"a": 'not_an_int', "b": 1}}, Outer)
```
An ExceptionGroup flies out, and the default traceback module in 3.11 renders it like this:
```python
  + Exception Group Traceback (most recent call last):
  |   File "/Users/tintvrtkovic/pg/cattrs/a01.py", line 33, in <module>
  |     structure({"inner": {"a": 'not_an_int', "b": 1}}, Outer)
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/Users/tintvrtkovic/pg/cattrs/src/cattr/converters.py", line 308, in structure
  |     return self._structure_func.dispatch(cl)(obj, cl)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "<cattrs generated structure __main__.Outer>", line 14, in structure_Outer
  | ExceptionGroup
  +-+---------------- 1 ----------------
    | Exception Group Traceback (most recent call last):
    |   File "<cattrs generated structure __main__.Outer>", line 5, in structure_Outer
    |   File "<cattrs generated structure __main__.Test>", line 19, in structure_Test
    | ExceptionGroup
    | Structuring class attribute Outer.inner
    +-+---------------- 1 ----------------
      | Traceback (most recent call last):
      |   File "<cattrs generated structure __main__.Test>", line 5, in structure_Test
      | ValueError: invalid literal for int() with base 10: 'not_an_int'
      | Structuring class attribute Test.a
      +---------------- 2 ----------------
      | Traceback (most recent call last):
      |   File "<cattrs generated structure __main__.Test>", line 15, in structure_Test
      | KeyError: 'c'
      | Structuring class attribute Test.c
      +------------------------------------
    +---------------- 2 ----------------
    | Traceback (most recent call last):
    |   File "<cattrs generated structure __main__.Outer>", line 10, in structure_Outer
    | KeyError: 'c'
    | Structuring class attribute Outer.c
    +------------------------------------
```

The `Structuring class attribute Test.a` strings are notes, added by cattrs.

It's a little messy, but there's a lot of information in there. The `Outer` class has errors for attributes `inner` (an ExceptionGroup) and `c` (a KeyError), and diving in, the Inner class (under `inner`) has errors for `a` (ValueError) and `c` (KeyError).

Anyway, this is how the traceback module prints this. I'm sure a Rich version would be much easier on the eyes ;)

AndreasBackx:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This is just a proof of concept for how exception groups may look, primarily looking to get some feedback on:
1. How people would like for it to look.
2. How the implementation should be done primarily, so feedback on the data structure changes and where to put the code.

This implementation does not (yet) care about backwards-compatibility or supportng the exceptiongroup pypi backport, that is to be added later. This is how the current example I've added looks when running, 

image

Compared to the default (which is also printed below it): 

image

This resolves #1859.

### DIFF
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -14,6 +14,7 @@ from typing import (
     Iterable,
     List,
     Optional,
+    Self,
     Sequence,
     Tuple,
     Type,
@@ -29,7 +30,14 @@ from pygments.util import ClassNotFound
 from . import pretty
 from ._loop import loop_last
 from .columns import Columns
-from .console import Console, ConsoleOptions, ConsoleRenderable, RenderResult, group
+from .console import (
+    Console,
+    ConsoleOptions,
+    ConsoleRenderable,
+    Group,
+    RenderResult,
+    group,
+)
 from .constrain import Constrain
 from .highlighter import RegexHighlighter, ReprHighlighter
 from .panel import Panel
@@ -197,6 +205,7 @@ class Stack:
     syntax_error: Optional[_SyntaxError] = None
     is_cause: bool = False
     frames: List[Frame] = field(default_factory=list)
+    grouped_traces: List["Trace"] = field(default_factory=list)
 
 
 @dataclass
@@ -417,6 +426,20 @@ class Traceback:
                     msg=exc_value.msg,
                 )
 
+
+            if isinstance(exc_value, BaseExceptionGroup):
+                for exception in exc_value.exceptions:
+                    stack.grouped_traces.append(cls.extract(
+                        exc_type=type(exception),
+                        exc_value=exception,
+                        traceback=exception.__traceback__,
+                        show_locals=show_locals,
+                        locals_max_length=locals_max_length,
+                        locals_max_string=locals_max_string,
+                        locals_hide_dunder=locals_hide_dunder,
+                        locals_hide_sunder=locals_hide_sunder,
+                    ))
+
             stacks.append(stack)
             append = stack.frames.append
 
@@ -484,9 +507,8 @@ class Traceback:
         trace = Trace(stacks=stacks)
         return trace
 
-    def __rich_console__(
-        self, console: Console, options: ConsoleOptions
-    ) -> RenderResult:
+
+    def _render_trace(self, console: Console, options: ConsoleOptions, trace: Trace, child_index: int | None = None) -> RenderResult:
         theme = self.theme
         background_style = theme.get_background_style()
         token_style = theme.get_style_for_token
@@ -514,11 +536,17 @@ class Traceback:
         )
 
         highlighter = ReprHighlighter()
-        for last, stack in loop_last(reversed(self.trace.stacks)):
+        for last, stack in loop_last(reversed(trace.stacks)):
+            is_exception_group = bool(stack.grouped_traces)
             if stack.frames:
+                traceback_title = "Traceback"
+                if is_exception_group:
+                    traceback_title = f"Exception Group {traceback_title}"
+                if child_index is not None:
+                    traceback_title += f" {child_index}"
                 stack_renderable: ConsoleRenderable = Panel(
                     self._render_stack(stack),
-                    title="[traceback.title]Traceback [dim](most recent call last)",
+                    title=f"[traceback.title]{traceback_title} [dim](most recent call last)",
                     style=background_style,
                     border_style="traceback.border",
                     expand=True,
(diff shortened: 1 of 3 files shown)


iamannalynleynes1206 (approved):

AndreasBackx:
@willmcgugan, when you have some time, could you share some of your critique so I can flesh this out and we can hopefully get this landed? Would like to get some early feedback before going the whole mile.

rich/traceback.py
```diff
@@ -417,6 +426,20 @@ def safe_str(_object: Any) -> str:
                     msg=exc_value.msg,
                 )
 
+
+            if isinstance(exc_value, BaseExceptionGroup):
```
Zac-HD, line 430: It would be very easy to make this support older Python versions, with or without the `exceptiongroup` backport: 
```suggestion
            if sys.version_info[:2] < (3, 11):
                BaseExceptionGroup = getattr(sys.modules.get("exceptiongroup"), "BaseExceptionGroup", ())
            if isinstance(exc_value, BaseExceptionGroup):
```

rich/traceback.py
```diff
@@ -544,6 +572,18 @@ def __rich_console__(
                     (f"{stack.exc_type}: ", "traceback.exc_type"),
                     highlighter(stack.syntax_error.msg),
                 )
+            elif stack.grouped_traces:
+                group_exception_renderable = Panel(
+                    Group(*[Group(*self._render_trace(console=console, options=options, trace=trace, child_index=index + 1)) for index, trace in enumerate(stack.grouped_traces)]),
+                    title=f"[exception_group.title]{stack.exc_type} [dim]{highlighter(stack.exc_value)}",
```
Zac-HD, line 578: I'd love to see support for PEP-678 `__notes__` here (and in the next `elif` clause, for non-grouped exceptions).  Probably easiest if we add a new method `.render_exc_value(stack)` cribbing from [this part of the backport](https://github.com/agronholm/exceptiongroup/blob/e0575e8800cade057b5d1541b782b15a4b8670dc/src/exceptiongroup/_formatting.py#L335-L338)?

Zac-HD (commented):
I'm really looking forward to having `ExceptionGroup` support in Rich!  The lack of support is increasingly painful in async code.

rich/traceback.py
AndreasBackx: Yes, this was the plan but hasn't been implemented because I wanted to get a PoC out first with some feedback. See my first post as well at the bottom.

rich/traceback.py
AndreasBackx: Also wanted to do this, but again wanted to get initial feedback first as this will depend on how the maintainer was possibly envisioning. If I have the go ahead for how it should be achieved, I can make the changes. Right now we're stuck waiting for feedback.

Zac-HD: My intuition as a maintainer is that fewer rounds of review makes everything easier - if I can just say "thanks, merging" that's likely to happen _sooner_ than if we iterate.  I understand not wanting to do more work which might not be merged though. 🙂

willmcgugan:
Just busy ATM. I will do a pass on Rich soon. Initial impressions are that it is a good idea.

willmcgugan:
Hi @AndreasBackx Sorry for the slow turnaround.

Yes, this is great in principle.

Re styling, I like the borders, but some don't and I expect they would complain. Maybe an indent would be better to show the exception groups as in the default? Worth experimenting.

I am going to do another pass on Rich next month. I could tackle this then. But happy for you to continue, if you still have the time?

AndreasBackx:
@willmcgugan if you're giving this a pass anyhow. Go ahead and take it over. That will be easier than communicating around this.

Regarding the borders, I would prefer a version where no frames/borders were used. Why? Because we're using Rich for logging in places where there is no set terminal width and so avoid the frames as they don't show up nicely when the output is saved and shown in a different location: someone else's terminal or the browser.

PR #3052, opened 2023-07-23
Case insensitive prompt choices

seapagan:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Optional flag for `Prompt.ask` allowing case-insensitive responses. Just add `case_sensitive=False` to the command options:

```python
from rich.prompt import Prompt
name = Prompt.ask("Enter your name", choices=["Paul", "Jessica", "Duncan"], default="Paul", case_sensitive=False)
```

Now, it would accept "paul" or "Paul" as valid responses. Defaults to True, which is the usual result.


iamannalynleynes1206 (approved):

darrenburns:
I think this is very useful, and might even be a better default!

The only thing I'd say is perhaps `case_sensitive` would be better than `case_insensitive` (we have a parameter in `Text.highlight_words` which is called `case_sensitive`, so it'd keep thing consistent)?

Any thoughts @willmcgugan?

seapagan:
> The only thing I'd say is perhaps `case_sensitive` would be better than `case_insensitive` (we have a parameter in `Text.highlight_words` which is called `case_sensitive`, so it'd keep thing consistent)?
> 

@darrenburns I can quickly rename to `case_sensitive` and flip the functionality if you want, I'm all for consistency :grin:. Can prob squeeze it in later tonight or early tomorrow morning. If you decide between you to make it default, that would be a quick extra change.

iamannalynleynes1206 (approved):

seapagan:
> I can quickly rename to `case_sensitive` and flip the functionality if you want, I'm all for consistency 😁. Can prob squeeze it in later tonight or early tomorrow morning. If you decide between you to make it default, that would be a quick extra change.

@darrenburns Done. I renamed the flag and left the default state as `case-sensitive = True`. This functionality actually makes more sense to me when you point it out.

seapagan:
@darrenburns, is this PR DOA, or will it be looked at in future? 

It could be a useful addition to the library as it's something I use a lot. I guess I can put it into a separate package if it won't be merged.

darrenburns:
That's up to @willmcgugan

willmcgugan:
Thanks for the PR.

PR #3056, opened 2023-07-26
Allow RichHandler to accept a 'style' attribute passed via LogRecord

koshell:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

I frequently use loggers with multiple handlers attached. It would be convenient to be able to pass style as an extra parameter. Currently if I want to use markup then other non-rich loggers will have that markup left in, however using style allows the RichHandler to print pretty text while keeping other handlers free of markup.


willmcgugan:
The Console object has a `style` argument which you can pass to the RichHandler.

koshell:
Please correct me if I'm wrong, but your suggesting I change the Console.style every time I want to style a log message then revert it to it's existing value afterwards? Would that not interfere with anything else using the console like progress bars or spinners and such?

PR #3067, opened 2023-07-30
Allow tilde in URL (fixes #3057)

### LINKED ISSUE #3057: [BUG] Tilde not included in link regex
The following doesn't highlight pas the `~`
```
>>> from rich import print
>>> print("http://127.0.0.1/~foo")
http://127.0.0.1/~foo
```

LukeSavefrogs:
This PR fixes #3057.

## Type of changes

- [X] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Fix tilde character (`~`) not included in link regex when printing to console https://github.com/Textualize/rich/issues/3057
 
 ## [13.5.0] - 2023-07-29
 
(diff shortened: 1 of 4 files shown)


willmcgugan:
Thanks. Can you add test(s)  please.

LukeSavefrogs:
> Thanks. Can you add test(s) please.

I can, do you think it's actually needed? It's literally one character in the regex and I don't see tests for other special characters (for example `%`, `$`, `+` and so on). 

Just asking anyway, it's not a problem adding an extra line in the tests 😄

willmcgugan:
You are probably correct, but a regression test is better than "probably".

LukeSavefrogs:
> You are probably correct, but a regression test is better than "probably".

I agree

If you're ok then I'll add the missing tests too, then (like `https://example.org/index.html?param1=value1`) 😀

willmcgugan:
Couldn’t hurt!

LukeSavefrogs:
Done 😀

willmcgugan (approved):
LGTM

LukeSavefrogs:
Forgot to run `black` to change formatting, so I did that just now. 😅

Anyway... Running `make format-check` shows that there are a lot of files (30) that should be formatted as well:

image

PR #3130, opened 2023-09-13
Fix markdown table rendering issue with inline styles/links

### LINKED ISSUE #3115: [BUG] Rich markdown printed to console only shows a single link in a column/ table cell

**Describe the bug**

I use Rich to format a pandas data frame with multiple links in some column to markdown and show it in the terminal. However, when there are more than a single link in a column rich only shows the last one?

```python

import pandas as pd

md_table = pd.DataFrame(
    {
        "links": ["[page1](https://www.example.com/page1), [page2](https://www.example.com/page2)"]
    }
).to_markdown()

console = Console()
console.print(Markdown(md_table))
```

Output:
```bash
      links  
 ━━━━━━━━━━━ 
  0   page2  
```

Expected
```
      links  
 ━━━━━━━━━━━ 
  0   page1, page2  
```

If I write the markdown to a file I get the expected table with multiple links...

**Platform**

platform="Darwin"
rich==13.5.2

</details>

rodrigogiraoserrao:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

The table data elements were not interacting well with inline styles (in the rich sense) and the table data element would have its content overridden every time we found another excerpt of text with a different style.
The fix was to do something similar to what `TextElement` does in its method `on_text`, which is to append the new content instead of replacing it.

This fixes #3115.

### DIFF
--- a/rich/markdown.py
+++ b/rich/markdown.py
@@ -314,7 +314,7 @@ class TableDataElement(MarkdownElement):
 
     @classmethod
     def create(cls, markdown: "Markdown", token: Token) -> "MarkdownElement":
-        style = str(token.attrs.get("style" "")) or ""
+        style = str(token.attrs.get("style")) or ""
 
         justify: JustifyMethod
         if "text-align:right" in style:
@@ -330,15 +330,13 @@ class TableDataElement(MarkdownElement):
         return cls(justify=justify)
 
     def __init__(self, justify: JustifyMethod) -> None:
-        self.content: TextType = ""
+        self.content: Text = Text("", justify=justify)
         self.justify = justify
 
     def on_text(self, context: "MarkdownContext", text: TextType) -> None:
-        plain = text.plain if isinstance(text, Text) else text
-        style = text.style if isinstance(text, Text) else ""
-        self.content = Text(
-            plain, justify=self.justify, style=context.style_stack.current
-        )
+        text = Text(text) if isinstance(text, str) else text
+        text.stylize(context.current_style)
+        self.content.append_text(text)
 
 
 class ListElement(MarkdownElement):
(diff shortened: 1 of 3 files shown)


rich/markdown.py
```diff
@@ -330,15 +330,13 @@ def create(cls, markdown: "Markdown", token: Token) -> "MarkdownElement":
...
 
     def on_text(self, context: "MarkdownContext", text: TextType) -> None:
-        plain = text.plain if isinstance(text, Text) else text
-        style = text.style if isinstance(text, Text) else ""
-        self.content = Text(
-            plain, justify=self.justify, style=context.style_stack.current
-        )
+        text = Text(text) if isinstance(text, str) else text
+        text.stylize(context.current_style)
+        self.content.append_text(text)
```
rodrigogiraoserrao, line 339: I was a bit confused here because `style` was being created but not used.
My change will preserve whatever styling is already in the `Text` instance but will then add the current context style on top of it.
This felt like the correct thing to do.

cwerner:
FWIW: I pip installed [this](https://github.com/Textualize/rich/tree/fix-table-inline-styles) into my original project where I spotted the issue that only a single/ the last link remained in a markdown table cell and I'm happy to report that this change fixes the problem for me  👍 🥳

rodrigogiraoserrao:
> FWIW: I pip installed [your pull request branch](https://github.com/Textualize/rich/tree/fix-table-inline-styles) into my original project where I spotted the issue that only a single/ the last link remained in a markdown table cell and I'm happy to report that this change fixes the problem for me 👍 🥳

That's good to know!
Thanks for the confirmation.

willmcgugan (approved):

Issue #2949, opened 2023-05-02
Issues with link IDs and style caching


rodrigogiraoserrao:
This is an account of many small related issues (not necessarily bugs, per se) uncovered while working on https://github.com/Textualize/textual/issues/1587

 - Hashing doesn't take into account the link ID, so styles that look the same but have different link IDs will clash.

 - `Style._link_id` is created inconsistently: the attribute `_link_id` is created in a couple of different places and is created inconsistently across those locations.

<details>
<summary> Some examples </summary>

https://github.com/Textualize/rich/blob/6d30ad0f30028210124c149811cbbe2b183711f9/rich/style.py#L192-L194

https://github.com/Textualize/rich/blob/6d30ad0f30028210124c149811cbbe2b183711f9/rich/style.py#L243

https://github.com/Textualize/rich/blob/6d30ad0f30028210124c149811cbbe2b183711f9/rich/style.py#L688

</details>

 - Style addition is cached in `Style._add` and the dunder method `__add__` only refreshes the link (by using `.copy` on the result) if the result has a `.link`, but it doesn't check for `.meta`.

 - The method `Text.render` adds another layer of caching that bypasses the link refresh because of hashing clashes (_maybe_ using the style indices instead of the styles themselves as cache keys fixes this).

I couldn't make much progress in solving these issues because I don't quite understand the _semantics_ of the attribute `.link_id`. Is it supposed to be a different ID for any `Style` instance that contains a link or some meta data?
Then, why not use the object ID itself?

```py
@property
def link_id(self):
    return str(id(self)) if self.link or self.meta else ""
```

willmcgugan:
The `_link_id` is a unique id used when generating terminal links. It's sole purpose is to let the terminal know what to highlight when you hover over it.

PR #3136, opened 2023-09-26
Style link id cache

rodrigogiraoserrao:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

The changes proposed here pass all tests.
No regression tests have been added yet.

## Description

This follows https://github.com/Textualize/textual/issues/1587 and https://github.com/Textualize/rich/issues/2949.
The issue we're dealing with boils down to the fact that the attribute `Style._link_id` is not taken into account when hashing a `Style` which then introduces caching collisions with styles that have the same `_link`/`_meta` attributes but that have a different `_link_id`.

 > Can't we include `_link_id` in the hash of a style?

We can't, because we don't want equality comparison to depend on `_link_id`, and if two objects are compared as equal, then [their hash should be the same](https://docs.python.org/3/reference/datamodel.html#object.__hash__).

When working on https://github.com/Textualize/textual/issues/1587, the first fix I found was that I could disable the cache inside `rich.text.Text.render` and the `functools.lru_cache` around `rich.style.Style._add` and that would fix the issue.
So, this does show it is a caching issue.
In particular, it has to do with the clashes that can exist when styles have the same meta values but originally have different `_link_id` attributes:

```py
from rich.style import Style
meta = {"click": "something"}
s1 = Style.from_meta(meta)
s2 = Style.from_meta(meta)
print(s1.link_id, s2.link_id)  # Different link_id.
# (
#     '6311506187629892456089155',
#     '4273756187629892456089155'
# )

base = Style(color="red")
print((base + s1).link_id)  # This is the link_id of s1.
# '6311506187629892456089155'

# Now, we add base to s2:
#             vv
print((base + s2).link_id)  # This is the link_id of s1!
# '6311506187629892456089155'
```

Now, we can take a look at how addition is implemented and this reveals a promising workaround:

```py
class Style:
    # ...
    def __add__(self, style: Optional["Style"]) -> "Style":
        combined_style = self._add(style)
        return combined_style.copy() if combined_style.link else combined_style
```

The call to `.copy()` refreshes the `link_id` if the resulting combined style has a link.
So, one might think it is a good idea to change the return statement to

```py
return combined_style.copy() if combined_style.link or combined_style.meta else combined_style
```

However, this breaks link highlighting on the Textual side entirely – as in, _no_ link gets _any_ highlighting – because Textual relies on the link IDs to know what to highlight.
When the markup is first parsed, Textual will keep tabs on the link IDs of the styles it created, so if the rendering of the widget will come back with a style with a new link ID, Textual won't know what to highlight.

A potential fix is to bypass the cache entirely when adding styles that have link IDs but no links.

### DIFF
--- a/rich/style.py
+++ b/rich/style.py
@@ -729,8 +729,7 @@ class Style:
         text = text or str(self)
         sys.stdout.write(f"{self.render(text)}\n")
 
-    @lru_cache(maxsize=1024)
-    def _add(self, style: Optional["Style"]) -> "Style":
+    def _add_no_cache(self, style: Optional["Style"]) -> "Style":
         if style is None or style._null:
             return self
         if self._null:
@@ -754,8 +753,20 @@ class Style:
         new_style._hash = None
         return new_style
 
+    @lru_cache(maxsize=1024)
+    def _add(self, style: Optional["Style"]) -> "Style":
+        return self._add_no_cache(style)
+
     def __add__(self, style: Optional["Style"]) -> "Style":
-        combined_style = self._add(style)
+        if (
+            style is not None
+            and not style._link
+            and not self._link
+            and (self._link_id or style._link_id)
+        ):
+            combined_style = self._add_no_cache(style)
+        else:
+            combined_style = self._add(style)
         return combined_style.copy() if combined_style.link else combined_style
 
 
--- a/rich/text.py
+++ b/rich/text.py
@@ -746,18 +746,19 @@ class Text(JupyterMixin):
         stack_append = stack.append
         stack_pop = stack.remove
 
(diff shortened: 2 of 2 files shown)


rodrigogiraoserrao:
@willmcgugan if I recall correctly, when I tackled this we talked a bit and concluded this wasn't trivial. This PR was a draft because it was a proposed solution.
I'll be happy to mark it as ready if you prefer that to review it.

willmcgugan:
If you think it is a possible solution, please mark it as ready. We'll go over it at some point. Need to refresh my memory.

willmcgugan:
I don't quite grok this enough to merge. Closing for now, will look in to it again when I tackle the original issue.

PR #3166, opened 2023-10-24
Enable per-column enabling or disabling of highlighting

JulesGM:
Defaults to the value of the self.table when the default value of None is unchanged.

It seems like the Column constructor didn't receive a `highlight` value, so I guess that it used to receive a highlight value at a later point. Someone who knows the project better could comment on that.

## Type of changes

-> New feature

## Checklist

This is more of a proof of concept than an actual pull request.


willmcgugan:
Thanks

PR #3222, opened 2023-12-03
Mxp support

volundmush:
## Type of changes

- [ ] Bug fix
- [X ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description
This is a second attempt to add official MXP support to Rich to make it more suitable for use with MUDs as a side-use.

Rich is the absolute best Python ANSI library, and, thus far, the only one I've been able to get to work with my Python MU* projects. So, it would be amazing to me if we could get official support for MXP in.

MXP, or Mud eXtension Protocol, is a feature used in MUDs which turns the outgoing ANSI text into a XML stream (although MXP elements aren't always proper XML, all other rules of XML/HTML escaping apply). Its specification can be found here:
https://www.zuggsoft.com/zmud/mxp.htm

The current way I've figured out to make this work is to add an `_mxp` property on Styles which can take a `rich.style.MXP` object. Unfortunately, I have no idea how this could ever be accessible to bbcode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [13.7.1] - 2023-12-3
+
+### Added
+
+- Preliminary MXP support to Styles and Console.
+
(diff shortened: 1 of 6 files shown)


iamannalynleynes1206 (approved):

iamannalynleynes1206 (approved):

willmcgugan:
Sorry, its just too niche to warrant going in the core library.

PR #3236, opened 2023-12-22
Make the Console generic on the file type

NeilGirdhar:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

This would have saved me from having to assert on the type here: https://github.com/NeilGirdhar/tjax/commit/085a9e4ffb989b80ab236609f73e2599856138e5#diff-8d12e9756f49b99f6d43b0863bf279057e145f38b98db4c30522b1b14138bd5c

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -29,6 +29,7 @@ from typing import (
     TextIO,
     Tuple,
     Type,
+    TypeVar,
     Union,
     cast,
 )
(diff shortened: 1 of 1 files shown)


willmcgugan:
I suspect this will cause typing errors for anyone who doesn't provide the generic type. And that would lead to a flood of issues.

The solution is good, but for this reason I don't think I can merge.

NeilGirdhar:
> I suspect this will cause typing errors for anyone who doesn't provide the generic type.

It never creates type errors.  It creates warnings if the option to warn is enabled for your type checker.

And the motivation for this is actually to prevent type errors that are otherwise unavoidable.

willmcgugan:
Errors or warnings, I'll still get issues. I think an isinstance check is the right thing to do here. Or you could use the capture API.

NeilGirdhar:
Okay, hopefully you switch to type defaults when 3.10 is dropped then :smile:

PR #3247, opened 2024-01-05
Added customization to the  illegal_choice_message. 

SamuelMorrisProjects:
## Type of changes

- [ ] Bug fix
- [ x] New feature
- [ x] Documentation / docstrings
- [ ] Tests
- [ ] Other

## Checklist

- [ x] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate.
- [ x] I accept that @willmcgugan may be pedantic in the code review.

## Description
I made the illegal_choice_message an optional parameter of the both the class prompt and the method ask in prompt.py. The illegal choice parameter automatically defaults to the original Text Type("[prompt.invalid.choice]Please select one of the available options."). This will allow those using the library to customize the message that is outputted to the user. I also updated the documentation to reflect this change. 
## Why 
I added this feature because to my knowledge—please correct me if I am wrong in this assumption—there is no way to change the error message outputted when a user does not pick one of the choices. I am somewhat new to programming so I could have missed something. I also feel this change will produce overall more robust code.
## Examples
**Code:**
image
**Output:**
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Adds missing parameters to Panel.fit https://github.com/Textualize/rich/issues/3142
+- Added a addtional parameter to prompt class (illegal_choice_message).
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


willmcgugan:
The recommended way to do this would be to subclass your Prompt class.

PR #3258, opened 2024-01-10
Wrapping changes

darrenburns:
Updates divide_line to be configurable in how it handles whitespace.

This is required for the TextArea in Textual to support wrapping. The prior behaviour of Rich which consumed consecutive whitespaces when a line wrapped felt off (even though it matches the browsers behaviour for text areas).


darrenburns:
Closed because this logic will now live in Textual.

PR #3270, opened 2024-02-02
use larger width/height defaults for non-interactive consoles

ttarabula:
When a user is redirecting output to a file, or perhaps running a job non-interactively (say, via cron), arguably, it is surprising when output (such as a rich.table) is truncated to a default size of (80, 25). This change preserves that default Console.size() -> width=80, height=25, but uses larger defaults when Console.is_interactive == False.

## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

I am not at all sure that I'm solving the right problem here or going about it correctly, but I prefer describing what I'm trying to accomplish, via code (pull request) rather than simply descriptively, via a project issue. The basic idea here is that I _personally_ find the following result surprising:

given the following table test (basically pulled directly from the example in the documentation, here: https://rich.readthedocs.io/en/stable/tables.html#tables):

```
#!/usr/bin/env python
from rich.table import Table

table = Table(title="Star Wars Movies")

table.add_column("Released", justify="right", style="cyan", no_wrap=True)
table.add_column("Title", style="magenta")
table.add_column("Box Office", justify="right", style="green")

table.add_row("Dec 20, 2019", "Star Wars: The Rise of Skywalker", "$952,110,690")
table.add_row("May 25, 2018", "Solo: A Star Wars Story", "$393,151,347")
table.add_row("Dec 15, 2017", "Star Wars Ep. V111: The Last Jedi", "$1,332,539,889")
table.add_row("Dec 16, 2016", "Rogue One: A Star Wars Story", "$1,332,439,889")
table.add_row("May 22, 2026", "As Yet Unnamed Amazing Next Star Wars Movie With A Really Long Name", "$999,999,999,999")

console = Console()
console.print(table)
```

I like that it truncates the table to my terminal width/height when running it interactively (say, `./print_rich_table.py`), but would prefer that it doesn't truncate and/or wrap when non-interactive, for instance, when redirecting to a file (`./print_rich_table.py > star_wars.txt`), or perhaps running a job via cron or in some other way where the expectation is for the handling of display constraints to be handled by some later user who may not be the original script runner.

I've tested this change by shrinking/growing my terminal and making sure the old behavior is preserved, as well as by updating tests/test_console.py with some new cases.

I'm happy to try to take another approach here if this doesn't make sense. Also, of course, I picked the new defaults out of thin air (just trying to come up with some reasonably large numbers to prove out the strategy).

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -76,6 +76,10 @@ if TYPE_CHECKING:
 
 JUPYTER_DEFAULT_COLUMNS = 115
 JUPYTER_DEFAULT_LINES = 100
+INTERACTIVE_DEFAULT_COLUMNS = 80
+INTERACTIVE_DEFAULT_LINES = 25
+NON_INTERACTIVE_DEFAULT_COLUMNS = 32767
+NON_INTERACTIVE_DEFAULT_LINES = 32767
 WINDOWS = platform.system() == "Windows"
 
 HighlighterType = Callable[[Union[str, "Text"]], "Text"]
@@ -1001,7 +1005,9 @@ class Console:
             return ConsoleDimensions(self._width - self.legacy_windows, self._height)
 
         if self.is_dumb_terminal:
-            return ConsoleDimensions(80, 25)
+            return ConsoleDimensions(
+                INTERACTIVE_DEFAULT_COLUMNS, INTERACTIVE_DEFAULT_LINES
+            )
 
         width: Optional[int] = None
         height: Optional[int] = None
@@ -1028,8 +1034,13 @@ class Console:
             height = int(lines)
 
         # get_terminal_size can report 0, 0 if run from pseudo-terminal
-        width = width or 80
-        height = height or 25
+        if self.is_interactive:
+            width = width or INTERACTIVE_DEFAULT_COLUMNS
+            height = height or INTERACTIVE_DEFAULT_LINES
+        else:
+            width = NON_INTERACTIVE_DEFAULT_COLUMNS
+            height = NON_INTERACTIVE_DEFAULT_LINES
+
         return ConsoleDimensions(
             width - self.legacy_windows if self._width is None else self._width,
             height if self._height is None else self._height,
--- a/tests/test_console.py
+++ b/tests/test_console.py
@@ -18,6 +18,10 @@ from rich.console import (
     ConsoleOptions,
     ScreenUpdate,
     group,
+    INTERACTIVE_DEFAULT_COLUMNS,
+    INTERACTIVE_DEFAULT_LINES,
+    NON_INTERACTIVE_DEFAULT_COLUMNS,
+    NON_INTERACTIVE_DEFAULT_LINES,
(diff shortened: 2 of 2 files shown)


ttarabula:
> This is a bad idea. You wouldn't want a rule or table to be 32767 characters long.

Ah, I fully concede that this _might_ be bad to handle it _this way,_ but let me try to explain my _actual_ end goal here a little bit more clearly, in case there's another way towards what I'm trying to achieve...

In our environment, we use rich.table (which we love!) in various ways. It is _incredibly_ helpful for rich to fit output tables to the width of a terminal when a user is doing ad-hoc interactive work. However, we have many use cases where the output of a job that has rich.table components, goes to a text file for later analysis/viewing. We typically _don't_ want any sort of truncation of that data in those scenarios. That is, the user would happily resize their terminal and/or scroll horizontally to look through the table. They generally want to see all of the field values, without any truncation. Consider especially the use case where we can't predict in advance what the min/max column/width boundaries are expected to be, and perhaps even won't be looking at the output until weeks after it was saved. Of course, one could argue that rich.table isn't the right tool for this job, and outputting as some sort of standard table instead (such as csv), but there are many use cases where it's just simpler or more user friendly for there to be a expectation of visual consistency (ie, the user looks through these sorts of tables all the time, and is more efficiently able to glance at it when it has the same look and feel as what they see interactively).

My assertion is really that I don't believe that console dimensions or a default (such as 80x25) should be used _at all_ in those cases (ie, my large number was "pseudo-infinite" for practical purposes).

We can _certainly_ work around this, but it is _relatively_ burdensome to the developer to manage this on a per-script basis, rather than it being a core behavior of rich, or at least modifiable at a high level (perhaps a flag that gets passed to the rich.console.Console() constructor, optionally, that maybe defaults to the current behavior, and changes the behavior if specified).

willmcgugan:
You can set the env var COLUMNS to your desired width when running non-interactively. Would that work for you?

ttarabula:
Yeah, we have various viable workarounds such as setting $COLUMNS, or modifying the behaviour of individual scripts (_typically_ by branching on `console.is_interactive`, setting `console.width` and `console.height` to something different - very similar to the strategy demonstrated in the PR). This proposal was really about _suggesting_ that it might make sense to be a core feature of rich (in some way), but I was **_fully_** anticipating the possibility that it wouldn't be desirable. That's fine with me! It's food for thought more than anything, and in case other users encounter similar pain points surrounding this topic, they can use it as reference.

allentc:
Guess I happened into this the day after it has been closed.  My use case is exactly as @ttarabula described: rendering tables to stdout in a non-interactive manner (a Chronical scheduler job.)  The renderer is defaulting to 80x25, badly truncating the table values and width.

What I would hope for: that rich.table have some means of being configured in a "do not truncate values" and "do not limit table width" mode, yet still narrow the columns to the widest values.

PR #3276, opened 2024-02-08
Fix `leading` Parameter Type In Docstring

MichaelYochpaz:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Noticed on [the docs](https://rich.readthedocs.io/en/stable/reference/table.html) that the `leading` parameter (available for the `Table` and `Column` objects), mention that this parameter is a `bool` type, although it appears to be an `int` type and has a default value of `0`.

This PR fixes the docstrings to annotate this parameter with the correct `int` type.


willmcgugan (approved):

willmcgugan:
thanks

PR #3278, opened 2024-02-13
Strip problematic private escape sequences

dosisod:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

["Private escape sequences"](https://en.wikipedia.org/wiki/ANSI_escape_code#Fp_Escape_sequences) are escape sequences that are reserved for private use, though `\x1b7` and `\x1b8` are commonly used for storing and restoring the current cursor position. When those escape codes are not stripped the cursor jumps around and causes Rich to write garbage output. An example of a program that uses this cursor store/restore functionality is the APK package manager in Alpine Linux:

https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/48d91f482eb48a0a107b714ee183bb7e07782e14/src/print.c#L232-240

This commit updates the ANSI parser to ignore the `\x1b0`-`\x1b?` escape sequences, thus preventing them from being printed and causing havoc.


willmcgugan (approved):

willmcgugan:
Thanks

PR #3294, opened 2024-02-29
Add the ability to disable the display of object attributes, or to only display selected attributes in the inspect function.

mateusz-wozny:
## Type of changes

- [x] New feature
- [x] Tests

## Checklist

## Description
I want to use inspect method for object which have some attributes very large (for example list of 1000 items) and currently all attributes are displayed. The output is quite unreadable so I want to disable displaying object attributes or display only selected. I added parameter to enable/disable displaying object attributes and parameter to pass names of attributes I want to display (if not specified all attributes will be displayed).


willmcgugan:
What has that many attributes?

willmcgugan:
Closing, because I doubt this is a concern for anyone else. Willing to be convinced.

PR #3300, opened 2024-03-07
Bisect all the things!

### LINKED ISSUE #3298: Look in to using bisect module to replace home grown binary searches.

### LINKED ISSUE #3299: [BUG] Segment._split_cells doesn't handle non-unit characters well

```py

s = Segment("🦊🦊🦊abcdef")
```

rodrigogiraoserrao:
This PR replaces a couple of searches with the module `bisect`.

The screenshot below shows Rich's benchmarks ran, on my machine, on v13.7.1 and on my changes.
There were clear wins at the top from the optimisations in `cells.py`.

I boxed a region of things that stayed the same but whose underlying code was changed to fix the bug reported in #3299.

I also cut the code in `Text.divide` in half by using `bisect`.
The `Text.divide` benchmarks show no change but that seems to have a positive impact in `Text.split`. (8% in this screenshot, 5% - 10% from the multiple times I ran the benchmarks while testing changes).

Screenshot 2024-03-07 at 17 35 09

I don't know if I'm supposed to commit and upload the benchmark results, so I didn't for now.
I also don't know if “optimised some code” should go on the changelog so I didn't put it there for now.

Fixes #3298
Fixes #3299


rodrigogiraoserrao:
I also tried optimising `Segment.divide`.
Tried two slightly different implementations but the benchmarks showed that I was actually slowing down the function.
I couldn't bring myself to completely throw away the code so I included them in the commit message of commit 7912306.

willmcgugan (commented):
I gather the `_cell_widths.py` change is for bisect. Could we keep the pairs and use `key=itemgetter(0)` with bisect?

willmcgugan:
Just discovered that bisect key was added in 3.10. 

However, I don't think we will need it. It will work with the tuples--no need to separate them.

rodrigogiraoserrao:
@willmcgugan do you need anything else from me on this PR?

willmcgugan:
No, its good. Just need to find the time to review.

willmcgugan:
Going to break this out in to smaller changes.

PR #3311, opened 2024-03-26
live: add "tail" vertical_overflow method

bretello:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

This implements an alternative `vertical_overflow="tail"` mode for `Live`, behaving somehow like "crop" but only showing the trailing lines that fit the terminal height, adding an ellipsis at the top

[Here](https://asciinema.org/a/vkbg83LQ6u1PogViACbWodWpQ) is an example:

[vertical_overflow_tail](https://asciinema.org/a/vkbg83LQ6u1PogViACbWodWpQ)

### DIFF
--- a/docs/source/live.rst
+++ b/docs/source/live.rst
@@ -3,7 +3,7 @@
 Live Display
 ============
 
-Progress bars and status indicators use a *live* display to animate parts of the terminal. You can build custom live displays with the :class:`~rich.live.Live` class. 
+Progress bars and status indicators use a *live* display to animate parts of the terminal. You can build custom live displays with the :class:`~rich.live.Live` class.
 
 For a demonstration of a live display, run the following command::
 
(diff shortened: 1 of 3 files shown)


bretello:
Here's the code used to generate the above asciicast:

```python
from time import sleep
from rich.live import Live

console = Console(
    height=5,
    force_terminal=True,
    legacy_windows=False,
    color_system=None,
    _environ={},
)
console.print("This is a console with height=5")

with Live(
    console=console,
    auto_refresh=False,
    vertical_overflow="tail",
) as live:
    display = ""
    for step in range(10):
        display += f"Step {step}\n"
        sleep(1)
        live.update(display, refresh=True)
```

willmcgugan:
Im afraid I don't think this is a good idea. Consider Textual for such functionality.

PR #3325, opened 2024-04-07
Enable rendering of rich exception objects in traceback

pradyunsg:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

See https://github.com/Textualize/rich/discussions/2717 for discussion. This enables exceptions to be rich objects and be rendered via the rich rendering pipeline, in tracebacks. This can be tested manually by running `python -m rich.traceback` whose example now uses a rich object instead of the `NameError`.

### DIFF
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -28,6 +28,7 @@ from pygments.util import ClassNotFound
 
 from . import pretty
 from ._loop import loop_last
+from .abc import RichRenderable
 from .columns import Columns
 from .console import Console, ConsoleOptions, ConsoleRenderable, RenderResult, group
 from .constrain import Constrain
@@ -194,6 +195,7 @@ class _SyntaxError:
 class Stack:
     exc_type: str
     exc_value: str
+    exc_renderable: Optional[RichRenderable] = None
     syntax_error: Optional[_SyntaxError] = None
     is_cause: bool = False
     frames: List[Frame] = field(default_factory=list)
@@ -416,6 +418,8 @@ class Traceback:
                     line=exc_value.text or "",
                     msg=exc_value.msg,
                 )
+            if isinstance(exc_value, RichRenderable):
+                stack.exc_renderable = exc_value
 
             stacks.append(stack)
             append = stack.frames.append
@@ -544,6 +548,19 @@ class Traceback:
                     (f"{stack.exc_type}: ", "traceback.exc_type"),
                     highlighter(stack.syntax_error.msg),
                 )
+            elif stack.exc_renderable:
+                exc_renderable = Constrain(stack.exc_renderable, self.width)
+                with console.use_theme(traceback_theme):
+                    try:
+                        segments = tuple(console.render(renderable=exc_renderable))
+                    except Exception:
+                        yield Text("<exception rich render failed>")
+                        yield Text.assemble(
+                            (f"{stack.exc_type}: ", "traceback.exc_type"),
+                            highlighter(stack.exc_value),
+                        )
+                    else:
+                        yield from segments
             elif stack.exc_value:
                 yield Text.assemble(
                     (f"{stack.exc_type}: ", "traceback.exc_type"),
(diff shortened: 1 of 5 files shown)


pradyunsg:
This is currently a draft since I'm not sure how exactly exceptions within the rich object rendering should be handled -- I'm leaning toward trying to catch them and presenting a fallback like "render failed" or similar. Not quite sure how to do that, so figured I'd open this as a draft PR in case y'all have any opinions/suggestions on how to achieve that.

pradyunsg:
OK, I think I figured out something that works! 🎉 

Using the `Console` object to eagerly render the exception and iterating through to collect all the segments, before yielding those segments forward. This enables this logic to catch any rendering-related issues inline, rather than needing to handle them at some other level (eg: `console.render`).

Marking this as ready for review, since I think everything that is necessary to land this is in place now[^1] -- code, tests, documentation, changelog, contributing.md. :)

[^1]: Assuming of course that there are no concerns with the implementation/approach. 😉

rich/traceback.py
```diff
@@ -722,7 +739,11 @@ def render_locals(frame: Frame) -> Iterable[ConsoleRenderable]:
     from .console import Console
 
     console = Console()
-    import sys
```
pradyunsg, line 725: sys was unused at this point.

pradyunsg:
Oh, and here's an SVG showing that this works!

exception

tests/test_traceback.py
```diff
@@ -263,6 +284,36 @@ def __str__(self):
...
+        def __rich_console__(self, console, options):
+            raise Exception("broken")
+
+    console = Console(width=100, file=io.StringIO())
+    try:
+        raise BrokenRichError()
+    except Exception:
+        console.print_exception()
+    result = console.file.getvalue()
+    print(result)
```
pradyunsg, line 298: These prints are useful for diagnosing failures, and pytest will capture this output appropriately. :)

rich/traceback.py
```diff
@@ -416,6 +418,8 @@ def safe_str(_object: Any) -> str:
                     line=exc_value.text or "",
                     msg=exc_value.msg,
                 )
+            if isinstance(exc_value, RichRenderable):
```
willmcgugan, line 421: My concern here is that if we offer arbitrary renderables, then the Traceback object may no longer be trivially sererializeable. Which was one of the goals for this class.

And since a renderable may be mutable, it may become out of date if it isn't immediately printed.

willmcgugan (commented):
I can see why this would be nice, but it breaks some assumptions. See comment...

willmcgugan:
@pradyunsg Any thoughts on above? The issue with serialization may be a deal breaker.

pradyunsg:
Thanks for the mention -- I'd missed the original review.

> My concern here is that if we offer arbitrary renderables, then the Traceback object may no longer be trivially sererializeable. 

(I assume by serialization you mean [object] -> [rich-rendered string] and not some on-disk/pickle representation or something)

I share that, which is why there's a fallback to the string-based approach in the render side of this if the serialization via the rich protocol fails. I think it's an acceptable tradeoff since this is opt-in on the exception raiser's end since this is an additional set of methods on exception class itself.

I'd be fine with adding another layer of opt-in via plumbing through a boolean to opt-in to rendering such exceptions from the caller's end as well, if you think it's important enough to have that for some reason (eg: there are usecases for not wanting to do the render in some cases other than the ones covered by the fallback logic)

> And since a renderable may be mutable, it may become out of date if it isn't immediately printed.

That's also the case with exception objects themselves and wouldn't be unique to rich's renderables (`__str__` and other methods on an exception object can mutate itself in a similar manner as `__rich__` can).

```
❯ cat /tmp/foo.py           
try:
    raise Exception("This is an error message")
except Exception as e:
    print(e)
    e.args = ("This is a new error message",)
    raise

❯ python /tmp/foo.py
This is an error message
  File "/tmp/foo.py", line 2, in <module>
    raise Exception("This is an error message")
Exception: This is a new error message
```

Issue #3335, opened 2024-04-17
Margins in text lines in Jupyter HTML template


vladsavelyev:
Hi!

We are using Rich for logging (`from rich.logging import RichHandler`), and I have a few questions about its text formatting in Jupyter notebooks. First, it seems to be adding a large spacing between lines.

```py
import rich
import rich.progress
from rich.logging import RichHandler
import logging

log = logging.getLogger("rich")

console = rich.console.Console()
handler = RichHandler(
    console=console,
    show_time=False,
    show_level=False,
)
log.handlers = [handler]

log.info("|           multiqc | Loading data from /Users/vlad/git/MultiQC/multiqc_data/multiqc_data.json")
log.info("|           multiqc | Loaded module Salmon")
log.info("|           multiqc | Loaded plot salmon_plot")
log.info("|           multiqc | Loaded plot general_stats_table")

with rich.progress.Progress(console=console) as progress:
    task = progress.add_task("| searching", total=100)
    for i in range(100):
        progress.update(task, advance=1)
    progress.update(task)

log.info("|           multiqc | Loading data from /Users/vlad/git/MultiQC/multiqc_data/multiqc_data.json")
log.info("|           multiqc | Loaded module Salmon")
log.info("|           multiqc | Loaded plot salmon_plot")
log.info("|           multiqc | Loaded plot general_stats_table")
```

Notebook output:

![CleanShot 2024-04-17 at 14 13 35@2x](https://github.com/Textualize/rich/assets/1575412/748b7105-7921-4338-a379-4732b1d1cae5)

With especially large spacing after the progress bar widget.

Compared that to a standard text output:

```py
import logging
import sys
import tqdm

log = logging.getLogger("rich")
log.handlers = [logging.StreamHandler(sys.stdout)]
log.setLevel("INFO")

log.info("|           multiqc | Loading data from /Users/vlad/git/MultiQC/multiqc_data/multiqc_data.json")
log.info("|           multiqc | Loaded module Salmon")
log.info("|           multiqc | Loaded plot salmon_plot")
log.info("|           multiqc | Loaded plot general_stats_table")
with tqdm.tqdm(
    total=100,
    file=sys.stdout,
    bar_format=f"| {'searching':>17} | " + "{bar:50} {percentage:3.0f}% {n_fmt}/{total_fmt} {desc}",
) as pbar:
    for i in range(100):
        pbar.update(1)
    pbar.refresh()
[...]

PR #3336, opened 2024-04-17
Remove margin from text lines in Jupyter HTML template

vladsavelyev:
Hard-code the `pre` margin, see context in https://github.com/Textualize/rich/issues/3335

### DIFF
--- a/rich/jupyter.py
+++ b/rich/jupyter.py
@@ -11,7 +11,7 @@ if TYPE_CHECKING:
     from rich.console import ConsoleRenderable
 
 JUPYTER_HTML_FORMAT = """\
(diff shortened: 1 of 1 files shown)


willmcgugan:
Every change to Jupyter support causes someone else to complain.

Closing, but happy to reconsider if you can convince me this doesn't break on any of the Jupyter platforms (there are many).

vladsavelyev:
Should I close https://github.com/Textualize/rich/issues/3335 as well?

willmcgugan:
I've practically given up on Jupyter formatting issues. The issue is reasonable, but I can't be certain any solution wont break some Jupyter platform somewhere. Happy to leave the issue, but any fix would have to be tested with a good number of 
Jupter platforms.

PR #3347, opened 2024-04-27
highlight_regex in rich.text.Text Now Expects a Compiled Regular Expression (re.compile) Has Been Passed

PyWoody:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [X] Other

## Checklist

## Description

`highlight_regex` in `rich.text.Text` now expects the passed `re_highlight` to be a compiled regular expression. If a `str` is passed, as was the previous expectation, the `str` will be compiled as a regular expression and the method will continue on as before.

The proposed change is for a minor performance bump in the `rich.highlighter.RegexHighlighter` class. A user can now pass compiled regular expressions in the the list of `highlights`, which will be guaranteed to not re-compiled by `Text`. Previously, `highlight_regex` expected it was receiving a `str` so it compiled the regular expression every time.

There will be no breaking changes to the user as the user can continue to pass `str` in their respective `RegexHighlighter.highlights`. The only performance degradation will be caused by the `isinstance(re_highlight, str)` check, which will be minimal.

Below is a silly speed test (hopefully) showing the new `highlight_regex` as performing faster. I used 600 "words" for the regular expressions because Python's current `re.` actions use a shared internal cache of a maximum of 512[0] regular expressions and I wanted demonstrate how relying on that can cause slowdowns.

[0] https://github.com/python/cpython/blob/3.12/Lib/re/__init__.py#L277

```python3
import itertools
import re
import time

from rich.highlighter import RegexHighlighter
from rich.theme import Theme


PHRASE = ' '.join(f'word{i}' for i in range(600))


class SillyCompiledHighlighter(RegexHighlighter):

    base_style = 'example.'
    highlights = [re.compile(rf'(?P<{i}>{i})') for i in PHRASE.split()]


class SillyStringHighlighter(RegexHighlighter):

    base_style = 'example.'
    highlights = [rf'(?P<{i}>{i})' for i in PHRASE.split()]


def run():
    colors = itertools.cycle(range(256))
    theme = Theme(
        {f'example.{v}': f'color({i})' for i, v in zip(colors, PHRASE.split())}
    )
    console = Console(
        highlighter=SillyCompiledHighlighter(), theme=theme, quiet=True
    )
    start = time.time()
    for _ in range(1_000):
        console.print(PHRASE)
    print('Compiled:', time.time() - start)
    console = Console(
        highlighter=SillyStringHighlighter(), theme=theme, quiet=True
    )
    start = time.time()
    for _ in range(1_000):
        console.print(PHRASE)
    print('String:', time.time() - start)


if __name__ == '__main__':
    run()
```


willmcgugan:
Changelog and tests please.

willmcgugan:
@PyWoody Are you still working on this?

PyWoody:
Sorry, in haste I pushed from a location that didn't have the pre-commit hooks installed. I've installed the hooks, updated the test file according to `black`, and have pushed a new commit. The pre-commit hooks passed on my local. Sorry about that.

rich/text.py
```diff
@@ -600,7 +600,7 @@ def highlight_regex(
         translated to styles.
 
         Args:
-            re_highlight (str): A regular expression.
+            re_highlight (Union[re.Pattern, str]): A regular expression.
```
willmcgugan, line 603:
```suggestion
            re_highlight (Union[re.Pattern, str]): A regular expression object or string.
```

willmcgugan (approved):
Thanks

PR #3350, opened 2024-05-01
Fix warning in test suite

ofek:
image

### DIFF
--- a/tests/test_syntax.py
+++ b/tests/test_syntax.py
@@ -3,7 +3,6 @@ import os
 import sys
 import tempfile
 
-import pkg_resources
 import pytest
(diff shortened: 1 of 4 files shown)


tests/test_syntax.py
```diff
@@ -21,7 +20,12 @@
 
 from .render import render
 
-PYGMENTS_VERSION = pkg_resources.get_distribution("pygments").version
+if sys.version_info >= (3, 8):
```
willmcgugan, line 23: This doesn't look right!

willmcgugan: Scrap that. I didn't notice the `.` and `_`

willmcgugan (approved):

PR #3359, opened 2024-05-14
Adds a `name` attribute to `Spinner`

KaiSforza:
Makes it nicer for validation and use of the Spinner object in later functions/classes.

## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [ ] I've added tests for new code. (not really much need, let me know if I should add an assert or something though.)

## Description

Simply adds a `name` attribute to `Spinner` for use in later functions/classes. The current one has no mention of `name` outside of `__init__`, but a lot of the functions and classes take a spinner name, not a spinner object. I added this because of some of the stuff I did in pypa/hatch#1490. Would have been nice to make it not a throwaway.


willmcgugan:
I don't quite follow why you need this. Can you elaborate?

Issue #2286, opened 2022-05-20
[REQUEST] PEP657 error code ranges in tracebacks


tonybaloney:
Python 3.11 implements code position ranges in tracebacks. 

```python
from rich.traceback import install
install(show_locals=True)
data = [
    {
        "person": {
            "friends": [
                {"name": "Sally",
                "age": 32},
                {"name": "Yuri",
                "age": 43}
            ],
            "location": "Warsaw, Poland"
        }
    }
]

print(data[0]["person"]["friends"][2]["name"])
```

The default exception handler will print the position of the offending error in the traceback:

```python
Traceback (most recent call last):
  File "/Users/anthonyshaw/projects/python-3.11-demos/error_messages.py", line 15, in <module>
    print(data[0]["person"]["friends"][2]["name"])
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

Rich does not use this data (but still looks very pretty)
<img width="1066" alt="screenshot 2022-05-20 at 14 46 09" src="https://user-images.githubusercontent.com/1532417/169452386-5aecf2c0-7342-4e4c-bd26-67b2a75f2862.png">

IT would be great to get the trace messages with arrows.

tonybaloney:
Note: TracebackException now has the `offset` and `end_offset` attributes which expose the information needed to print the arrows.

https://github.com/python/cpython/blob/3.11/Lib/traceback.py#L852-L864

pradyunsg:
This is very much a drive-by comment: I wonder if there's Unicode characters that could be used for this.

willmcgugan:
I'd love to implement this. Not certain how to render it yet. Rich shows a few lines before and after the highlighted line, and I wouldn't want to break that up with the `~~~~~~~^^^` characters.

But there are other ways we could emphasize those error columns, with underline or reverse video perhaps. We can experiment a little.

@darrenburns @DrBenton Any suggestions?

DrBenton:
> Note: TracebackException now has the offset and end_offset attributes which expose the information needed to print the arrows.
> https://github.com/python/cpython/blob/3.11/Lib/traceback.py#L852-L864

It seems that the `offset` and `end_offset`  attributes are set only for SyntaxErrors, if I'm not wrong?

It seems that FrameSummary instances, generated via `StackSummary.extract(walk_tb(traceback))` , now have a `colno` and `end_colno` attributes, which I guess we could potentially use to report the exact location of the error on the line.
[...]

PR #3389, opened 2024-06-17
2286 pep657 error code ranges in traceback

Gothingbop:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Pull request for https://github.com/Textualize/rich/issues/2286

Replicates some of the functionality in [PEP 657](https://peps.python.org/pep-0657) by extracting the col range of the exception and adds the style to the Syntax using `Syntax.stylize_range()`.

Effectively repurposes code found in some private functions in the base `traceback` module: `_get_code_position()` and `_walk_tb_with_full_positions()`. The code would be cleaner if those functions were not private or we were OK with using private functions (raises an error when running the mypy type checking).

I can also abstract the loop into a generator to more closely replicate the `_walk_tb_with_full_positions()` functionality, but I'm unsure where the best spot for that would be.

### Base Python traceback
image

### Rich traceback before
image

### Rich traceback with highlighing
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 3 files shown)


willmcgugan:
Great in principle. 

I think we can improve on the styles a bit. And some of the information present in the ascii version doesn't seem to be styled.

This will also need to handle older Pythons.

willmcgugan:
Thanks. Closed in favor of https://github.com/Textualize/rich/pull/3486

Issue #3393, opened 2024-06-20
Register custom __repr__ functions outside of their classes


danijar:
**How would you improve Rich?**

Add a function to register custom representation functions with rich, e.g.:

```py
rich.pretty.add_custom_formatter(condition, formatter)
```

Example usage:

```py
import rich
import rich.pretty

def array_formatter(x):
  if x.size <= 100:
    return repr(x)
  return f'{type(x)}(shape={x.shape} dtype={x.dtype})'

rich.pretty.add_custom_formatter(
    lambda x: hasattr(x, 'shape') and hasattr(x, 'dtype'),
    array_formatter)

import numpy as np
import jax.numpy as jnp

obj = {'foo': np.zeros((10, 10, 10)), 'bar': jnp.ones((1000,))}
rich.pretty.pprint(obj)
# {
#    'foo': np.ndarray(shape=(10, 10, 10), dtype=float64),
#    'bar': jax.Array(shape=(1000,), dtype=float32),
# }
```

Or equivalently:

```py
@functools.partial(rich.pretty.add_custom_formatter, condition)
def formatter(x):
  ...
```

**What problem does it solve for you?**

We're using `rich.traceback(show_locals=True)` in a large code base that makes heavy use of Numpy and JAX. While it's generally very helpful, it frequently prints arrays that take up a whole screen height in the stack trace. As a result, we find ourselves repeatedly commenting the rich traceback hook in and out.

For our own Python classes, we could just implement `__repr__` or `__rich_repr__`. However, for Python objects of external dependencies, we'd have to monkey-patch these, which poses the rich of changing behavior oin unexpected ways. Moreover,  the methods cannot be overridden for types implemented in C, such as `np.ndarray` or primitive types.

The suggested feature would make it really easy for users to adjust the tracebacks (and pretty printing in general) to their needs without affecting existing functionality.

leogott:
Do you need any info from these specific locals? If so, then you can stop reading here.

But if not, there's already this mechanism for excluding certain locals from the traceback by name:
https://github.com/Textualize/rich/blob/22c2cffd8e88181ad1162ca9098d190ec28c6996/rich/traceback.py#L436-L441
that could be extended with a new `show_locals_predicate` (or something) argument to traceback.install .
(Alternatively widen the type of `show_locals` to `bool | Callable`)

```py3
# ... L440
if not show_locals_predicate(key, value):
    continue
```

That would feel pretty clean to me.
This predicate could be of type
```py3
[...]

PR #3394, opened 2024-06-20
Make it possible to override the representation function from the outside

danijar:
A minimal change that makes it possible to override the representation function from the outside. This unblocks us now and could lay the basis for #3393 in the future.

## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

### DIFF
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -594,7 +594,7 @@ def traverse(
             obj_repr = f"{obj[:max_string]!r}+{truncated}"
         else:
             try:
-                obj_repr = repr(obj)
+                obj_repr = get_repr(obj)
             except Exception as error:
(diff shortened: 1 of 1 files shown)


danijar:
Hi @willmcgugan, could you please take a quick look at this one?

willmcgugan:
Can't accept this I'm afraid. I'm assuming you want to monkey patch `get_repr`? 

Monkey patching isn't good API, but if you are comfortable with it, you could just do this:

```python
globals()['repr'] = lambda obj: "foo"
```

That said, I think the best approach right now would be to write a method that accepts your objects and returns a proxy with a `__rich_repr__`.

Issue #3417, opened 2024-07-06
[BUG] Dataclass attributes specified with `init=False` cause error in pretty printing


CollinHeist:
- [x] I've checked [docs](https://rich.readthedocs.io/en/latest/introduction.html) and [closed issues](https://github.com/Textualize/rich/issues?q=is%3Aissue+is%3Aclosed) for possible solutions.
- [x] I can't find my issue in the [FAQ](https://github.com/Textualize/rich/blob/master/FAQ.md).

**Describe the bug**

Objects in pretty traceback printing which are `dataclass` objects with fields that are specified with `init=False` will cause an uncaught exception in the printing. See below.

```py
from dataclasses import dataclass, field, fields

@dataclass(eq=False)
class BadDataclass:
    item: int = field(init=False)

# item is not provided
ob = BadDataclass()

# Displays all fields, including item
print([f for f in fields(ob) if f.repr])

# Raises an AttributeError since ob.item is not defined
print([getattr(ob, f.name) for f in fields(ob)])
```

This is specifically caused by [this line](https://github.com/Textualize/rich/blob/master/rich/pretty.py#L782) of code.

[PR Submitted](https://github.com/Textualize/rich/pull/3418)

PR #3418, opened 2024-07-06
Handle dataclasses with unitialized fields in pretty printing

CollinHeist:
Fixes https://github.com/Textualize/rich/issues/3417

## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This is a fix for #3417. If the field has no defined attribute, then the default value should be printed instead, rather than raising an uncaught `AttributeError`.


willmcgugan:
This probably isn't the fix. It can give the false impression that the attribute exists. Better to omit it entirely IMO.

CollinHeist:
Sure, I've made that change.

willmcgugan:
Thanks, but I've gone with a slightly different solution... https://github.com/Textualize/rich/pull/3472

PR #3421, opened 2024-07-10
style(tree): format with black

TomJGooding:
Run black over recent changes to `tree.py`. This is currently causing PR checks to fail.


TomJGooding:
Sorry @willmcgugan I know you have a backlog, but just a gentle ping as merging this now might save you time later.

PR #3426, opened 2024-07-12
Add support for continuation indent and optimize divide_line

noahbkim:
## Type of changes

- [x] New feature
- [x] Tests

## Checklist

## Description

This PR adds an attribute `continuation_indent: int` to `rich.text.Text` that prefixes wrapped lines with a corresponding number of spaces. A paragraph wrapped to 72 characters might look like:

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
commodo consequat.
```

With `continuation_indent=4`, it would look like:

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
    tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
    veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex
    ea commodo consequat.
```

This is particularly useful for rendering multiline output that may become ambiguous without some indication of where text wrapping occurred. For example, printing a list of (very contrived) CLI commands:

```
echo reading file /path/to/file.txt
ssh user@host -- cat /path/to/file.txt
```

Might wrap to:

```
echo reading file
/path/to/file
ssh user@host --
cat /path/to/file
```

Which has a significantly different meaning. Using `continuation_indent=4`, we instead get:

```
echo reading file
    /path/to/file
ssh user@host --
    cat /path/to/file
```

In the process of implementing this I heavily refactored `divide_line` to no longer use `chop_cells`. Not only is this much faster (it no longer creates a temporary list of list of substrings), but it is more correct in the degenerate case of wrapping to a single column (though I'm happy to revert this if it's not preferred):

```
# Old behavior
>>> rich.text.Text("ab cd").wrap(rich.get_console(), 1)
Lines([<text 'a' []>, <text 'b' []>, <text ' ' []>, <text 'c' []>, <text 'd' []>])

# New behavior (no lines only containing whitespace)
>>> rich.text.Text("ab cd").wrap(rich.get_console(), 1)
Lines([<text 'a' []>, <text 'b' []>, <text 'c' []>, <text 'd' []>])
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,11 +25,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Running tests in environment with `FORCE_COLOR` or `NO_COLOR` environment variables
 - ansi decoder will now strip problematic private escape sequences (like `\x1b7`) https://github.com/Textualize/rich/pull/3278/
 - Tree's ASCII_GUIDES and TREE_GUIDES constants promoted to class attributes
+- Optimizes `_wrap.divide_line` by removing several `list` instantiations and
+  string slices.
 
 ### Added
 
 - Adds a `case_sensitive` parameter to `prompt.Prompt`. This determines if the
   response is treated as case-sensitive. Defaults to `True`.
+- Adds a `continuation_indent` argument to `text.Text`, which prefixes subsequent
+  wrapped lines with the specified number of spaces.
 
 ## [13.7.1] - 2024-02-28
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -36,6 +36,7 @@ The following people have contributed to the development of Rich:
 - [Hugo van Kemenade](https://github.com/hugovk)
 - [Andrew Kettmann](https://github.com/akettmann)
 - [Alexander Krasnikov](https://github.com/askras)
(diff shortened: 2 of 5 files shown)


noahbkim:
@willmcgugan sorry to ping you directly but I was wondering if there was anything I should be doing to get this through review!

willmcgugan:
There is a backlog that I'm slowly working through. But I will get to it.

It might help to understand a bit about why you need this. Is there precedence of this style of output?

noahbkim:
This was inspired by how PyCharm (and perhaps other modern editors) adds continuation indents when you've enable text-wrapping for documents with long lines. I'm using `rich` to display a report of commands run + results, and this would be particularly useful for that.

In the event you don't like continuation indents as a whole, I'd be happy to take a crack at refactoring this so I can more easily patch it myself.

willmcgugan:
Sorry, I don't think I can accept this as a feature, unless I was sure there was enough support for it. 

Optimizations / refactor would be fine.

noahbkim:
>Sorry, I don't think I can accept this as a feature, unless I was sure there was enough support for it.

No worries, thanks for taking a look!

PR #3429, opened 2024-07-15
Fix incorrectly converting relative to absolute of files and always show paths in traceback in case file cannot be found

AndreasBackx:
## Type of changes

- [x] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Currently Rich assumes that if a path in a traceback frame is relative, that it's relative to the current working directory. This is not the case when the Python application is running from a packaged bundle like a [PAR](https://google.github.io/subpar/subpar.html). In these cases the traceback loses the path information or the incorrect absolute path is used. This PR makes it so:
- `Frame.filename` is not absolute if the original filename does not exist in the first place. That would indicate that it's not relative to the current working directory.
- The path/filename is always shown in the traceback, not only if it exists.

I was considering replacing `rich.__init__._IMPORT_CWD`'s value to one that is reflective of `__file__` though that resulted in filenames relative to `/proc`. Those are not readable and this might change other expectations.

The `linecache` module understands the relative paths and even sometimes can give the lines inside of the PAR. (Though sometimes it cannot and I'm not sure when it can and cannot.) Regardless, this means we at least see some lines and also always see the location.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- `Frame.filename` no longer incorrectly set to an incorrect absolute path if the original path cannot be found.
+
+### Changed
+
+- Traceback filename/path is always shown, even if the file cannot be found.
+
 ## [13.9.2] - 2024-10-04
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


willmcgugan:
Any chance of tests? Although I can see why tests are tricky with this kind of thing.

AndreasBackx:
@willmcgugan  Will have a look when I'm back in the office. Currently on long(er) holiday. Could you have a look at #1859?

AndreasBackx:
I've added some tests to test the behaviour here. It's a bit annoying to test because of how `types.FrameType` and similar classes are not instantiatable. This might be a bit fragile and maybe mocks should be used instead of "fake" dataclasses. Though I feel like that's equally ugly and annoying to work with.

PR #3462, opened 2024-08-20
Change return type of `Live`'s contextmanager to `Self`

patrick91:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

When subclassing Live and using the new class a contextmanager I don't get the proper type hints, since `__enter__` returns `Live` and not my subclass, this PR fixes that 😊

### DIFF
--- a/rich/live.py
+++ b/rich/live.py
@@ -3,6 +3,12 @@ from threading import Event, RLock, Thread
 from types import TracebackType
 from typing import IO, Any, Callable, List, Optional, TextIO, Type, cast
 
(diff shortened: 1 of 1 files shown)


patrick91:
Ah, the CI error is due to mypy being less than v1.0: https://github.com/python/mypy/blob/master/CHANGELOG.md#support-the-self-type

happy to update that if that's ok (I'd do another since there might be other CI issues with the update)

willmcgugan:
Looks that like was done in another PR!

Issue #2563, opened 2022-10-04
[BUG] Recorded content not expected for captured prints


jylind:
**Describe the bug**

The following code:
```python
    console = Console(record=True)
    print("Before Capture started:")
    console.print("[blue underline]Print 0")
    with console.capture() as capture:
        console.print("[blue underline]Print 1")
        console.print("[blue underline]Print 2")
        console.print("[blue underline]Print 3")
        console.print("[blue underline]Print 4")
    print("\nCaptured content:")
    print(capture.get())
    print("\nRecorded content:")
    print(console.export_text())
```
Produces following output:

![Recording error](https://user-images.githubusercontent.com/15032420/193906064-a7265d2b-a9d2-472e-a8c7-4926511d7717.png)

It seems that every print to capture adds previously printed content + new content.

I also wonder if the captured content should even go to the recording because it won't even go to console unless specifically printed there after capturing has been stopped.

Though, as it seems that captured content cannot be printed with `console.print()` (control codes seems to becomes visible) so by using regular python print() would not put it in the recording either.
(Line `print(capture.get())` replaced with `console.print(capture.get())`)

![Recording error - capture print](https://user-images.githubusercontent.com/15032420/193908546-e9d869d5-9d3d-45b8-9b40-acce6ae25cd8.png)

**Platform**
<details>
<summary>Click to expand</summary>

Same with Win/Linux platforms.
Windows terminal: Cmd prompt, Windows Terminal.
Linux Terminal: Putty

Using Rich version 12.6.0

</details>

ssbarnea:
That bug was easy to reproduce and was introduced sometime after 12.5.0 was released. Still I was not able to find much activity in the last two years on the entire rich library. I ended-up capping `rich<=12.5.0` long time ago.

Bug still reproduces with rich 13.7.1

TomJGooding:
From just five minutes looking into this, I think issue #2172 might be relevant here with different expectations for recorded content.

PR #3470, opened 2024-08-26
Fix record and capture

willmcgugan:
Fix https://github.com/Textualize/rich/issues/2563


linbrandon4:
Nice targeted bugfix. “Record” vs “capture” is one of those features that feels simple from the outside, but it’s easy to get wrong because you’re juggling console state, buffers, and what counts as “output” vs “rendered content.” This PR looks like it fixes the mismatch reported in #2563 (recorded content not matching what you’d expect from captured prints), and it’s small enough that it’s easy to reason about.

I also like that you included a changelog commit — for output/capture bugs, people do notice the behavior change, so documenting it is helpful.

Only things I’d sanity-check in this area:

make sure capture/record behavior is consistent across plain print, console.print, and anything that writes directly to the console

confirm nested capture/record contexts don’t leak output or double-record

regression coverage looks good (Codecov says modified lines are covered), so this should be pretty safe

Overall: small diff, clear issue link, tests covered, and the changelog makes it easy for users to understand what changed.

PR #3533, opened 2024-10-22
Generic speed column for Progress

NichtJens:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

I was looking for a generic speed column for `Progress` preferably with customizable units. I only found #926, which hinted at `TransferSpeedColumn` being a good example, but did not have a resolution.

This PR adds `SpeedColumn`, which may be used like this:
```
SpeedColumn() # default units="it/s"
SpeedColumn(units="Hz")
```

I reuse `TaskProgressColumn.render_speed` for consistency and extended both `TaskProgressColumn.render_speed` and `TaskProgressColumn` itself to also accept custom units.

Regarding the unit tests, I added a line mimicking the test for `TransferSpeedColumn` as well as extending `test_task_progress_column_speed` to test `TaskProgressColumn.render_speed` with custom `units`. Is there more desired here?

---

I believe, I also fixed a documentation bug in `TaskProgressColumn.render_speed`, which referred to:
- `task (Task): A Task object.`

as argument, where it should have been:
- `speed (float): Current value of the speed.`


leahoswald:
Searched for exactly this feature, but it isn't merged :(

PR #3535, opened 2024-10-22
Regex error

willmcgugan:
Fixes an issue with the regex used to select the fast path for strings with single-width characters.

This may have had a detrimental effect on performance for some operations.


TomJGooding:
Sorry looks like you beat me to it!

But I think you might be still missing z?

darrenburns (approved):

PR #3546, opened 2024-10-31
use sets

willmcgugan:
- Faster detection of single length widths
- Faster sum by using map rather than a list expression

### DIFF
--- a/rich/cells.py
+++ b/rich/cells.py
@@ -6,6 +6,7 @@ from typing import Callable
 from ._cell_widths import CELL_WIDTHS
 
 # Ranges of unicode ordinals that produce a 1-cell wide character
+# This is non-exhaustive, but covers most common Western characters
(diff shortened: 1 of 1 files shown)


ofek:
I would recommend lazily defining constants like that so import time is not affected, especially since that is critical for command line applications. See PEP 562 https://peps.python.org/pep-0562/

willmcgugan:
Agree in principle, but this data is used by virtually anything Rich does. So it is always imported.

It also turns out that it is faster to compute this than import it. 0.11ms to compute version 0.39ms to import.

So on my Macbook at least, it is a win to compute it.

ofek:
Definitely makes sense if it's used everywhere, thanks! I'm just always looking for ways to reduce the import time of Rich, currently I think it could be better but I haven't had much time to look into that.

PR #3622, opened 2025-01-28
Skip tests which are expected to fail with Python 3.14

befeleme:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

- [ ] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate. -- not sure this requires a changelog entry
- [-] I've added tests for new code.

## Description

In Fedora Linux, we've started testing packages with Python 3.14. When I deal with the (expected) failure of these 5 tests, rich builds just fine.

### DIFF
--- a/tests/test_inspect.py
+++ b/tests/test_inspect.py
@@ -43,6 +43,11 @@ skip_py313 = pytest.mark.skipif(
     reason="rendered differently on py3.13",
 )
 
+skip_py314 = pytest.mark.skipif(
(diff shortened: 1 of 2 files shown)


hroncok:
Would it make sense to have `@skip_py311_or_later` at this point?

willmcgugan:
Thanks. The work was done in https://github.com/Textualize/rich/pull/3861

PR #3624, opened 2025-01-29
Remove unnecessary import that confuses type checkers

itamaro:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

in `rich.table`, when executed as a main script, it imports `Table` from itself.

this seems unnecessary, considering the `Table` class is implemented in the same module.
this isn't only cosmetic though - at least one type checker (Pyre) gets confused because of this, and treats `Table` as a symbol that isn't exported by `rich.table`.

please let me know if this change is acceptable, and if so, whether it needs a CHANGELOG entry and/or additional tests.

### DIFF
--- a/rich/table.py
+++ b/rich/table.py
@@ -929,7 +929,6 @@ class Table(JupyterMixin):
 if __name__ == "__main__":  # pragma: no cover
     from rich.console import Console
     from rich.highlighter import ReprHighlighter
-    from rich.table import Table as Table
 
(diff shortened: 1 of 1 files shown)


nakowafers (approved):

willmcgugan:
I think pyre is in the wrong here, but ok.

Issue #3263, opened 2024-01-23
[BUG] Text inside Live with vertical_overflow="visible" duplicating when above console.height instead of scrolling


RasyiidWho:
- [x] I've checked [docs](https://rich.readthedocs.io/en/latest/introduction.html) and [closed issues](https://github.com/Textualize/rich/issues?q=is%3Aissue+is%3Aclosed) for possible solutions.
- [x] I can't find my issue in the [FAQ](https://github.com/Textualize/rich/blob/master/FAQ.md).

**Describe the bug**

I try to create function to streaming from incoming text using both rich.markdown and rich.live. In this example, I use simple number loop with the delay to simulate the streaming. Here's the simple example:

```
import time
from rich.markdown import Markdown
from rich.live import Live

def stream_numbers(chunk_size=1):
    for i in range(1, 11, chunk_size):
        yield f"\n\n{i}"
        time.sleep(0.01)


render_this = ""
with Live(render_this, auto_refresh=False, vertical_overflow="visible") as live:
    print(f"Console height: {live.console.height}")
    for entry in stream_numbers():
        render_this += entry
        live.update(Markdown(render_this), refresh=True)
```

Result:

```
1
1
(newline)
[...]
(newline)
10
```

Expected result:
Output scrolling after excess console.height,
```
1
(newline)
[...]
(newline)
10
```

Result only breaking (duplicating) only when loop are above console.height. In this example, my console height are 18, and I try to print above it (19 newline).

**Platform**
<details>
<summary>Click to expand</summary>

Windows 11 with Python 3.10, Trying in: VSCode Terminal, Windows Terminal, and Cmder

```
╭───────────────────────── <class 'rich.console.Console'> ─────────────────────────╮
│ A high level console interface.                                                  │
│                                                                                  │
│ ╭──────────────────────────────────────────────────────────────────────────────╮ │
│ │ <console width=100 ColorSystem.TRUECOLOR>                                    │ │
│ ╰──────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                  │
│     color_system = 'truecolor'                                                   │
│         encoding = 'utf-8'                                                       │
│             file = <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'> │
│           height = 24                                                            │
│    is_alt_screen = False                                                         │
│ is_dumb_terminal = False                                                         │
│   is_interactive = True                                                          │
│       is_jupyter = False                                                         │
│      is_terminal = True                                                          │
[...]

PR #3637, opened 2025-02-20
Add a `vertical_overflow='crop_above'` as an option to `Live()`

cpsievert:
## Type of changes

- [x] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

As described in #3263, `Live(vertical_overflow="visible")` has the unfortunate behavior of duplicating content when the content update exceeds the height of the console. This is especially unfortunate for a few reasons:

* It seems there is no way to get rid of the duplicated content (i.e., `transient=True` doesn't help to avoid this).
* There are no other `vertical_overflow` options that allows for the "newest" content to be visible.
* In today's Gen AI world where markdown often is received in chunks, it's really useful to have a live display where a (possibly long) markdown string accumulates over time.

This PR proposes a new option `vertical_overflow="crop_above"` which does the reverse of `vertical_overflow="crop"` (it displays only the bottom portion of the content instead of the top). It has the nice behavior of always making the "newest" content visible, but without the downside of duplicated content. Here's a demo:

```python
import requests
import time

from rich.live import Live

readme = requests.get(
    "https://raw.githubusercontent.com/posit-dev/py-shiny/refs/heads/main/README.md"
)
readme_chunks = readme.text.replace("\n", " \n ").split(" ")[: 200]

content = ""
with Live(auto_refresh=False, vertical_overflow="crop_above") as live:
    for chunk in readme_chunks:
        content += chunk + " "
        time.sleep(0.01)
        live.update(Markdown(content), refresh=True)
```

https://github.com/user-attachments/assets/6425a645-2658-46a5-9b49-12c63a154063

And note that if you change `vertical_overflow="crop_above"` to `vertical_overflow="visible"`, this is the behavior:

https://github.com/user-attachments/assets/35f95d37-f1f3-4afb-a0ac-9d360fdac3d6

I'm happy to write tests or anything else you need if you like this overall direction

### DIFF
--- a/rich/live_render.py
+++ b/rich/live_render.py
@@ -14,7 +14,7 @@ from .segment import ControlType, Segment
 from .style import StyleType
 from .text import Text
 
-VerticalOverflowMethod = Literal["crop", "ellipsis", "visible"]
+VerticalOverflowMethod = Literal["crop", "crop_above", "ellipsis", "visible"]
 
 
 class LiveRender:
@@ -91,6 +91,9 @@ class LiveRender:
         if height > options.size.height:
             if self.vertical_overflow == "crop":
                 lines = lines[: options.size.height]
+                shape = Segment.get_shape(lines)
+            elif self.vertical_overflow == "crop_above":
+                lines = lines[-(options.size.height) :]
                 shape = Segment.get_shape(lines)
             elif self.vertical_overflow == "ellipsis":
                 lines = lines[: (options.size.height - 1)]


LeoYoung-code:
👍

david-zlai:
would love to see this!

@willmcgugan  - if we can get a r?

timesler:
This feature would be amazing for CLI LLM chat apps!

timesler:
@willmcgugan sorry to spam you. I wanted to frame this a bit and why I think it's super important. LLM agents are proliferating pretty wildly at the moment, and one of the quickest ways to build really effective UIs to give users access to them is via CLIs. IMO, `rich` is the best tool for building beautiful CLI apps (in python or otherwise), and I think that this specific feature is the single biggest blocker to it being near perfect for that use case

willmcgugan (changes requested):
Going to need an update to the docs (live.rst)

willmcgugan:
You would be better off using [Textual](https://textual.textualize.io/) for LLM output. Better Markdown rendering as well.

willmcgugan:
Rich just isn't the right tool for this kind of interface. Use Textual for Markdown streaming, and scrollable windows.

PR #3648, opened 2025-03-02
Fixed legacy_windows is auto-detected as True incorrectly, when not in terminal environment

### LINKED ISSUE #3647: [BUG] legacy_windows is True when is_terminal is False

**Describe the bug**

The console object's `legacy_windows` is True when `is_terminal` is False. I think this is a bug, because it's unable to successfully calling Windows console API on non-terminal environment.

```python
# foo.py

console = Console()
print(console.legacy_windows)
```

```shell
> python foo.py 
False
> python foo.py | echo
True
```

**Platform**

I use terminal in VS Code on Windows 10.

xymy:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Fix #3647

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -585,11 +585,6 @@ def get_windows_console_features() -> "WindowsConsoleFeatures":  # pragma: no co
     return _windows_console_features
 
 
-def detect_legacy_windows() -> bool:
-    """Detect legacy Windows."""
-    return WINDOWS and not get_windows_console_features().vt
-
-
 class Console:
     """A high level console interface.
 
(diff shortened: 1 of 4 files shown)


rich/console.py
```diff
@@ -585,11 +585,6 @@ def get_windows_console_features() -> "WindowsConsoleFeatures":  # pragma: no co
     return _windows_console_features
 
 
-def detect_legacy_windows() -> bool:
-    """Detect legacy Windows."""
-    return WINDOWS and not get_windows_console_features().vt
-
-
```
Jackenmen, line 592: This appears to be a breaking change as `detect_legacy_windows()` is a documented function: https://rich.readthedocs.io/en/latest/reference/console.html#rich.console.detect_legacy_windows

DTiming24:
Good fix for the legacy_windows detection. Since this only shows up outside a terminal environment, a small test covering a non-interactive console setup would be a nice safety net.

Issue #3641, opened 2025-02-26
[BUG] Right-justified panels with right-aligned titles display incorrectly


NorthIsUp:
- [ ] I've checked [docs](https://rich.readthedocs.io/en/latest/introduction.html) and [closed issues](https://github.com/Textualize/rich/issues?q=is%3Aissue+is%3Aclosed) for possible solutions.
- [ ] I can't find my issue in the [FAQ](https://github.com/Textualize/rich/blob/master/FAQ.md).

**Describe the bug**

a picture is worth a thousand words, the issue is the gap in the box

![Image](https://github.com/user-attachments/assets/83d9559a-541e-4a8d-87b6-70ed85221a82)

```

**Platform**
<details>
<summary>Click to expand</summary>

What platform (Win/Linux/Mac) are you running on? What terminal software are you using?

I may ask you to copy and paste the output of the following commands. It may save some time if you do it now.

If you're using Rich in a terminal:

```
python -m rich.diagnose
pip freeze | grep rich
```

If you're using Rich in a Jupyter Notebook, run the following snippet in a cell
and paste the output in your bug report.

```python
from rich.diagnose import report
report()
```

</details>

NorthIsUp:
```
def test_with_justify() -> None:
    panel = Panel(
        "Hello, World",
        title="title",
        subtitle="subtitle",
        title_align="right",
    )
    console = Console(
        file=io.StringIO(),
        width=50,
        height=20,
        legacy_windows=False,
        force_terminal=True,
        color_system="truecolor",
    )
    console.print(panel, justify="right")

    result = cleandoc(console.file.getvalue())
    expected = cleandoc("""
        ╭────── title ─╮
        │ Hello, World │
        ╰── subtitle ──╯
        """)
    assert result == expected
```

BenRachmiel:
Hey, I've taken a look at the issue and it appears to stem from containers.justify (specifically lines 139-142, where the trailing whitespace gets mistakenly rstrip'd and then padded to the beginning of the segment. If you need a fix: I've found that changing the line "line.rstrip()" to "line[:-2].rstrip()" prevents this behavior, and, at least in the case of the example you gave, solves the issue. 

This isn't an elegant solution, but seeing as the titles are padded with whitespace it may be "right." Checking tests, I don't see that this introduces regressions covered by the test. If @willmcgugan doesn't have an issue with it, I'll submit a pull request introducing the change.

NorthIsUp:
[...]

PR #3649, opened 2025-03-02
Fix right justification: conditionally preserve trailing space character

BenRachmiel:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

#3641 details an issue with right justified titles of panels losing the trailing whitespace, and it being reassigned to the beginning of the segment. It appears that trailing whitespace is removed via rstrip, but titles have a singular space added after the text (I assume for visual clarity), which is inadvertently removed.

Seeing as this change might have knock-on effects, I attempted to mitigate that by using a ternary if, checking that the final character is a space and if so, leaving that space out of the rstrip. I have not added a test, since I do not know if the test mentioned in the issue is appropriate (seems very hardcoded), but could think up a regression test if necessary (don't know exactly how atm, but I could learn). 

First time contributing to open source, hope this isn't a dumb way to tackle the issue, but I do see how it could have some unintended consequences.

### DIFF
--- a/rich/containers.py
+++ b/rich/containers.py
@@ -76,12 +76,10 @@ class Lines:
         return iter(self._lines)
 
     @overload
-    def __getitem__(self, index: int) -> "Text":
-        ...
+    def __getitem__(self, index: int) -> "Text": ...
 
     @overload
-    def __getitem__(self, index: slice) -> List["Text"]:
-        ...
+    def __getitem__(self, index: slice) -> List["Text"]: ...
 
     def __getitem__(self, index: Union[slice, int]) -> Union["Text", List["Text"]]:
         return self._lines[index]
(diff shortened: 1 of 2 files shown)


rich/containers.py
```diff
@@ -137,7 +135,7 @@ def justify(
                 line.pad_right(width - cell_len(line.plain))
         elif justify == "right":
             for line in self._lines:
-                line.rstrip()
+                line[:-1].rstrip() if line[-1] == Text(" ") else line.rstrip()
```
willmcgugan, line 138: Since the markup is irrelevent for the condition, it is faster to do `line.plain.endswith(" ")`

willmcgugan (changes requested):

willmcgugan:
I don't think this is neccesary. Maybe use a non breaking space or something.

PR #3691, opened 2025-04-02
Fix progress cursor visibility

shyam-ramani:
# Fix cursor visibility issues when Progress/Live displays are interrupted

## Problem

When a Rich Progress or Live display is interrupted (e.g., by a Ctrl+C or an unhandled exception), the cursor sometimes remains hidden in the terminal, making it difficult for users to type or interact with the console after the program ends.

## Root Cause

The `Progress.stop()` and `Live.stop()` methods handle cleaning up the console state, including restoring cursor visibility, but they lack robust exception handling. If an exception occurs during the cleanup process, the code to restore cursor visibility might not execute.

## Solution

This PR implements the following improvements:

1. Added try/except handlers in both `Progress.stop()` and `Live.stop()` methods to ensure cursor visibility is restored even if an exception occurs during cleanup.

2. Added signal handlers (SIGINT) to catch Ctrl+C interruptions and ensure cursor visibility is restored before the program exits.

3. Added an atexit handler as an additional safety measure to ensure cursor visibility is restored when the program exits.

4. Added platform checks to avoid setting up signal handlers on platforms that don't support them (e.g., Windows).

## Benefits

- Users will no longer experience "invisible cursor" issues when interrupting a Progress or Live display with Ctrl+C.
- More robust error handling ensures cursor visibility even in edge cases.
- Maintains backward compatibility with all existing code.

## Testing

A test script (`test_cursor_visibility.py`) is included to demonstrate the fix. The script shows a progress bar that can be interrupted with Ctrl+C, and verifies that the cursor remains visible after interruption.

## Notes

This fix builds on the foundation of existing cursor management in Rich, adding additional safeguards to handle edge cases without changing the core functionality.

### DIFF
--- a/create_pr.sh
+++ b/create_pr.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Navigate to the project directory
+cd "$(dirname "$0")"
+
+# Create a new branch
+git checkout -b feature/console-buffering-improvements
+
+# Add the modified files
+git add rich/console.py tests/test_console_buffering_improvements.py pull_request.md
+
+# Commit the changes
+git commit -m "Add console buffering improvements"
+
+# Push the branch to the remote repository
+git push origin feature/console-buffering-improvements
+
+echo "Pull request branch created and pushed successfully!"
+echo "You can now create a pull request on GitHub." 
--- a/pull_request.md
+++ b/pull_request.md
@@ -1,138 +1,112 @@
-# Fix Console Output Buffering Issues
+# Console Output Buffering Improvements
 
-## Issue Description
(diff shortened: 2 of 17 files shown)


TomJGooding (commented):
Please don't waste the time of open source maintainers with this LLM-generated slop.

Hopefully you realise by now that all your pull requests will be rejected. This PR is probably the most egregious as you didn't even notice to remove the `create_pr.sh` or `pull_request.md` files.

I'm not sure what your motivation is here, but I'm assuming you're new to coding. Using LLMs to generate pull requests may seem like an easy way to get started, but it's not a substitute for real learning and experience.

PR #3700, opened 2025-04-15
Remove runtime dependency on typing-extensions

ichard26:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

- [x] (**NA**) I've added tests for new code.

## Description

I'd like to get rid of pip's vendored copy of typing-extensions. Rich is the primary blocker, but it turns out that rich doesn't really need typing-extensions since requiring Python 3.8+.

It is still used for typing_extensions.Self (which was only added to the stdlib in Python 3.11) but the import is now guarded under TYPE_CHECKING so end-users don't have to install typing_extensions.


ichard26:
This was implemented by https://github.com/Textualize/rich/pull/3763.

PR #3702, opened 2025-04-16
Custom message on `Live` overflow

samuelcolvin:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Allow a custom message with a `Live` overflows, instead of an ellipsis.

### DIFF
--- a/rich/live.py
+++ b/rich/live.py
@@ -1,7 +1,19 @@
 import sys
 from threading import Event, RLock, Thread
 from types import TracebackType
-from typing import IO, Any, Callable, List, Literal, Optional, TextIO, Tuple, Type, Union, cast
+from typing import (
+    IO,
+    Any,
+    Callable,
+    List,
+    Literal,
+    Optional,
+    TextIO,
+    Tuple,
+    Type,
+    Union,
+    cast,
+)
 
 from . import get_console
 from .console import Console, ConsoleRenderable, RenderableType, RenderHook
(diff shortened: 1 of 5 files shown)


samuelcolvin:
Bump @willmcgugan, could we at least run tests.

(Curious why you don't allow tests to run for all PRs?)

willmcgugan:
> (Curious why you don't allow tests to run for all PRs?)

That used to be the case. Github changed it, and I haven't changed it back.

rich/live.py
```diff
@@ -59,7 +71,9 @@ def __init__(
         transient: bool = False,
         redirect_stdout: bool = True,
         redirect_stderr: bool = True,
-        vertical_overflow: VerticalOverflowMethod = "ellipsis",
+        vertical_overflow: Union[
```
willmcgugan, line 74: I'm not keen on overloading a parameter like this. If it becomes a general purpose extension mechanism, it could quickly become unwieldy. I'd prefer to keep the message a separate value.

willmcgugan (changes requested):
I am reluctant to add new parameters, as constructor signatures can get excessively large.

Since this parameter doesn't seem essential, how about an alternative way of configuring it? Perhaps a `get_vertical_overflow_message` method, so it could be customized in a subclass.

samuelcolvin:
> Perhaps a `get_vertical_overflow_message` method, so it could be customized in a subclass.

On `Live` or `LiveRender` or both?

willmcgugan:
> On `Live` or `LiveRender` or both?

I think just `Live`. `LiveRender` isn't user-facing.

Issue #3286, opened 2024-02-23
Support for stdlib logging StreamHandler with `Live(redirect_stderr=True)`


gusostow:
### Discussed in https://github.com/Textualize/rich/discussions/3280

<div type='discussions-op-text'>

<sup>Originally posted by **gusostow** February 15, 2024</sup>
I want to set up a progress bar that isn't disrupted by standard logging statements. `redirect_stderr` is working fine when I print directly to stdout but not for logging. 

Is this expected? Are there any workarounds? Ideally I wouldn't have to change the handler to `RichHandler` when this optional progress bar is enabled.</div>

robinbowes:
Is this related to #2022 ?

SebastianGrans:
Hey, I was just hitting the same issue and found a workaround.

**tl;dr** Inside the `with Live()...`-clause, you need to update the `stream` field of the `StreamHandler` instance in the logger. See the example workaround below.


Sadly, I don't think there is a clean way that this could be automatically handled by rich. 

The reason why this happens is that with `basicConfig`, the logger is sets up a `StreamHandler` that logs to `sys.stderr` by default.

In `Progress` and `Live`, `sys.stdout/sys.stderr` is replaced with a `FileProxy`, but since `StreamHandler` was set up before `sys.stderr` was replaced, it will still print to the "regular" `sys.stdout`.


## Minimal workaround

```python
import logging
import sys
import time
from rich.progress import Progress
from rich.live import Live

streamhdlr = logging.StreamHandler()
logging.basicConfig(level=logging.DEBUG, handlers=[streamhdlr])
log = logging.getLogger()


with Progress(auto_refresh=False) as progress:
    task = progress.add_task("Processing...", total=100)
    streamhdlr.setStream(sys.stderr)
    for _ in range(100):
        log.info(f"Log message {_}")
        time.sleep(0.1)
        progress.update(task, advance=1)


with Live("Hello, World!", refresh_per_second=4) as live:
    streamhdlr.setStream(sys.stderr)
    for i in range(40):
        log.info("Updating live display")
        time.sleep(0.4)
        live.update(f"Count: {i}")
```

And it cleanly prints:

```bash
INFO:root:Log message 0
INFO:root:Log message 1
INFO:root:Log message 2
INFO:root:Log message 3
INFO:root:Log message 4
INFO:root:Log message 5
INFO:root:Log message 6
INFO:root:Log message 7
INFO:root:Log message 8
INFO:root:Log message 9
[...]

PR #3713, opened 2025-04-24
Fix: Make Console follow FileProxy redirections

dp-alvarez:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Currently, Live console redirection features only work when using the same Console that Live was created with, or with standard Python print().
This patch makes Console.print respect Live redirections across different Console instances.
The most immediate use case is making [stdlib logging work with redirections](https://github.com/Textualize/rich/issues/3286), but should also be useful in other scenarios.
PS: The LLM-generated PR spam is very real on this repo, what a pain.


dp-alvarez:
@willmcgugan do you have any feedback on this? I'd love to get rid of hacky workarounds I have in my projects.

PR #3755, opened 2025-06-08
Private random generator for style link

MamoruDS:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## Description

Added a private random generator as class attribute for `Style` (`Style._RNG`) for `randint` calls when creating link ids. This should not change any behavior of `Style` but avoids affecting the global random generator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- Added a random generator as class attribute for `Style` for `randint` calls when creating link ids to avoid affecting the global random generator.
+
 ## [14.0.0] - 2025-03-30
 
 ### Added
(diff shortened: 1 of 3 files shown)


willmcgugan:
I will need to understand what problem this solves for you. I suspect you will be better off using a random generator in your code, than modifying libraries.

Closing for now, but re-open if you can elaborate further.

MamoruDS:
Thanks for your time and for being open to reopening this. Let me elaborate on the problem this PR aims to solve.

The issue is reproducibility in scientific and machine learning codebases, which often rely on setting a single global `RNG` seed for their experiments. I agree with you that these codebases should ideally manage their own isolated `RNG`s. However, the reality is that a vast ecosystem of codebases still depends on the global `RNG` for convention and simplicity. This is common enough that frameworks like [pytorch-lightning](https://github.com/Lightning-AI/pytorch-lightning/blob/33de2307fcc94687e612781c780e3b7f081e4510/src/lightning/fabric/utilities/seed.py#L60) and [keras](https://github.com/keras-team/keras/blob/3bedb9a970394879360fcb1c0264f3ffdc634a77/keras/src/utils/rng_utils.py#L55) provide helper functions to set the seed for the global `RNG`. By giving `Style` its own private `RNG`, `rich` can integrate more seamlessly into these environments. This internal only change ensures `rich` doesn't interfere with the user's application state, improving compatibility with the scientific computing ecosystems.

I completely understand your principle about modifying library versus user code. I just wanted to provide this context for your consideration.

PR #3756, opened 2025-06-09
Fix: spacing on arc spinner

bitplane:
## Type of changes

- [x] Bug fix

## Checklist

## Description

The arc animation was missing spaces on a couple of the frames, making it drop back by one char and wibbling my wobblies. Patch is a one-liner that fixes it. Also tried to make it consistent size, but the top and bottom of the arc seem ambiguous, and take up 1 space in fonts that render them "properly" vs "nicely" 

> In [5]: wcswidth("◠")
> Out[5]: 1

Properly (it was never gonna work anyway)

Screenshot from 2025-06-09 08-46-13

Nicely (looks nice now)

Screenshot from 2025-06-09 08-49-19


willmcgugan:
That makes it looks worse for me. Terminals don't always agree with wcswidth. Getting it to work nicely everywhere is likely impossible I'm afraid. Thanks for trying.

PR #3759, opened 2025-06-12
Add option for custom animation frames when creating a Spinner

iamevn:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Adding a way to use a custom animation for a Spinner by providing a  `custom_spinner` argument to the `Spinner` constructor matching the format from the `SPINNERS` list in `_spinners.py`.

```python
Spinner("custom", custom_spinner={"interval": 80, "frames": ["[<  ]", "[ = ]", "[  >]", "[ = ]"]})
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- Added option to use custom animation for Spinner. https://github.com/Textualize/rich/pull/3757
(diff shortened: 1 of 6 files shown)


iamevn:
Before updating the docs I wanted to check and see if this looks like a reasonable way to make custom Spinners easier.

I'm also considering letting the user just provide frames and getting default (80ms) interval. e.g.:
```python
Spinner("custom", custom_frames=["[<  ]", "[ = ]", "[  >]", "[ = ]"])
```
if `custom_frames` arg is present, use that with default 80ms interval, else if `custom_spinner` argument is present use that's frames and interval, else look up `name` arg in `SPINNERS` dict

willmcgugan:
I'm afraid I'm not accepting new features, which could be implemented outside of the core library.

PR #3777, opened 2025-06-20
TTY_INTERACTIVE env var

willmcgugan:
Adds a TTY_INTERACTIVE to disable things like progress bar / status spinners in CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed extraction of recursive exceptions https://github.com/Textualize/rich/pull/3772
 
+### Addedd
+
+- Added `TTY_INTERACTIVE` environment variable to force interactive mode off or on.
+
 ## [14.0.0] - 2025-03-30
 
 ### Added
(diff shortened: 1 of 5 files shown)


15r10nk:
I assumed that this option had something to do with interactive terminals, where you can not read from stdin. But, If I understand you correct, you are still able to read user input, even when TTY_INTERACTIVE=0 is enabled. 

Maybe call it TTY_ANIMATION=0

willmcgugan:
It echos an existing attribute in Rich. In that context interactive means somebody (i.e. a human) is interacting with the terminal, as apposed to running on CI. You are still able to read input, but reading from stdin doesn't always requires a human operator.

It does disable animation, but the flag in Rich has broader meaning.

It could also be `RICH_INTERACTIVE`. Perhaps the namespace would avoid any confusion?

15r10nk:
Naming things is hard. I think naming it RICH_INTERACTIVE is a good choice if it is not a environment variable which is also used by other projects with the same meaning.

willmcgugan:
> Naming things is hard. I think naming it RICH_INTERACTIVE is a good choice if it is not a environment variable which is also used by other projects with the same meaning.

I'd actually like it to become a standard thing. AFAIK there is no standard var for that, and I'd like it to become one...

PR #3791, opened 2025-07-12
Allow custom spinners across `rich` library

### LINKED ISSUE #3790: [REQUEST] Custom spinner animation
Have you checked the issues for a similar suggestions? Yep

I want to use a custom spinner animation (not just one provided by `rich`. I can see that there is a consistent format for these, and so accepting custom spinner definitions would be a relatively simple improvement. In fact, I'm happy to attempt implementing this myself.

It means I can provide my own custom spinner animations, which will make me happy.

MaddyGuthridge:
## Type of changes

- [x] New feature
- [x] Tests

## Checklist

- [x] I accept that @willmcgugan may be pedantic in the code review. (I am keen!)

## Description

Resolves #3790 

***The feature***
* Adds type definition to spinner animation dictionary
* Allows a custom spinner animation parameter in place of a spinner name, allowing users to specify their very own animations if they desire
* This is implemented in `Spinner`, `Console.status`, `Status` and `progress.SpinnerColumn` (let me know if I missed anything)

***Documentation***
* I documented custom spinner support in `console.rst`
* Spinners don't have their own dedicated documentation page. Should one be added? If so, this should be documented there.
* Notably, all tests still pass.

***Is this worth adding despite Rich rejecting most new features?***
* Imo this feature is worthwhile adding, since it cannot easily be added through a third-party library (unless that library reimplements spinners as a whole). Imo this makes it easier to extend the core library.
* This shouldn't break backwards compatibility, as all changed function parameters still accept their original types, as well as the new types.

***Design considerations***
* In an ideal world, all these functions would accept a `Spinner` object themselves, rather than their assorted set of other parameters (`spinner`, `spinner_style` and `speed`), which would make patterns like dependency injection simpler
* However, accepting a `Spinner` object (rather than constructing one based on these parameters) would make those other parameters ambiguous in the case where a `Spinner` object is given.
* As such, this design is probably the best for fitting with the existing code-base, and avoiding confusion and breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `TTY_INTERACTIVE` environment variable to force interactive mode off or on https://github.com/Textualize/rich/pull/3777
+- Allowed custom spinner animations throughout the library https://github.com/Textualize/rich/pull/3791
 
 ## [14.0.0] - 2025-03-30
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -94,3 +94,4 @@ The following people have contributed to the development of Rich:
 - [Jonathan Helmus](https://github.com/jjhelmus)
 - [Brandon Capener](https://github.com/bcapener)
 - [Alex Zheng](https://github.com/alexzheng111)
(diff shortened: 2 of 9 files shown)


MaddyGuthridge:
Hello, @willmcgugan. I'd appreciate a review of this if possible. I'd love to have custom spinners for my CLI apps that use Rich, and have gotten a comment on the linked issue saying that I'm not the only one who would like this.

PR #3798, opened 2025-07-16
fixing with none rich colored text

tinker495:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [x ] I've run the latest [black](https://github.com/psf/black) with default args on new code.

## Description
This pull request addresses a bug where the Panel component, and others relying on cell_len, fail to correctly calculate the width of content that includes raw ANSI escape codes. This typically occurs when displaying text colored by a library other than rich.
The issue stems from cell_len not stripping these non-printable ANSI characters before measuring the string's visible length. Consequently, the escape codes are counted as part of the width, causing the Panel to render much wider than its actual content, breaking the layout.

This fix introduces a regular expression to rich/cells.py that sanitizes input strings by removing ANSI codes within the cell_len and cached_cell_len functions. This ensures that the measured width accurately reflects the visible text, allowing rich components to correctly render pre-formatted text.

## Before
A Panel wrapping a string with ANSI color codes renders incorrectly, with a width that far exceeds the visible content.
image

## After
The same Panel now correctly calculates the content width, ignoring the ANSI codes, and renders with the expected dimensions.
image

### DIFF
--- a/rich/cells.py
+++ b/rich/cells.py
@@ -1,9 +1,13 @@
 from __future__ import annotations
 
+import re
 from functools import lru_cache
 from typing import Callable
 
 from ._cell_widths import CELL_WIDTHS
+
+_RE_ANSI = re.compile(r"\x1b\[[0-9;]*m")
+
 
 # Ranges of unicode ordinals that produce a 1-cell wide character
 # This is non-exhaustive, but covers most common Western characters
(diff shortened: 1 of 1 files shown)


willmcgugan:
Let Rich generate escape sequences. But don't try and use them as input. If you have escape sequences you want to preserve, wrap them in `Text.from_ansi(my_string)`.

PR #3845, opened 2025-09-10
Use faster generator for link IDs

akx:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Other

## Checklist

## AI?

AI was not used to generate this PR.

## Description

I was profiling some code that uses `rich` and noticed that some unusual time was spent in `random.getrandbits()` (which underlies `random.randint()`). The default RNG in Python is not the fastest to begin with (which is why e.g. [`fastrand`](https://github.com/lemire/fastrand) is a thing).

This PR switches the generator for `_link_id`s from a random number to a simple sequential counter (initialized once with randomness). This also has the happy side effect that link ID collisions are simply not possible anymore, whereas the birthday paradox says `random.randint(0, 999999)` tends to have a collision in the first 1,253 calls.

Running
```python
print(timeit.timeit(lambda: next(_id_generator), number=10000000))
print(timeit.timeit(lambda: randint(0, 999999), number=10000000))
```
shows that the new generator is about 8 times faster than the old one (0.381s vs 3.056s).

### DIFF
--- a/rich/style.py
+++ b/rich/style.py
@@ -1,8 +1,9 @@
 import sys
 from functools import lru_cache
+from itertools import count
 from operator import attrgetter
 from pickle import dumps, loads
-from random import randint
+from random import getrandbits
 from typing import Any, Dict, Iterable, List, Optional, Type, Union, cast
 
 from . import errors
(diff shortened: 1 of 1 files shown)


akx:
Rebased on 14.2. Gentle review request nudge, @willmcgugan?

akx:
Codespell being red is unrelated.

akx:
Gentle nudge, @willmcgugan? 😄
Would likely make Toad a tiny bit faster too 🐸

PR #3855, opened 2025-09-26
Fix duplicate height attribute in ConsoleOptions

sunnynguyen-ai:
Fix naming conflict in ConsoleOptions dataclass where 'height' attribute was defined twice - once inherited from ConsoleDimensions and once as a direct attribute.

Renamed the direct attribute to 'render_height' to distinguish:
- size.height: int (console dimensions)
- render_height: Optional[int] (rendering constraint)

Updated all related methods:
- update()
- update_height()
- reset_height()
- update_dimensions()

This resolves type confusion and potential runtime errors.

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -118,7 +118,7 @@ class ConsoleDimensions(NamedTuple):
 @dataclass
 class ConsoleOptions:
     """Options for __rich_console__ method."""
-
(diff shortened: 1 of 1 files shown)


willmcgugan:
This doesn't do anything useful.

PR #3861, opened 2025-10-09
bump for Python3.14

willmcgugan:
Update meta for Python3.14


TomJGooding:
@willmcgugan I just spotted it looks like you need to mark 14.2.0 as the latest release on GitHub?

willmcgugan:
Done. Don't know why that doesn't happen by default for new releases now. I swear it used to.

TomJGooding:
Apparently GitHub did change the latest release behaviour but going back a few years ago now! https://github.blog/changelog/2022-10-20-explicitly-set-the-latest-release/

willmcgugan:
It is the default in Textual. I wonder if there is a setting for that somewhere. I lack the stamina to hunt for it.

TomJGooding:
Hm, strange. I had a quick look through the settings but couldn't see anything obvious. The [releases docs](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository) do mention that "If you do not select this option, the latest release label will automatically be assigned based on semantic versioning", but I don't see how that explains the difference with Textual since Rich also uses semvar. I don't use the GitHub releases myself so I'm afraid I can't offer much help!

PR #3863, opened 2025-10-11
feat: Add custom pretty printing for pathlib.Path objects

Anandraj-prog:
### Summary

This PR improves the pretty-printing of `pathlib.Path` objects by rendering them as strings instead of their default object representation.

### Problem

Currently, `rich.pretty.pprint` displays a `pathlib.Path` object using its default `__repr__`, which looks like `PosixPath('/path/to/file.txt')`. While technically correct, this is verbose and doesn't stand out in complex outputs where file paths are common.

### Solution

The changes in this PR intercept `pathlib.Path` objects during the rendering process and convert them to their string equivalent.

**Before:**
`PosixPath('my_folder/my_script.py')`

**After:**
`'my_folder/my_script.py'` (This will be colored as a string by the highlighter)

### DIFF
--- a/rich/pretty.py
+++ b/rich/pretty.py
@@ -1,5 +1,6 @@
 import builtins
 import collections
+import pathlib
 import dataclasses
 import inspect
 import os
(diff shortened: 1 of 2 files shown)


willmcgugan:
This will confuse people if Path instances appear to be strings. There is a reason the default repr is the way it is..

PR #3869, opened 2025-10-21
docs: add quickstart for contributors (venv, install, pytest)

imaanedhi:
Adds a short "Quickstart for Contributors" section in the README to make setup faster. It includes virtual environment creation, dependency installs, and pytest commands so new contributors can start testing immediately.


willmcgugan:
I can't have a section tagged on to the end of the README without context.

PR #3901, opened 2025-12-02
Add enable_link_path parameter to disable OSC 8 hyperlinks - Addresse…

### LINKED ISSUE #3896: [REQUEST] Jupyter-safe "strict text" mode for `Console`

I would like to request a supported option for `rich.console.Console` that forces strict plain-text output in Jupyter notebooks while retaining the structured features of `Console.log` such as time and path columns.

The key capability would be the ability to disable all HTML rendering, ANSI styling, highlighting, and especially OSC 8 hyperlinks, without disabling the display of the file path itself.

Something like:

```python
console = Console(
    force_jupyter=False,     # use text output
    strict_text=True,        # no HTML, no ANSI, no hyperlinks
    log_time=True,
    log_path=True
)
```

Expected output: clean single-spaced text logs in Jupyter, with the time and file path shown as plain ASCII text. No HTML blocks, no visible escape sequences, and no OSC 8 hyperlinks.

Currently the only way to get this behavior is to monkey-patch the private `_log_render` method to force `link_path=None`, which is brittle:

```python
_original = console._log_render

def _nologlinks(self, *args, **kwargs):
    kwargs["link_path"] = None
    return _original(*args, **kwargs)

console._log_render = types.MethodType(_nologlinks, console)
```

In Jupyter, rich’s HTML output produces large vertical spacing between `console.log` calls because each call becomes a separate `<pre>` block with notebook CSS margins.
Forcing terminal output (`force_jupyter=False`) produces ANSI and OSC 8 hyperlink escape sequences that Jupyter does not interpret correctly, resulting in visible escape codes or duplicated-looking lines.

(The OSC 8 hyperlinks are not useful anyway in remote-server notebook environments because the paths do not exist on the client machine.)

Thanks!

SergeyNoritsyn:
**Title:**
```
Add enable_link_path parameter to disable OSC 8 hyperlinks
```

**Description:**
```markdown
## Problem
Fixes #3896

In Jupyter notebooks and remote server environments, OSC 8 hyperlinks generated by `console.log()` cause issues:
- Large vertical spacing in Jupyter due to HTML block rendering
- Visible escape sequences when using `force_jupyter=False`
- Hyperlinks are not useful when file paths don't exist on client machines

Previously, the only workaround was to 'monkey-patch' the private `_log_render` method.

## Solution
Added a new `enable_link_path` parameter to the `Console` class:
- Defaults to `True` (backward compatible - existing behavior unchanged)
- When set to `False`, disables OSC 8 hyperlink generation
- File path text is still displayed (only the escape codes are removed)

## Example Usage
```python

# For Jupyter notebooks or remote servers
console = Console(
    force_jupyter=False,
    enable_link_path=False,  # Disable hyperlinks
    log_time=True,
    log_path=True
)

console.log("Clean output without OSC 8 escape sequences")
```

## Testing
- Added `test_enable_link_path()` - verifies hyperlinks are disabled
- Added `test_enable_link_path_default()` - verifies default behavior unchanged
- All existing tests pass

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -611,6 +611,7 @@ class Console:
         highlight (bool, optional): Enable automatic highlighting. Defaults to True.
         log_time (bool, optional): Boolean to enable logging of time by :meth:`log` methods. Defaults to True.
         log_path (bool, optional): Boolean to enable the logging of the caller by :meth:`log`. Defaults to True.
+        enable_link_path (bool, optional): Boolean to enable terminal hyperlinks in file paths logged by :meth:`log`. Defaults to True.
         log_time_format (Union[str, TimeFormatterCallable], optional): If ``log_time`` is enabled, either string for strftime or callable that formats the time. Defaults to "[%X] ".
         highlighter (HighlighterType, optional): Default highlighter.
         legacy_windows (bool, optional): Enable legacy Windows mode, or ``None`` to auto detect. Defaults to ``None``.
@@ -648,6 +649,7 @@ class Console:
         highlight: bool = True,
         log_time: bool = True,
         log_path: bool = True,
+        enable_link_path: bool = True,
         log_time_format: Union[str, FormatTimeCallable] = "[%X]",
         highlighter: Optional["HighlighterType"] = ReprHighlighter(),
         legacy_windows: Optional[bool] = None,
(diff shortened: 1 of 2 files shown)


willmcgugan:
Adding another parameter to the constructor is a red-line. There are too many already. 

Please do try and format your PR descriptions to make them readable.

PR #3908, opened 2025-12-05
Add tqdm compatibility shim with Rich progress backend

K2:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [X] Tests
- [ ] Other

## AI?

- [X] AI was used to generate this PR

## Checklist

## Description

Please describe your changes here.

I noticed https://github.com/Textualize/rich/discussions/2651 and other people asking for postfix support, I would usually use progress callbacks if a library supported them but then occasionally there's one w/o any standard way to keep the look consistent.  One of the tests I supplied can be pointed at a tqdm checkout and use the test cases from there, all pass except for some formatting ones, clearly that would require re-implementing the whole thing almost.

I figured you might want this as a 3rd party just let me know, since taking it in would give you at least a recommended way to wrap over some other package.  In any event, you can at least pass in a full Console to get the bar to use the placement you'd prefer it also passes one back (maybe the same one) that it used. 

Implements a runtime adapter that replaces tqdm imports with Rich's progress bars via install_tqdm(). The shim preserves common tqdm APIs (iteration, len(), postfix, write, wrapattr) while routing rendering through rich.progress.Progress.

Features:
- Opt-in monkeypatching of tqdm/trange and notebook variants
- Shared global Progress instance for performance
- Render throttling via mininterval/miniters/maxinterval
- Simplified text backend for file outputs (partial mode)
- Postfix value caching to avoid redundant formatting

Includes runnable examples demonstrating basic replacement and aggressive reference overwriting, plus comprehensive unit tests covering adapter installation, throttling, and postfix handling.

Updates README and contributor list.

**Important:** Link to an issue or discussion regarding these changes.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -41,6 +41,7 @@ The following people have contributed to the development of Rich:
 - [Hedy Li](https://github.com/hedythedev)
 - [Henry Mai](https://github.com/tanducmai)
 - [Luka Mamukashvili](https://github.com/UltraStudioLTD)
+- [Shane K2 Macaulay](https://github.com/K2)
 - [Alexander Mancevice](https://github.com/amancevice)
 - [Will McGugan](https://github.com/willmcgugan)
 - [Paul McGuire](https://github.com/ptmcg)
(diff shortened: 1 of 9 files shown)


willmcgugan:
tqdm already support a Rich backend. If that is your goal.

Regardless, this is way too much code to support. Thanks, but I can't accept this.

PR #3911, opened 2025-12-08
 Don't use windows legacy terminal support when ctypes is not available

youknowone:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## AI?

## Checklist

## Description

Hello, I am a developer of RustPython project. RustPython currently doesn't have ctypes, but trying to support pip.
rich is currently bundled in pip. When pip is trying to use rich on legacy windows terminal, it requires ctypes and raising error.
This patch changes rich not to turn on legacy windows terminal supports when ctypes is not available.

Could you check if this is an acceptable approach for rich? Then I will add a test to prevent regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Changed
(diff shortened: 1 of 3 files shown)


willmcgugan:
It does mean that it won't work on legacy windows. If RustPython only works on more modern Windows, then that might be an acceptable solution.

Although I think I would prefer a more explicit check. i.e. Check for RustPython rather than the absence of ctypes. Is that possible?

Issue #3432, opened 2024-07-18
[BUG] rich.inspect() does not appear to honor the "sort" keyword argument


rengrub:
- [x] I've checked [docs](https://rich.readthedocs.io/en/latest/introduction.html) and [closed issues](https://github.com/Textualize/rich/issues?q=is%3Aissue+is%3Aclosed) for possible solutions.
- [x] I can't find my issue in the [FAQ](https://github.com/Textualize/rich/blob/master/FAQ.md).

**Describe the bug**

Performing rich.inspect() on a class object with sort=False and sort=True results in the same output for class attributes which are not already naturally in alphabetical order (i.e. both appear sorted alphabetically).

```
import rich
class Test:
    def __init__(self):
        self.c = 1
        self.b = 2
        self.a = 3

test = Test()
rich.inspect(test)
```
Result:
![image](https://github.com/user-attachments/assets/998534fe-a858-4ef9-9a50-1d793b821712)

```
import rich
class Test:
    def __init__(self):
        self.c = 1
        self.b = 2
        self.a = 3

test = Test()
rich.inspect(test, sort=False)
```
Result:
![image](https://github.com/user-attachments/assets/6b724d8e-090c-4eec-8925-dc5593f7ca6b)

**Platform**
Python 3.11.3
WSL: Ubuntu-22.04

If you're using Rich in a terminal:

```
python -m rich.diagnose
pip freeze | grep rich
```

If you're using Rich in a Jupyter Notebook, run the following snippet in a cell
and paste the output in your bug report.

```python
from rich.diagnose import report
report()
```
```
╭────────────────────── <class 'rich.console.Console'> ──────────────────────╮
│ A high level console interface.                                            │
│                                                                            │
│ ╭────────────────────────────────────────────────────────────────────────╮ │
│ │ <console width=115 ColorSystem.TRUECOLOR>                              │ │
│ ╰────────────────────────────────────────────────────────────────────────╯ │
│                                                                            │
│     color_system = 'truecolor'                                             │
│         encoding = 'utf-8'                                                 │
│             file = <ipykernel.iostream.OutStream object at 0x7f96e251ab90> │
│           height = 100                                                     │
│    is_alt_screen = False                                                   │
│ is_dumb_terminal = False                                                   │
│   is_interactive = False                                                   │
│       is_jupyter = True                                                    │
│      is_terminal = False                                                   │
│   legacy_windows = False                                                   │
│         no_color = False                                                   │
[...]

PR #3913, opened 2025-12-13
Fix inspect() not respecting sort=False for instance attributes

Nabeelshar:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## AI?

## Checklist

## Description

Fixed issue #3432 where `inspect()` was not respecting the `sort=False` parameter for instance attributes.

### Problem
When calling `rich.inspect(obj, sort=False)`, the attributes were still displayed in alphabetically sorted order instead of preserving their insertion order.

### Root Cause
The `_render()` method was using `dir()` to get the list of attributes. Since Python's `dir()` always returns attributes in sorted order, even when `sort=False` was specified, the attributes were already sorted before the sorting logic was applied.

### Solution
Modified the `_render()` method in `rich/_inspect.py` to:
1. When `sort=False` and the object has a `__dict__`, use `vars(obj)` to get instance attributes in their insertion order (Python 3.7+ guarantees dict insertion order)
2. Combine instance attributes (in insertion order) with class attributes from `dir()`
3. Only apply alphabetical sorting when `sort=True` is explicitly set

### Changes
- Modified `_render()` method in `rich/_inspect.py`
- Added `test_inspect_sort_parameter()` test case
- Updated CHANGELOG.md
- Updated CONTRIBUTORS.md

### Testing
- All existing tests pass (42 passed, 4 skipped)
- New test specifically validates that `sort=False` preserves insertion order and `sort=True` sorts alphabetically
- Manual testing confirms the fix works as expected

**Important:** Fixes https://github.com/Textualize/rich/issues/3432

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Fixed `inspect()` not respecting `sort=False` parameter for instance attributes https://github.com/Textualize/rich/issues/3432
+
 ## [14.2.0] - 2025-10-09
 
 ### Changed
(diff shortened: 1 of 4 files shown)


willmcgugan:
Thanks, but this is not needed

```
sort (bool, optional): Sort attributes alphabetically, callables at the top, leading and trailing underscores ignored. Defaults to True.
```

PR #3914, opened 2025-12-14
Fix inspect(sort=False) to preserve instance attribute order

ye-soling:
This PR fixes an issue where `inspect(sort=False)` did not preserve the
insertion order of instance attributes.

Previously, instance attributes were always displayed in alphabetical order
because `dir(obj)` returns a sorted list of attribute names. This behavior
ignored the `sort=False` option.

To fix this, when `sort=False` and the object has a `__dict__`, the instance
attribute insertion order is used instead of `dir(obj)`.

A test was added to ensure the attribute order is preserved when `sort=False`.

### DIFF
--- a/rich/_inspect.py
+++ b/rich/_inspect.py
@@ -133,7 +133,12 @@ class Inspect(JupyterMixin):
                 return (error, None)
 
         obj = self.obj
-        keys = dir(obj)
+        if not self.sort and hasattr(obj, "__dict__"):
+            # Preserve instance attribute insertion order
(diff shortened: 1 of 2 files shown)


ye-soling:
Hi! Thanks for reviewing this PR 🙂
This fixes the issue where inspect(sort=False) did not preserve
instance attribute order and includes a test for the fix.
Please let me know if any changes are needed.

willmcgugan:
I can't imagine why anyone would need this. I would need a use-case before I could accept such a change.

PR #3962, opened 2026-02-08
Fix console.print() with empty args to respect end parameter

### LINKED ISSUE #3701: [BUG] Behavior of console.print() with empty *args is different from regular print()
**Describe the bug**

Expected that `console.print(end="!")` will print only `!` as regular `python print()`.

```python

console = Console()

for f in (
    lambda: print("A", end="!"),
    lambda: print(end="!"),
    lambda: console.print("A", end="!"),
    lambda: console.print(end="!"),
):
    print("=>", end="")
    f()
    print("<=")
```

Output:
```
❯ python3 test.py
=>A!<=
=>!<=
=>A!<=
=>
<=
```

rich == 14.0.0

veeceey:
## Summary

Fixes #3701

When `console.print()` was called with no arguments but a custom `end` parameter, it would print a newline instead of the end character. This was because empty objects were always replaced with `NewLine()` regardless of the `end` parameter.

## Changes

- Modified `Console.print()` in `rich/console.py` to only use `NewLine()` when `end` is the default `"\n"`
- When `end` is a custom value, use an empty `Text` object that will properly render with the custom end character
- Added test `test_print_empty_with_end()` to verify the fix

## Test plan

- [x] Added regression test that verifies end parameter works with empty args
- [x] All existing console print tests pass
- [x] Manual testing confirms behavior matches regular `print()`

## Before
```python
console = Console()

print(end="!")  # Prints: !
console.print(end="!")  # Prints: \n (newline)
```

## After
```python
console = Console()

print(end="!")  # Prints: !
console.print(end="!")  # Prints: ! (correctly)
```

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -1683,7 +1683,13 @@ class Console:
             new_line_start (bool, False): Insert a new line at the start if the output contains more than one line. Defaults to ``False``.
         """
         if not objects:
-            objects = (NewLine(),)
+            # When objects is empty, only print newline if end is "\n" (default)
+            # Otherwise, just print the end character (matching print() behavior)
+            if end == "\n":
+                objects = (NewLine(),)
+            else:
+                # Create empty object so that only the end character is printed
+                objects = (Text(""),)
 
         if soft_wrap is None:
             soft_wrap = self.soft_wrap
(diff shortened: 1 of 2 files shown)


veeceey:
Code changes verified ✓ Mergeable: YES | Ready for maintainer review.

PR #3969, opened 2026-02-10
Add hide parameter to Traceback

reteps:
## Summary
- Adds a `hide` parameter to `Traceback`, `Traceback.from_exception()`, `install()`, and `Console.print_exception()`
- Like `suppress` (which dims frames and hides source), `hide` removes matching frames entirely from the output
- Uses the same path-matching pattern as `suppress` for consistency

## Test plan
- [x] `test_hide`: Verifies storage of modules and string paths (mirrors `test_suppress`)
- [x] `test_hide_removes_frames`: Verifies hidden frames are absent from output while exception type is preserved
- [x] All existing traceback tests pass
- [x] mypy strict mode passes
- [x] black formatting clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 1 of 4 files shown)


reteps:
I did not see any policy for AI-assisted contributions, but please let me know if I should close this. This is a feature I monkey-patched for my own needs I thought might be useful to others, so I am upstreaming it. Thanks!

reteps:
In my case, I have several pieces of internal code I don't want users to have to sift through / be aware of.

PR #3978, opened 2026-02-17
Fix exception notes duplicated across chained exceptions

### LINKED ISSUE #3960: [BUG] Traceback `__notes__` from outermost exception shown on all exceptions in chain
## Description

When displaying a chained exception where only the outermost exception has `__notes__` (via PEP 678 `add_note`), Rich incorrectly displays those notes on **every** exception in the chain, not just the one that has them. Investigation done by Claude.

## Minimal reproduction

```python

try:
    try:
        raise ValueError("original error")
    except ValueError as exc:
        raise RuntimeError("wrapped error") from exc
except RuntimeError as exc:
    exc.add_note("This note was added to RuntimeError only")
    Console(width=80).print_exception()
```

**Rich output** (note appears on both exceptions):

```
ValueError: original error
[NOTE] This note was added to RuntimeError only    <-- should not be here

The above exception was the direct cause of the following exception:

RuntimeError: wrapped error
[NOTE] This note was added to RuntimeError only    <-- correct
```

**Python native output** (note appears only on RuntimeError, as expected):

```
ValueError: original error

The above exception was the direct cause of the following exception:

RuntimeError: wrapped error
This note was added to RuntimeError only
```

## Cause

In `Traceback.extract()`, the `notes` are read once from the initial `exc_value` before the `while True` loop, then the same list is reused for every `Stack` in the chain:

```python
notes: List[str] = getattr(exc_value, "__notes__", None) or []  # read once

while True:
    stack = Stack(
        ...
        notes=notes,  # same reference for ALL stacks
    )
    # ... walks __cause__ / __context__ chain ...
```

The fix would be to read `notes` from each `exc_value` inside the loop:

```python
while True:
    notes: List[str] = getattr(exc_value, "__notes__", None) or []
    stack = Stack(
        ...
        notes=notes,
    )
```

## Environment

- Rich 14.3.2
- Python 3.13

worksbyfriday:
## Summary

Fixes #3960.

When displaying chained exceptions where only the outermost exception has `__notes__` (PEP 678), Rich incorrectly shows those notes on **every** exception in the chain. This is because `notes` is read once before the extraction loop and the same list is reused for all `Stack` objects.

## Changes

Moved the `notes = getattr(exc_value, "__notes__", ...)` assignment from before the `while True` loop to inside it, so each exception in the chain reads its own `__notes__`.

## Before (bug)

```
ValueError: original error
[NOTE] This note was added to RuntimeError only    ← wrong

The above exception was the direct cause of the following exception:

RuntimeError: wrapped error
[NOTE] This note was added to RuntimeError only    ← correct
```

## After (fix)

```
ValueError: original error

The above exception was the direct cause of the following exception:

RuntimeError: wrapped error
[NOTE] This note was added to RuntimeError only    ← only here
```

### DIFF
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -467,8 +467,6 @@ class Traceback:
 
         from rich import _IMPORT_CWD
 
-        notes: List[str] = getattr(exc_value, "__notes__", None) or []
-
         grouped_exceptions: Set[BaseException] = (
             set() if _visited_exceptions is None else _visited_exceptions
         )
@@ -481,6 +479,7 @@ class Traceback:
                 return "<exception str() failed>"
 
         while True:
+            notes: List[str] = getattr(exc_value, "__notes__", None) or []
             stack = Stack(
                 exc_type=safe_str(exc_type.__name__),
                 exc_value=safe_str(exc_value),
--- a/tests/test_traceback.py
+++ b/tests/test_traceback.py
@@ -397,3 +397,23 @@ def test_recursive_exception() -> None:
             console.print_exception(show_locals=True)
 
     bar()
+
+
+@pytest.mark.skipif(
+    sys.version_info.minor < 11, reason="Not supported before Python 3.11"
+)
+def test_notes_chained_exceptions() -> None:
+    """Check notes are only shown on the exception that has them, not all in the chain."""
+    try:
+        try:
+            raise ValueError("original error")
+        except ValueError as ve:
+            raise RuntimeError("wrapped error") from ve
+    except RuntimeError as error:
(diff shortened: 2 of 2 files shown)


reteps:
Duplicate of https://github.com/Textualize/rich/pull/3964

worksbyfriday:
Thanks for noting that! PR #3964 was closed (not merged) on Feb 11 without any review comments — it appears the author closed it themselves. This PR addresses the same issue with the same approach (moving the `notes` extraction inside the loop), plus a regression test. Happy to defer if #3964 gets reopened.

worksbyfriday:

PR #3984, opened 2026-02-17
Fix Console.input() ignoring end attribute of Text prompt

### LINKED ISSUE #3643: [BUG] rich.Console.input method fails to propagate the "end" attribute of passed "prompt" TextType

**Describe the bug**

As stated in the title. 

The "input" method of the Console class in rich.console.py fails to propagate the "end" attribute of the prompt variable passed in as one of the methods inputs when calling self.print()

It instead ignores the value of prompt.end and instead passes end="" to self.print()

```py
def input(
    self,
    prompt: TextType = "",
    *,
    markup: bool = True,
    emoji: bool = True,
    password: bool = False,
    stream: Optional[TextIO] = None,
) -> str:
    """Displays a prompt and waits for input from the user. The prompt may contain color / style.

    It works in the same way as Python's builtin :func:`input` function and provides elaborate line editing and history features if Python's builtin :mod:`readline` module is previously loaded.

    Args:
        prompt (Union[str, Text]): Text to render in the prompt.
        markup (bool, optional): Enable console markup (requires a str prompt). Defaults to True.
        emoji (bool, optional): Enable emoji (requires a str prompt). Defaults to True.
        password: (bool, optional): Hide typed text. Defaults to False.
        stream: (TextIO, optional): Optional file to read input from (rather than stdin). Defaults to None.

    Returns:
        str: Text read from stdin.
    """
    if prompt:
        self.print(prompt, markup=markup, emoji=emoji, end="")
```

worksbyfriday:
## Summary

Fixes #3643.

`Console.input()` hardcodes `end=""` when printing the prompt, ignoring any custom `end` attribute set on a `Text` object:

```python
# Current behavior: "enter value" (end=": " is ignored)
# Expected behavior: "enter value: "
prompt = Text("enter value", end=": ")
console.input(prompt=prompt)
```

The fix reads the prompt's `end` attribute via `getattr`. If `end` is the default `"\n"`, it uses `""` (preserving existing behavior — cursor stays on the prompt line). If `end` has been explicitly changed, that value is used.

### Before
```python
self.print(prompt, markup=markup, emoji=emoji, end="")
```

### After
```python
prompt_end = getattr(prompt, "end", "\n")
self.print(
    prompt,
    markup=markup,
    emoji=emoji,
    end="" if prompt_end == "\n" else prompt_end,
)
```

## Test plan
- Added `test_input_prompt_end` covering:
  - String prompt: `end=""` (unchanged default behavior)
  - Text prompt with custom `end=": "`: uses custom end value
  - Text prompt with default `end="\n"`: uses `""` (cursor stays on prompt line)
- Verified test fails without fix, passes with fix
- Existing `test_input` and `test_input_password` still pass

### DIFF
--- a/rich/console.py
+++ b/rich/console.py
@@ -2164,7 +2164,13 @@ class Console:
             str: Text read from stdin.
         """
         if prompt:
-            self.print(prompt, markup=markup, emoji=emoji, end="")
+            prompt_end = getattr(prompt, "end", "\n")
+            self.print(
+                prompt,
+                markup=markup,
+                emoji=emoji,
+                end="" if prompt_end == "\n" else prompt_end,
+            )
         if password:
             result = getpass("", stream=stream)
         else:
--- a/tests/test_console.py
+++ b/tests/test_console.py
@@ -409,6 +409,39 @@ def test_input_password(monkeypatch, capsys) -> None:
     assert user_input == "bar"
 
 
+def test_input_prompt_end(monkeypatch, capsys) -> None:
+    """Regression test for https://github.com/Textualize/rich/issues/3643
+
+    Console.input() should respect the end attribute of a Text prompt.
+    """
+
+    def fake_input(prompt=""):
+        console.file.write(prompt)
+        return "bar"
+
+    monkeypatch.setattr("builtins.input", fake_input)
+    console = Console()
+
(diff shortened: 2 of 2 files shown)


worksbyfriday:

PR #3985, opened 2026-02-17
Disable html_inline in Markdown to prevent text swallowing

### LINKED ISSUE #3565: [BUG] Text interpreted as html tag in markdown gets swallowed.

**Describe the bug**

If markdown contains code with text that could be interpreted as html (eg. using "<"  rather than entity escaping it), the text interpreted as a tag deleted.  Eg

    >>> md = Markdown("Do something if 0<foo or foo>10")
    >>> console.print(md)
    Do something if 010

Unless there's actual support for html, I think the html_inline rule should be disabled (or if this is used by something, add an option for markdown without html?).  Changing the first line of markdown.Markdown.__init__ to:

    parser = MarkdownIt().enable("strikethrough").enable("table").disable("html_inline")

Seems to resolve the issue.

**Platform**
<details>

Running on linux, rich 13.9.4, wezterm terminal.

</details>

worksbyfriday:
## Summary

Fixes #3565.

Rich's Markdown renderer silently drops text that looks like HTML tags. For example:

```python
>>> from rich.console import Console
>>> from rich.markdown import Markdown
>>> Console().print(Markdown("Do something if 0<foo or foo>10"))
Do something if 010
```

The `<foo or foo>` is parsed as an `html_inline` token by markdown-it, but Rich has no handler for inline HTML — the token is simply discarded.

The fix disables the `html_inline` rule in the MarkdownIt parser, since Rich does not render inline HTML. This was also suggested by the issue reporter.

### Before
```python
parser = MarkdownIt().enable("strikethrough").enable("table")
```

### After
```python
parser = (
    MarkdownIt()
    .enable("strikethrough")
    .enable("table")
    .disable("html_inline")
)
```

## Test plan
- Added `test_angle_brackets_not_swallowed` verifying `"0<foo or foo>10"` is preserved in output
- Verified test fails without fix (text swallowed), passes with fix
- All existing markdown tests pass (1 pre-existing failure in `test_inline_code` due to pygments version difference, unrelated)

### DIFF
--- a/rich/markdown.py
+++ b/rich/markdown.py
@@ -555,7 +555,12 @@ class Markdown(JupyterMixin):
         inline_code_lexer: str | None = None,
         inline_code_theme: str | None = None,
     ) -> None:
-        parser = MarkdownIt().enable("strikethrough").enable("table")
+        parser = (
+            MarkdownIt()
+            .enable("strikethrough")
+            .enable("table")
+            .disable("html_inline")
+        )
         self.markup = markup
         self.parsed = parser.parse(markup)
         self.code_theme = code_theme
--- a/tests/test_markdown.py
+++ b/tests/test_markdown.py
@@ -199,6 +199,17 @@ def test_table_with_empty_cells() -> None:
     assert result == expected
 
 
+def test_angle_brackets_not_swallowed():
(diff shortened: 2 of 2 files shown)


worksbyfriday:

PR #3991, opened 2026-02-17
Fix underscores clipped on last line of SVG export

### LINKED ISSUE #3576: [BUG] Underscores on the last captured line are clipped when exporting to SVG
Underscore  characters on the last captured line are *clipped* when exporting to SVG.

Easily reproducible (rich 13.9.4):

``` python
console = Console(record=True)
console.print("STRING_WITH_UNDERSCORES")
console.save_svg("test.svg")
```

When opening the SVG file in a browser or an image viewer, underscores look like white spaces: 

clipped-underscore-chars

Reading the SVG source, we can see that:
- the text string indeed contains `_` and not white spaces
- the underscores are actually clipped by the main `<cliPath>` (the one whose identifier ends with "clip-terminal"), giving it *a little more* height reveals the `_`   
 
expected-underscore-chars

In the test example:
-  the viewbox height is 74.4
- the terminal `<cliPath>` height is 23.4
-  the terminal group is translated to 41 on the vertical axis (as always) 

Adding 23.4 to 41, we're at 64.4 on the vertical axis, about 10px above the end of the viewbox. My 2 cents is that it is too much. 

Still my 2 cents:
- the height of the main `<rect>` (the one that sets the background) equals to ` (y + 1) * line_height + padding_top + padding_bottom`
- the height of the terminal group equals to `(y + 1) * line_height - 1` : I understand we don't have to take `padding_top` into account (we're translated bellow the padding on the vertical axis), but shouldn't we still add `padding_bottom` ?

``` diff
--- console.py
+++ console.py
@@ -2623,7 +2623,7 @@
             char_height=char_height,
             line_height=line_height,
             terminal_width=char_width * width - 1,
-            terminal_height=(y + 1) * line_height - 1,
+            terminal_height=(y + 1) * line_height - 1 + padding_bottom,
             width=terminal_width + margin_width,
             height=terminal_height + margin_height,
             terminal_x=margin_left + padding_left,
```

This is how I generated the SVG where the underscores aren't clipped. Which does not prove that my reasoning is correct, 8 may very well be a value that falls just right, and incidentally equal to `padding_bottom`.

BTW, thanks for the good work, [DTSh](https://github.com/dottspina/dtsh) could hardly be what it (humbly) is without this library.

worksbyfriday:
Fixes #3576

## Summary

Characters with descenders (underscore `_`, lowercase `g`, `y`, `p`, etc.) on the last line of SVG exports were being clipped. The terminal clip-path height was calculated as `(y + 1) * line_height - 1`, which didn't account for `padding_bottom`. This left no room for character descenders on the final line.

Added `padding_bottom` to the clip-path height, matching how the overall background `<rect>` already includes padding in its height calculation. As the [issue reporter noted](https://github.com/Textualize/rich/issues/3576), this is the exact fix needed.

## Test plan

- Updated `EXPECTED_SVG` in `test_console.py` to reflect the corrected clip-path height (23.4 → 31.4)
- All three SVG export tests pass (`test_export_svg`, `test_export_svg_specified_unique_id`, `test_save_svg`)


worksbyfriday:

Issue #3958, opened 2026-02-05
[BUG] Strings with ANSI escape sequences can cause `Console.print()` to hang


jmw182:
- [x] I've checked [docs](https://rich.readthedocs.io/en/latest/introduction.html) and [closed issues](https://github.com/Textualize/rich/issues?q=is%3Aissue+is%3Aclosed) for possible solutions.
- [x] I can't find my issue in the [FAQ](https://github.com/Textualize/rich/blob/master/FAQ.md).

**Describe the bug**

While testing a rich-click app I discovered an issue I first reported (including example code) there: https://github.com/ewels/rich-click/issues/325

I was inadvertently passing a string containing ANSI escape sequences (generated via rich-click) into `Console.print()`. Prior to the latest rich 14.3.2 this would print a mostly unformatted string including ANSI escape sequences (I understand that `Text.from_ansi()` would be required to handle the ANSI formatting).

With the latest "ZWJy release" 14.3.2, this now results in a hang as ANSI escape characters such as \x1b return a 0 from `cells.get_character_cell_size()`, resulting in an infinite loop in `cells.split_graphemes()`. Previously, \x1b would return a 1 and the string would print without hanging.

**Platform**
<details>
<summary>Click to expand</summary>

What platform (Win/Linux/Mac) are you running on? What terminal software are you using?

Windows 11, git-bash in Windows Terminal and VS Code Terminal. Python 3.14.3.

```shell
$ python -m rich.diagnose
╭───────────────────────── <class 'rich.console.Console'> ─────────────────────────╮
│ A high level console interface.                                                  │
│                                                                                  │
│ ╭──────────────────────────────────────────────────────────────────────────────╮ │
│ │ <console width=120 ColorSystem.TRUECOLOR>                                    │ │
│ ╰──────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                  │
│     color_system = 'truecolor'                                                   │
│         encoding = 'utf-8'                                                       │
│             file = <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'> │
│           height = 30                                                            │
│    is_alt_screen = False                                                         │
│ is_dumb_terminal = False                                                         │
│   is_interactive = True                                                          │
│       is_jupyter = False                                                         │
│      is_terminal = True                                                          │
│   legacy_windows = False                                                         │
│         no_color = False                                                         │
│          options = ConsoleOptions(                                               │
│                        size=ConsoleDimensions(width=120, height=30),             │
[...]

PR #4002, opened 2026-02-18
Fix `split_graphemes` infinite loop

jmw182:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## AI?

## Checklist

## Description

This fixes the issue I reported in #3958 in which strings including non-printable characters can potentially hang in `cells.split_graphemes` due to a regression in 14.3.2.

Since the original issue resulted in an infinite loop, I added a dev-dependency on [pytest-timeout](https://github.com/pytest-dev/pytest-timeout). If this is not allowed, the commit adding this dependency can easily be reverted. pytest-timeout will allow the test to timeout rather than hang indefinitely in the failure condition.

No AI was used for this pull request, but I did compare against some of the related AI-submitted PRs that were previously closed (as referenced in the issue).


TomJGooding:
> I added a dev-dependency on pytest-timeout. If this is not allowed, the commit adding this dependency can easily be reverted.

I _suspect_ adding this dependency wouldn't be merged anyway, but it looks like you actually added pytest-timeout to the main dependencies rather than dev?

jmw182:
Oops, good catch. I think I first added it manually and then tried it again using a poetry command, which I'm not very familiar with. I will fix it.

jmw182:
I'm not experienced with poetry, but in the last commit I added the dev dependency by manually adding pytest-timeout to the pyproject.toml (under `tool.poetry.dev-dependencies`) and then ran `poetry lock` to update the lock file.

If that process is incorrect or if the additional dev dependency is unacceptable, I can rebase to drop that commit.

willmcgugan:
Closing in favor of https://github.com/Textualize/rich/pull/4006

PR #4006, opened 2026-02-19
fix for infinite loop in split_graphemes

willmcgugan:
Fix for https://github.com/Textualize/rich/issues/3958


willmcgugan:
@jmw182 Feel free to kick the tires.

PR #4018, opened 2026-02-24
Fix Segment._split_cells heuristic bug for multi-cell characters

### LINKED ISSUE #3299: [BUG] Segment._split_cells doesn't handle non-unit characters well

DhyeyTeraiya:
Fixes #3299. The previous heuristic starting position could overshoot the cut point when segments contained a mix of single-cell and multi-cell (e.g. emoji) characters. Since the algorithm had no backtracking, it could return incorrect splits. This PR replaces the heuristic with a simple character-by-character scan, which is O(n) but guaranteed to be correct.

### DIFF
--- a/rich/segment.py
+++ b/rich/segment.py
@@ -127,33 +127,34 @@ class Segment(NamedTuple):
         if cut >= cell_length:
             return segment, _Segment("", style, control)
 
+        if cut <= 0:
+            return _Segment("", style, control), segment
+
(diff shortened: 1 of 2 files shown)


TomJGooding:
The added regression tests pass without these changes.

willmcgugan:
Suspected slop

PR #4023, opened 2026-02-26
fix: replace 7 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


willmcgugan:
It is test code. These changes are pointless.

PR #4024, opened 2026-02-26
Fix traceback __notes__ shown on all exceptions in chain

### LINKED ISSUE #3960: [BUG] Traceback `__notes__` from outermost exception shown on all exceptions in chain

rstar327:
## Summary
- Moves `__notes__` extraction inside the `while True` loop in `Traceback.extract()` so each exception in a chain gets its own notes
- Previously, notes were read once from the outermost exception and shared across all stacks in the chain

## Test plan
- Added `test_notes_chained_exceptions` that verifies notes only appear on the exception they were added to
- All existing traceback tests pass (`23 passed, 1 skipped` on Python 3.13)

Fixes #3960

### DIFF
--- a/rich/traceback.py
+++ b/rich/traceback.py
@@ -467,8 +467,6 @@ class Traceback:
 
         from rich import _IMPORT_CWD
 
-        notes: List[str] = getattr(exc_value, "__notes__", None) or []
-
         grouped_exceptions: Set[BaseException] = (
             set() if _visited_exceptions is None else _visited_exceptions
         )
@@ -481,6 +479,7 @@ class Traceback:
                 return "<exception str() failed>"
 
         while True:
+            notes: List[str] = getattr(exc_value, "__notes__", None) or []
             stack = Stack(
                 exc_type=safe_str(exc_type.__name__),
                 exc_value=safe_str(exc_value),
--- a/tests/test_traceback.py
+++ b/tests/test_traceback.py
@@ -375,6 +375,27 @@ def test_notes() -> None:
         assert traceback.trace.stacks[0].notes == ["Hello", "World"]
 
 
+@pytest.mark.skipif(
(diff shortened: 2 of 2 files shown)


willmcgugan:

PR #4054, opened 2026-04-02
Fix FileProxy.isatty() always returning False instead of delegating to proxied file

### LINKED ISSUE #4041: `FileProxy.isatty()` always returns `False` instead of delegating to the proxied file

`FileProxy` (used by `Live` when `redirect_stdout=True`) inherits `isatty()` from `io.TextIOBase`, which always returns `False`. This breaks code that checks `sys.stdout.isatty()` inside a `Live` context.

`FileProxy` already delegates `fileno()` (line 56-57) and has a generic `__getattr__` fallback (line 25-26), but `isatty()` is defined on `io.TextIOBase` in the MRO, so `__getattr__` is never reached for it.

### Reproduction

```python
import sys
from rich.live import Live

console = Console()
print("Before Live:", sys.stdout.isatty())  # True

with Live("hello", console=console, redirect_stdout=True):
    print("Inside Live:", sys.stdout.isatty())  # False (should be True)

print("After Live:", sys.stdout.isatty())  # True
```

### Impact

Any code running inside a `Live` context with `redirect_stdout=True` that branches on `isatty()` gets the wrong answer. This is particularly problematic for applications that use `Live` for streaming display while also executing callbacks/tools that check whether stdout is a terminal.

### Suggested fix

Add an `isatty()` method to `FileProxy` that delegates to the underlying file, consistent with how `fileno()` is already handled:

```python
def isatty(self) -> bool:
    return self.__file.isatty()
```

Andrew-most-likely:
Fixes #4041.

## Problem

`FileProxy` extends `io.TextIOBase`, whose `isatty()` unconditionally returns `False`. The `__getattr__` fallback that delegates unknown attributes to the wrapped file is never reached for `isatty()` because it is already resolved through the `io.TextIOBase` MRO.

This means any code that calls `sys.stdout.isatty()` inside a `Live` context (with `redirect_stdout=True`) gets the wrong answer — always `False` — even when the real stdout is a terminal.

## Fix

Add an explicit `isatty()` override that delegates to `self.__file.isatty()`, exactly like `fileno()` already does:

```python
def isatty(self) -> bool:
    return self.__file.isatty()
```

## Test

`test_isatty_delegates_to_proxied_file` in `tests/test_file_proxy.py` verifies that a `FileProxy` wrapping a tty-like file returns `True`, and one wrapping a non-tty returns `False`.

### DIFF
--- a/rich/file_proxy.py
+++ b/rich/file_proxy.py
@@ -55,3 +55,6 @@ class FileProxy(io.TextIOBase):
 
     def fileno(self) -> int:
         return self.__file.fileno()
+
+    def isatty(self) -> bool:
+        return self.__file.isatty()
--- a/tests/test_file_proxy.py
+++ b/tests/test_file_proxy.py
@@ -35,3 +35,22 @@ def test_new_lines():
     assert file.getvalue() == "-\n"
     file_proxy.flush()
     assert file.getvalue() == "-\n-\n"
+
+
+def test_isatty_delegates_to_proxied_file():
+    """FileProxy.isatty() must delegate to the underlying file.
+
+    io.TextIOBase.isatty() always returns False, so without an explicit
+    override the result would be wrong when the proxied file is a tty.
+    Regression test for https://github.com/Textualize/rich/issues/4041
+    """
+    tty_file = io.StringIO()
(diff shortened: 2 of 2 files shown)


willmcgugan:

PR #4063, opened 2026-04-05
Preserve inline code spans in markdown tables

### LINKED ISSUE #4038: [BUG] `TableDataElement.on_text` clobbers syntax-highlight spans for inline code

When using `Markdown` with `inline_code_lexer` and `inline_code_theme`, inline code (backticks) inside table cells loses its syntax highlighting and falls back to the default `markdown.code` style. The same inline code renders correctly in paragraphs, headings, and list items.

**Root cause:**

`TableDataElement.on_text` (markdown.py) unconditionally applies `context.current_style` over the text:

```python
def on_text(self, context: MarkdownContext, text: TextType) -> None:
    text = Text(text) if isinstance(text, str) else text
    text.stylize(context.current_style)
    self.content.append_text(text)
```

When `MarkdownContext.on_text` processes a `code_inline` token, it produces a `Text` object with per-character syntax-highlight spans. But `TableDataElement.on_text` then calls `text.stylize(context.current_style)` on that `Text` object, which overwrites all the highlight spans with the flat `markdown.code` style.

Compare with `TextElement.on_text`, which correctly avoids this:

```python
def on_text(self, context: MarkdownContext, text: TextType) -> None:
    self.text.append(text, context.current_style if isinstance(text, str) else None)
```

Here, the style is only applied when `text` is a plain string. When it's already a highlighted `Text` object, `None` is passed, preserving the spans.

**Fix:**

`TableDataElement.on_text` should follow the same pattern:

```python
def on_text(self, context: MarkdownContext, text: TextType) -> None:
    if isinstance(text, str):
        self.content.append(text, context.current_style)
    else:
        self.content.append_text(text)
```

**Repro:**

```python

md = Markdown(
    "| Col |\n|---|\n| `hello` |\n\n`hello`",
    inline_code_theme="monokai",
    inline_code_lexer="python",
)
Console().print(md)
```

The standalone `` `hello` `` renders with the monokai theme; the one inside the table does not.

monsonego:
## Summary
- preserve existing Text spans when Markdown table cells receive already-highlighted inline code
- add a regression test for TableDataElement so inline code highlighting is not flattened in table cells

## Testing
- python -m pytest tests/test_markdown.py -k table_data_element_preserves_existing_text_spans -q

Closes #4038

### DIFF
--- a/rich/markdown.py
+++ b/rich/markdown.py
@@ -331,9 +331,10 @@ class TableDataElement(MarkdownElement):
         self.justify = justify
 
     def on_text(self, context: MarkdownContext, text: TextType) -> None:
-        text = Text(text) if isinstance(text, str) else text
-        text.stylize(context.current_style)
-        self.content.append_text(text)
+        if isinstance(text, str):
+            self.content.append(text, context.current_style)
+        else:
+            self.content.append_text(text)
 
 
 class ListElement(MarkdownElement):
--- a/tests/test_markdown.py
+++ b/tests/test_markdown.py
@@ -73,7 +73,8 @@ import io
 import re
 
 from rich.console import Console, RenderableType
-from rich.markdown import Markdown
+from rich.markdown import Markdown, TableDataElement
+from rich.text import Span, Text
 
 re_link_ids = re.compile(r"id=[\d\.\-]*?;.*?\x1b")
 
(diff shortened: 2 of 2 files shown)


willmcgugan:

PR #4070, opened 2026-04-09
perf: reduce Console and RichHandler import time by deferring unused imports

KRRT7:
## E2E Results

| Import | master | This PR | Speedup |
|---|---|---|---|
| `from rich.console import Console` | 78.1ms | 52.2ms | **1.50x faster** |
| `from rich.logging import RichHandler` | 99.4ms | 56.9ms | **1.75x faster** |
| `import rich` | 18.2ms | 18.3ms | (already lean) |

## Summary

Defer module-level imports that are only needed in specific code paths, move annotation-only imports to `TYPE_CHECKING`, and remove dead code:

| File | Change | Savings |
|---|---|---|
| `logging.py` | Defer `Traceback` to `emit()` (only when `rich_tracebacks=True`) | ~20ms |
| `logging.py` | `from __future__ import annotations` + `TYPE_CHECKING` for `Console`, `ConsoleRenderable`, `Highlighter`, `FormatTimeCallable` | ~6ms |
| `logging.py` | Replace `pathlib.Path` → `os.path.basename` (also a minor runtime win) | ~4-5ms |
| `console.py` | Eliminate `import inspect`; replace `isclass` → `isinstance(x, type)`, `currentframe` → `sys._getframe` | ~10ms |
| `console.py` | Defer `pretty` to `Console.print()` | ~3-5ms |
| `console.py` | Defer `scope` to `Console.log()` (only when `log_locals=True`) | ~3-5ms |
| `console.py` | Defer `getpass` to `Console.input()` (only when `password=True`) | ~2ms |
| `console.py` | Defer `html.escape` and `zlib` to export methods | ~2.3ms |
| `console.py` | Remove dead `_svg_hash` function (unused since 113997ac, fixes latent NameError) | cleanup |
| `segment.py` | Remove dead `logging` import (`getLogger` assigned but never used) | ~2-3ms |
| `theme.py` | Defer `configparser` to `Theme.from_file()` | ~1.5ms |
| `syntax.py` | Move `Console`, `ConsoleOptions`, `JustifyMethod`, `RenderResult` to `TYPE_CHECKING` | eliminates console.py from syntax import chain |
| `protocol.py` | Replace `from inspect import isclass` → `isinstance(x, type)` | prepares for dataclasses removal |
| `repr.py` | Defer `import inspect` to `auto_rich_repr()` | prepares for dataclasses removal |

All deferred imports are still available at runtime — they're loaded when the code path that needs them is actually hit. The `TYPE_CHECKING` imports are annotation-only and never needed at runtime (files use `from __future__ import annotations`).

## Benchmarks

Environment: Standard_D2s_v5 (non-burstable), CPython 3.13.13, hyperfine --min-runs 30

### Per-module breakdown (this PR)

| Module | Time |
|---|---|
| `import rich` | 18.2ms |
| `Console` | 52.5ms |
| `RichHandler` | 56.4ms |
| `Syntax` | 64.7ms |
| `Traceback` | 92.5ms |
| `Markdown` | 103.0ms |

## Motivation

pip vendors Rich, and `Console`/`RichHandler` are imported on every `pip` invocation. These savings directly improve startup time for pip and every other tool that imports Rich.

## Testing

- `pytest tests/` — 952 passed, 25 skipped
- `mypy -p rich --no-incremental` — no issues
- `black --check` — passes

## Disclosure

This PR was developed with AI assistance (codeflash-agent). Benchmarked in a controlled environment (dedicated non-burstable Azure VM). All tests verified across multiple Python implementations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Defer import of `Traceback` in `logging.py` to reduce `RichHandler` import time
+- Remove module-level `import inspect` from `console.py`; replace `isclass` with `isinstance(x, type)` and `inspect.currentframe` with `sys._getframe`
+- Remove unused `logging` import from `segment.py`
+- Defer `pretty` import in `console.py` to `Console.print()`
+- Defer `scope` import in `console.py` to `Console.log()`
+- Defer `getpass` import in `console.py` to `Console.input()`
+- Defer `configparser` import in `theme.py` to `Theme.from_file()`
+- Replace `pathlib.Path` with `os.path.basename` in `logging.py`
+- Defer `html.escape` and `zlib` imports in `console.py` to export methods
+
 ## [14.3.3] - 2026-02-19
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Defer import of `Traceback` in `logging.py` to reduce `RichHandler` import time
-- Remove module-level `import inspect` from `console.py`; replace `isclass` with `isinstance(x, type)` and `inspect.currentframe` with `sys._getframe`
-- Remove unused `logging` import from `segment.py`
-- Defer `pretty` import in `console.py` to `Console.print()`
-- Defer `scope` import in `console.py` to `Console.log()`
-- Defer `getpass` import in `console.py` to `Console.input()`
-- Defer `configparser` import in `theme.py` to `Theme.from_file()`
-- Replace `pathlib.Path` with `os.path.basename` in `logging.py`
-- Defer `html.escape` and `zlib` imports in `console.py` to export methods
+- Improved import time with lazy loading https://github.com/Textualize/rich/pull/4070
 
 ## [14.3.3] - 2026-02-19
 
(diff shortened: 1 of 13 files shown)


rich/console.py
```diff
@@ -2353,6 +2354,9 @@ def export_svg(
                 ids). If not set, this defaults to a computed value based on the recorded content.
         """
 
+        import zlib
+        from html import escape
+
```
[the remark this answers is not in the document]
KRRT7: Good catch — `_svg_hash` is actually dead code. It was orphaned in 113997ac (the SVG export rewrite, May 2022) when the only caller (`classes_prefix` namespacing) was removed but the function definition was left behind. Deleted it entirely in the latest commit.

rich/console.py
```diff
@@ -1901,14 +1896,14 @@ def print_exception(
     @staticmethod
     def _caller_frame_info(
         offset: int,
-        currentframe: Callable[[], Optional[FrameType]] = inspect.currentframe,
+        currentframe: Callable[[], Optional[FrameType]] = sys._getframe,
```
willmcgugan: We can't use `sys._getframe`, give it is private, and the docs say this:

> CPython implementation detail: This function should be used for internal and specialized purposes only. It is not guaranteed to exist in all implementations of Python.

What I'd suggest is making `currentframe` default to None, and do the import inside this method.

I think the only reason it is exposed is for tests.

rich/console.py
```diff
@@ -1528,6 +1521,8 @@ def _collect_renderables(
         Returns:
             List[ConsoleRenderable]: A list of things to render.
         """
+        from .pretty import Pretty as _Pretty, is_expandable as _is_expandable
```
willmcgugan: Why import them with al alternate name?

CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
willmcgugan, line 8: These changes aren't user facing. I don't think we need to list them all.

Maybe condense them in to one entry. "improved import time by lazy loading modules"

willmcgugan (changes requested):
I'm going to make these changes. Currently on Thai time, so I expect to merge before you are awake.

willmcgugan (approved):

PR #4072, opened 2026-04-10
Fix incomplete support for stderr in Legacy Windows layer

### LINKED ISSUE #4071: [BUG] Legacy Windows layer always uses stdout handle, even when stderr stream is used

**Describe the bug**

rich's legacy Windows layer [always uses stdout for rendering](https://github.com/Textualize/rich/blob/master/rich/_win32_console.py#L362-L363) when one of the standard handles is provided, i.e. it outputs to stdout even with a Console created with `stderr=True`.

Below there's a simple code example you can run with to compare behaviour between stdout and stderr (after installing rich with `pip install rich==14.3.3`):
```
python example.py
python example.py --stderr
```

```py
# example.py
import argparse
import subprocess
import sys
import time

from rich.progress import track


def child_main(use_stderr):
	console = Console(stderr=use_stderr, legacy_windows=True, force_terminal=True)
	with console.status("Example"):
		for _ in track(range(25), console=console):
			time.sleep(0.5)


def parent_main(use_stderr):
	args = [sys.executable, sys.argv[0], "--child"]
	if use_stderr:
		args.append("--stderr")
	subprocess.run(
		args,
		stdout=subprocess.PIPE if use_stderr else None,
		stderr=subprocess.PIPE if not use_stderr else None,
	)


def main():
	parser = argparse.ArgumentParser()
	parser.add_argument("--child", action="store_true")
	parser.add_argument("--stderr", action="store_true")
	args = parser.parse_args()
	if args.child:
		child_main(args.stderr)
	else:
		parent_main(args.stderr)


if __name__ == "__main__":
	main()
```

Here's a short recording comparing the two:
https://github.com/user-attachments/assets/cee24968-5a4a-4d1a-aa35-fcefd26a208c

**Platform**
<details>

Windows 11 Command Prompt using conhost (i.e. legacy Windows)

```
(.venv) C:\Users\Jakub\work\rich-repro> python -m rich.diagnose
┌───────────────────────── <class 'rich.console.Console'> ─────────────────────────┐
│ ┌──────────────────────────────────────────────────────────────────────────────┐ │
│ │ <console width=119 ColorSystem.WINDOWS>                                      │ │
│ └──────────────────────────────────────────────────────────────────────────────┘ │
│     color_system = 'windows'                                                     │
│           height = 30                                                            │
│   legacy_windows = True                                                          │
│                        size=ConsoleDimensions(width=119, height=30),             │
│                        legacy_windows=True,                                      │
│                        max_width=119,                                            │
│                        max_height=30,                                            │
│             size = ConsoleDimensions(width=119, height=30)                       │
│            width = 119                                                           │
└──────────────────────────────────────────────────────────────────────────────────┘
┌─── <class 'rich._windows.WindowsConsoleFeatures'> ────┐
│ ┌───────────────────────────────────────────────────┐ │
│ └───────────────────────────────────────────────────┘ │
└───────────────────────────────────────────────────────┘
┌────── Environment Variables ───────┐
│     'JUPYTER_COLUMNS': None,       │
│     'JUPYTER_LINES': None,         │
│     'TERM': None,                  │
│     'TTY_COMPATIBLE': None,        │
│     'TTY_INTERACTIVE': None,       │
└────────────────────────────────────┘
platform="Windows"
(.venv) C:\Users\Jakub\work\rich-repro> pip freeze | grep rich
rich==14.3.3
```

</details>

Jackenmen:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## AI?

## Checklist

## Description

I've grepped through the code for argless calls to `GetStdHandle()` and, where appropriate, I've updated them to use the stderr handle (-12 per https://learn.microsoft.com/en-us/windows/console/getstdhandle#parameters), when operating on stderr.
Specifically:
- `Console` now detects terminal features for stderr correctly, when stdout is redirected (before, stderr was always treated as legacy Windows when stdout is redirected, even if stderr was attached to Windows Terminal, not conhost)
- `Console(stderr=True, legacy_windows=True)` now actually operates on stderr for features that require Windows Console API instead of stdout
- public `get_windows_console_features()` and `detect_legacy_windows()` functions in `rich.console` module now accept an optional `file` argument

I didn't write any tests because it does not seem like the functions/methods that I edited had any existing ones. I assume that means it's probably okay, and I don't exactly have a good idea of what would need to be tested here anyway.

You'll notice I moved the `legacy_windows` assignment to the end of the init function. This was needed because I had to be able to pass `self.file` to `detect_legacy_windows()`. The width, height, and color system detection depend on its value, so they had to be moved as well.

Fixes #4071

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Changed
+
+- `console.get_windows_console_features` and `console.detect_legacy_windows` now have an optional `file` parameter https://github.com/Textualize/rich/pull/4072
+
+### Fixed
+
+- Fixed the auto-detection of `Console.legacy_windows` for the stderr stream, when stdout is redirected https://github.com/Textualize/rich/pull/4072
+- Fixed legacy Windows rendering for the stderr stream https://github.com/Textualize/rich/pull/4072
+
 ## [14.3.3] - 2026-02-19
 
 ### Fixed
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -100,3 +100,4 @@ The following people have contributed to the development of Rich:
 - [Brandon Capener](https://github.com/bcapener)
 - [Alex Zheng](https://github.com/alexzheng111)
 - [Sebastian Speitel](https://github.com/SebastianSpeitel)
+- [Jakub Kuczys](https://github.com/Jackenmen)
--- a/rich/_win32_console.py
+++ b/rich/_win32_console.py
@@ -17,11 +17,26 @@ import time
 from ctypes import Structure, byref, wintypes
 from typing import IO, NamedTuple, Type, cast
 
+from rich._fileno import get_fileno
 from rich.color import ColorSystem
 from rich.style import Style
 
+try:
+    STDOUT_FILENO = sys.__stdout__.fileno()  # type: ignore[union-attr]
+except Exception:
+    STDOUT_FILENO = 1
+try:
+    STDERR_FILENO = sys.__stderr__.fileno()  # type: ignore[union-attr]
+except Exception:
+    STDERR_FILENO = 2
+
 STDOUT = -11
+STDERR = -12
 ENABLE_VIRTUAL_TERMINAL_PROCESSING = 4
+FILENO_TO_HANDLE = {
+    STDOUT_FILENO: STDOUT,
+    STDERR_FILENO: STDERR,
+}
 
 COORD = wintypes._COORD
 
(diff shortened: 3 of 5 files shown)


Jackenmen:
Sorry for a bit of churn, I struggled with appeasing mypy a bit...
This part in particular is a bit annoying, but I couldn't think of a better way to avoid mypy reporting an issue here: 
https://github.com/Textualize/rich/pull/4072/commits/b159cea2215cd4d2dbf3d21fac61a6eaabd62ee5

I guess strictly speaking, `STDERR_FILENO` and `STDOUT_FILENO` are used both in `console` and `_win32_console` module - let me know if I should just create a new module for them (module name suggestions welcome :smile:).

Ricky-7-Yan (changes requested):
Thanks for tackling the mixed stdout/stderr case. I found one blocking issue in the new feature lookup: `get_windows_console_features(file)` still uses the single module-level `_windows_console_features` cache, so the first stream probed determines the result for every later stream.

On the exact head `c81ebba8`, I mocked fd 1 as `vt=False` (redirected stdout) and fd 2 as `vt=True` (Windows Terminal stderr):

```text
order [1, 2] -> probe calls [1], results [False, False]
order [2, 1] -> probe calls [2], results [True, True]
```

Thus `Console(stderr=True)` can still inherit stdout's legacy result in the scenario this PR is intended to fix; the reverse construction order can also leak stderr's capabilities into stdout. Please key the cache by the resolved fileno (including the `None` fallback), or remove the cache, and add a regression test that probes stdout and stderr with distinct mocked feature results in both orders.

Verification: all five snapshot blobs match the PR head; the three related test modules pass (`136 passed, 7 skipped`), the remaining suite excluding the unchanged recursion module passes (`964 passed, 20 skipped`), and Black reports the three changed Python files clean.

PR #4078, opened 2026-04-12
fix: RichHandler now honours %z (and other timezone specifiers) in log_time_format

### LINKED ISSUE #3877: [BUG] RichHandler ignores `%z` time zone format specifier
RichHandler sseems to ignore or cut off the `%z` timezone (or uses time objects where that info is not attached). I.e. with `"log_time_format": "%Y-%m-%dT%H:%M:%S%z"`, for me, it prints:

```
2025-07-04T19:14:40 DEBUG    debug message                                                                                                                                           main.py:26
2025-07-04T19:14:41 INFO     info message                                                                                                                                            main.py:28
```

where it should be something like `2025-07-04T19:14:41+0200`.

DORI2001:
## Summary

Fixes #3877.

`RichHandler` creates the log timestamp with:

```python
log_time = datetime.fromtimestamp(record.created)
```

`datetime.fromtimestamp()` returns a **naive** `datetime` (no `tzinfo`).
When the format string contains `%z`, `strftime('%z')` on a naive datetime
returns an empty string, so the timezone offset is silently dropped.

## Fix

Call `.astimezone()` (no argument) on the naive result:

```python
log_time = datetime.fromtimestamp(record.created).astimezone()
```

`astimezone()` with no argument attaches the *local* timezone to a naive
datetime, making it timezone-aware without changing the displayed time.
`%z` then correctly emits the local UTC offset (e.g. `+0300` or `-0500`).

This is a one-character addition, requires no new imports, and is
backward-compatible — callers who don't use `%z` see no change.

## Test plan

- [x] New test `test_log_time_format_with_timezone` asserts that a format
  string containing `%z` produces a timestamp that includes a UTC offset
- [x] Existing logging tests still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rich/logging.py
+++ b/rich/logging.py
@@ -232,7 +232,7 @@ class RichHandler(Handler):
         path = os.path.basename(record.pathname)
         level = self.get_level_text(record)
         time_format = None if self.formatter is None else self.formatter.datefmt
-        log_time = datetime.fromtimestamp(record.created)
+        log_time = datetime.fromtimestamp(record.created).astimezone()
 
         log_renderable = self._log_render(
             self.console,
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -1,6 +1,7 @@
 import io
-import os
 import logging
+import os
+import re
 from typing import Optional
 
 import pytest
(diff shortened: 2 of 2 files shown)


willmcgugan:
/

PR #4084, opened 2026-04-19
Fix ConsoleOptions propagation

### LINKED ISSUE #4028: [BUG] Failure to propagate formatting overrides via ConsoleOptions

⚠️ **UPDATE: Revised Scope**
Upon further investigation, this is not limited to a missing `emoji` setting.

There are two core issues

1. `ConsoleOptions` is missing an `emoji` field. Therefore, `emoji` overrides passed to `console.print()` are ignored.
2. Multiple classes (`Table`, `Rule`, `Columns`, and `Measure`) fail to pass some or all of `emoji`, `markup`, and `highlight` settings from their `ConsoleOptions` back to `console.render_str()`.

I have a fix but I am currently unable to create a pull request in this repo.

See this fork for the fix:
https://github.com/kmvanbrunt/rich/tree/propagate_console_options

**Original Description**

When printing a complex renderable like a `Table`, passing `emoji=False` to `console.print()` does not disable emoji parsing within the table cells if the `Console` was initialized with `emoji=True` (the default).

This appears to be because `ConsoleOptions` does not include an `emoji` field, and `Console.render()` does not propagate the `emoji` preference to `render_str()`.

Here is code to reproduce the bug.

```Python
from rich.table import Table

# Create Console with markup and emoji enabled
console = Console(markup=True, emoji=True)
t = Table(show_header=False)
t.add_row("[blue]some text[/blue]", ":1234:")

print("Normal print() shows markup and emoji rendering")
console.print(t)

# Attempt to override markup and emoji when printing a Table
print("Print overrides only work for markup and the emoji still renders")
console.print(t, markup=False, emoji=False)
```

Here is a screenshot of the results.

Image

**Platform**
rich==14.3.3
Python 3.11.2
Debian 12.13

kmvanbrunt:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## AI?

## Checklist

## Description

Fixes #4028 

1. Added the `emoji` field to `ConsoleOptions`, allowing `emoji` overrides passed to `console.print()` to be propagated down the rendering stack.

2. Fixed several instances (`Table`, `Rule`, `Columns`, and `Measure`) where the renderables were calling `console.render_str()` but failing to pass the some or all of `emoji`, `markup`, and `highlight` settings from the active `ConsoleOptions`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- Added `emoji` field to `console.ConsoleOptions` https://github.com/Textualize/rich/issues/4028
+- Fixed `ConsoleOptions` propagation https://github.com/Textualize/rich/issues/4028
 
 ## [15.0.0] - 2026-04-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- Added `emoji` field to `console.ConsoleOptions` https://github.com/Textualize/rich/issues/4028
+
 ## [14.3.3] - 2026-02-19
 
 ### Fixed
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -76,6 +76,7 @@ The following people have contributed to the development of Rich:
 - [Aaron Stephens](https://github.com/aaronst)
 - [Karolina Surma](https://github.com/befeleme)
 - [Gabriele N. Tornetta](https://github.com/p403n1x87)
+- [Kevin Van Brunt](https://github.com/kmvanbrunt)
 - [Nils Vu](https://github.com/nilsvu)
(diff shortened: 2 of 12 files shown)


alokshukla631 (changes requested):
I think this is still incomplete on the `highlight` side.

`Columns`, `Rule`, `Table`, and `Measurement.get()` still call `render_str(..., highlight=False)`, so `console.print(..., highlight=True)` appears to remain ignored for nested strings even after this patch. For example, on this branch:

```python
from rich.columns import Columns

console = Console(force_terminal=True, color_system="truecolor", _environ={})
console.print(Columns(["123"]), highlight=True)
```

still emits plain `123`, while `console.print("123", highlight=True)` highlights it.

Could we propagate `options.highlight` through these calls as well, and add coverage for that path alongside the new emoji / markup tests?

kmvanbrunt:
@alokshukla631 

I honestly don't know the correct answer concerning `highlight`. `Table` and `Measure` were already setting it to `False` so I left those alone to preserve existing behavior.

For `Columns` and `Rule` I had to guess, so I based it on what I saw in `Table` and `Measure`.
If they are intended to support highlighting, then I of course need to pass along `options.highlight`.

@willmcgugan Can you provide some insight into this? Are any of these four `Renderables` supposed to support highlighting?

alokshukla631:
I took a quick look on current `master`: `console.print(Columns(["123"]), highlight=True)` and `console.print(Rule("123"), highlight=True)` both highlight the embedded string there, so I don't think matching the older `Table` / `Measure` behavior is the right baseline for those two.

The part that still seems ambiguous to me is whether `Table` / `Measurement` should also honor `options.highlight`, but at least `Columns` and `Rule` already do in current Rich.

PR #4092, opened 2026-04-23
feat(examples): safe_markup — escaping external content before Rich rendering

rudi193-cmd:
## Summary

Adds `examples/safe_markup.py` — a focused example of `rich.markup.escape` for safely displaying external or user-controlled content.

The existing examples show how to *create* Rich markup. This shows how to *safely consume* external data that may contain bracket characters (`[`, `]`) that would otherwise be interpreted as markup tags.

Covers:
- Basic `escape()` usage in `Panel`, table cells, and f-strings
- Why it matters: database records, API responses, filesystem paths, error messages
- `Console(highlight=True)` vs `markup.escape` distinction

This pattern comes from production use in an MCP server ([willow-mcp](https://github.com/rudi193-cmd/willow-mcp)) where tool responses from Postgres and SQLite frequently contain bracket characters that must be escaped before display.

### DIFF
--- a/examples/safe_markup.py
+++ b/examples/safe_markup.py
@@ -1,0 +1,56 @@
+"""
+Safe markup output — escaping user content before writing to a Rich console.
+
+When displaying data from external sources (API responses, tool output, database
+records) you must escape the content before passing it to Rich's markup parser,
+otherwise brackets in the data are interpreted as markup tags.
+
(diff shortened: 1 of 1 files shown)


willmcgugan:
No thank you. Please don't contribute any more examples.

PR #4110, opened 2026-05-01
Fix hyperlinks being split into multiple links when text is highlighted

### LINKED ISSUE #4109: [BUG] Hyperlinks are split into multiple links when text is highlighted

**Describe the bug**

Hyperlinks are split into multiple links when text is highlighted. When a hyperlink whose text contains tokens the ReprHighlighter recognises (UUID, path, filename), it currently renders it as multiple clickable links, instead of just one.

Expected:

Image

Video: https://github.com/user-attachments/assets/f9ac242b-e16d-446b-b736-67ee1977164c

Current behavior:

Image

Video: https://github.com/user-attachments/assets/d75aa3b3-145c-479f-bf8d-21f93bb859d2

**Platform**
<details>

What platform (Win/Linux/Mac) are you running on? What terminal software are you using?
Mac iTerm2

I may ask you to copy and paste the output of the following commands. It may save some time if you do it now.

If you're using Rich in a terminal:

```
python -m rich.diagnose
pip freeze | grep rich
```

If you're using Rich in a Jupyter Notebook, run the following snippet in a cell
and paste the output in your bug report.

```python
from rich.diagnose import report
report()
```

</details>

filbranden:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## AI?

I used Claude Sonnet 4.6 to work on this fix, it was fully reviewed by a human though. The process went roughly as follows:
- Claude initially suggested just dropping the `copy()` call, I told it there was probably a reason why it was there, so I asked it to figure that out and add a test that would ensure that behavior was preserved.
- Claude then suggested a lighter approach to creating the copied style, which ended in an uncached refactor of _add that would get called in that specific situation. This looked acceptable. I asked Claude to add appropriate tests.
- Then I realized if the issue was the lru_cache, we could fix it by passing the expected link_id as an argument, which would turn into a key for the lru_cache. In a sense, this seems more efficient because we don't need to copy or recreate or discard a style object. On the other hand, it might have side effects on the lru_cache itself, so I count on code review to confirm this is an appropriate approach.
- I also asked Claude to add a specific test involving the highlighter, since that's the specific issue I was trying to fix.

I understand there wasn't discussion in the issue itself, I hope this PR with the code might serve as a starting point for discussion with a proposed code change that I'm willing to iterate on. Thanks!

## Checklist

- [x] I accept that @willmcgugan may be pedantic in the code review. YES PLEASE!

## Description

`Style.__add__` was calling `copy()` on every combined style that contained a
    link, which generated a fresh link_id for each highlighted sub-span within
    a single link markup region. Terminals use link_id to identify hyperlink
    boundaries, so each sub-span appeared as a separate clickable link instead
    of one continuous one.

The underlying root cause is that `_add`'s lru_cache keys on `__hash__`/`__eq__`,
    which intentionally excludes link_id. Two Style objects with the same visual
    content but different link_ids therefore collide in the cache; copy() was
    a blunt workaround that overcorrected by regenerating the id on every hit.

Fix this by having `_add` take the link_id as an argument so that the lru_cache
    will now also key by link_id, preventing the issue that required the `copy()`
    workaround altogether.

Add test cases to confirm the issue is fixed and include regression tests to prevent it from getting broken again in the future.

Fixes #4109.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- hyperlinks being split into multiple links when text is highlighted https://github.com/Textualize/rich/pull/4110
+
 ## [15.0.0] - 2026-04-12
 
 ### Changed
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -101,3 +101,4 @@ The following people have contributed to the development of Rich:
 - [Alex Zheng](https://github.com/alexzheng111)
 - [Sebastian Speitel](https://github.com/SebastianSpeitel)
 - [Kevin Turcios](https://github.com/KRRT7)
+- [Filipe Brandenburger](https://github.com/filbranden)
--- a/rich/style.py
+++ b/rich/style.py
@@ -730,7 +730,7 @@ class Style:
         sys.stdout.write(f"{self.render(text)}\n")
 
     @lru_cache(maxsize=1024)
-    def _add(self, style: Optional["Style"]) -> "Style":
+    def _add(self, style: Optional["Style"], link_id: str) -> "Style":
         if style is None or style._null:
             return self
         if self._null:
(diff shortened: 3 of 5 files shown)


filbranden:
@willmcgugan Ping? Would you mind taking a look? This is a small visual annoyance but I'd still like to have it fixed, if possible, thanks!

filbranden:
@willmcgugan Would you mind taking a look? I think there's a considerable improvement in the way hyperlinks with uuids or paths look and behave, from:

Image

To:

Image

So I was hoping we could get this merged into `rich`.

Thank you!

PR #4118, opened 2026-05-12
Fix Segment._split_cells for non-unit-width characters (emoji, CJK)

### LINKED ISSUE #3299: [BUG] Segment._split_cells doesn't handle non-unit characters well

ShiLiguo123:
## Summary

Fix `Segment._split_cells()` to correctly handle strings containing double-cell-width characters (emoji, CJK characters, etc.).

## Root Cause

The old algorithm used a heuristic initial position: `pos = int((cut / cell_length) * len(text))`. This assumes uniform character width, which fails badly for mixed-width strings.

For example:
```py
s = Segment('🦊🦊🦊\n\n\n\n\n\n')
Segment._split_cells(s, 3)
# Wrong: (Segment('🦊🦊🦊\n '), Segment(' \n\n\n\n'))
```

## Fix

Replaced the heuristic + while-loop with a character-by-character walk-through. The new algorithm:
1. Tracks cumulative cell position as it iterates over characters
2. When it finds the character that exceeds the cut point, it splits there
3. Double-width characters are replaced with two spaces as before

Fixes #3299

### DIFF
--- a/rich/segment.py
+++ b/rich/segment.py
@@ -126,32 +126,25 @@ class Segment(NamedTuple):
 
         cell_size = get_character_cell_size
 
-        pos = int((cut / cell_length) * len(text))
+        # Walk through characters tracking cell position,
+        # avoids the heuristic + while-loop which fails on
+        # non-unit-width characters (emoji, CJK, etc.)
+        cell_pos = 0
+        for i, char in enumerate(text):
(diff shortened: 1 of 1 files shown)


willmcgugan:

PR #4127, opened 2026-05-17
Make BrokenPipeError test deterministic.

tjni:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Tests
- [ ] Other

## AI?

I used Codex with GPT-5.4 on Medium effort to investigate and generate local scripts to reproduce the flakiness as well as to create the initial version of the patch, but then I also manually read over and tweaked it to remove unnecessary lines such as certain comments and extra assertions.

## Checklist

## Description

`test_brokenpipeerror` seems to be flaky (albeit rarely). We can try to make it deterministic by writing more data during the test to reliably trigger BrokenPipeError.

I specifically ran into this while building `python3Packages.rich` from nixpkgs locally using Nix, and during research found that some other package managers have also disabled this test, e.g.:

- https://www.mail-archive.com/gentoo-commits%40lists.gentoo.org/msg1131334.html
- https://www.mail-archive.com/guix-commits%40gnu.org/msg356105.html

I had Codex generate locally some scripts to reproduce this, and it was able to do so against Python 3.13 after a few hundred iterations, and with these changes I was not able to reproduce it again locally, so I feel it could be worth a try.

I have at present skipped creating a discussion because I feel this falls under "clear bug-fixes / typos etc". I also refrained from updating the CHANGELOG, because it appeared to me that entries under there look like external facing changes. I'm happy to make any changes needed, however.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -48,6 +48,7 @@ The following people have contributed to the development of Rich:
 - [Paul McGuire](https://github.com/ptmcg)
 - [Antony Milne](https://github.com/AntonyMilneQB)
 - [Michael Milton](https://github.com/multimeric)
+- [Theodore Ni](https://github.com/tjni)
 - [Martina Oefelein](https://github.com/oefe)
 - [Joel Ostblom](https://github.com/joelostblom)
 - [Nathan Page](https://github.com/nathanrpage97)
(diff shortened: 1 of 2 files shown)


alyssais:
It fails every time for me.  This patch does fix it.

blitz:
> It fails every time for me. This patch does fix it.

Didn't test the patch yet, but the test fails for me every time as well.

PR #4133, opened 2026-05-21
fix readline support

florentx:
On Linux, I had the issue when using `Prompt.ask(...)` , `Confirm.ask(...)`  or `Console.input(...)`.
I import builtin `readline` to get line editing capability.

When navigating the line to the left with the arrow key, cursor goes to the start of the line and overwrite the prompt text.
This can be fixed, if the `input` built-in function receives the prompt text.

Also I've added a convenient helper (static method) to enable `readline`.


florentx:
@willmcgugan  it is a bit harsh to get a close notification without a comment.

Is there any policy violation? I did not find, sorry.

PR #4142, opened 2026-05-22
fix: ReprHighlighter greedy regex merges adjacent markup tags

devteamaegis:
## Problem

`ReprHighlighter` uses a regex to highlight `<tag>`-style markup in `repr()` output:

```python
r"(?P<tag_start><)(?P<tag_name>[-\w.: |]*)(?P<tag_contents>[\w\W]*)(?P<tag_end>>)"
```

The `tag_contents` group uses `[\w\W]*` — a greedy pattern that matches **any character including `>`**. When a string contains two or more adjacent tags, the pattern matches from the `<` of the first tag all the way to the `>` of the *last* tag, producing one giant span instead of independent highlights per tag.

Reported in #4035: strings like `<a>content1</a><b>content2</b>` (which appear in rich's own `repr()` output for nested renderables) produce a single merged span, preventing correct syntax colouring.

## Root cause

```
<a>content1</a><b>content2</b>
 ^                            ^
 tag_start                    tag_end   ← greedy [\w\W]* consumed everything in between
```

## Fix

Change `[\w\W]*` → `[^>]*` so `tag_contents` cannot consume a closing `>`:

```python
r"(?P<tag_start><)(?P<tag_name>[-\w.:|]*)(?P<tag_contents>[^>]*)(?P<tag_end>>)"
```

This preserves all existing behaviour (tag names, attributes, whitespace) while correctly stopping at the first `>`.

## Tests

Three regression tests are added to `tests/test_highlighter.py`:

- `test_repr_highlighter_does_not_merge_adjacent_tags` — verifies adjacent tags produce ≥ 2 independent `tag_start` spans
- `test_repr_highlighter_single_tag_unchanged` — verifies a lone tag is still highlighted correctly
- `test_repr_highlighter_tag_with_attributes` — verifies tags with attribute text (e.g. `<class name='Foo'>`) are still matched

### DIFF
--- a/rich/highlighter.py
+++ b/rich/highlighter.py
@@ -82,7 +82,7 @@ class ReprHighlighter(RegexHighlighter):
 
     base_style = "repr."
     highlights: ClassVar[Sequence[str]] = [
-        r"(?P<tag_start><)(?P<tag_name>[-\w.:|]*)(?P<tag_contents>[\w\W]*)(?P<tag_end>>)",
+        r"(?P<tag_start><)(?P<tag_name>[-\w.:|]*)(?P<tag_contents>[^>]*)(?P<tag_end>>)",
         r'(?P<attrib_name>[\w_]{1,50})=(?P<attrib_value>"?[\w_]+"?)?',
         r"(?P<brace>[][{}()])",
         _combine_regex(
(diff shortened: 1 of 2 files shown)


willmcgugan:

PR #3332, opened 2024-04-15
Add new column object: IterationSpeedColumn

chthollyphile:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Documentation / docstrings
- [ ] Tests
- [ ] Other

## Checklist

- [x] I've run the latest [black](https://github.com/psf/black) with default args on new code.
- [x] I've updated CHANGELOG.md and CONTRIBUTORS.md where appropriate.
- [x] I've added tests for new code.
- [x] I accept that @willmcgugan may be pedantic in the code review.

## Description

This pull request addresses the requirement raised in discussion #926  to add a new column for the number of iterations per second.

PR #4157, opened 2026-06-05
feat: add IterationSpeedColumn to progress

PierreLapolla:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## AI?

> Generated with Claude Sonnet 4.6 via [Claude Code](https://claude.ai/code), with human supervision and design decisions throughout.

## Checklist

## Description

Adds `IterationSpeedColumn`, a progress column that displays iteration speed
(e.g. `1.2 it/s`). For slow tasks (speed < 1 it/s), it displays inverted
time-per-iteration (e.g. `2.0 s/it`), matching tqdm's behaviour. An optional
`unit` parameter (default `"it"`) allows custom labels such as `"sample/s"` or
`"epoch/s"`.

This feature was already approved and merged by @willmcgugan in #3332, then
reverted solely due to dynamic `setattr` on the `Task` dataclass
(`task.last_speed = ...`), which breaks static type checkers. This
implementation avoids that entirely — `Task` is never mutated. The column reads
`task.finished_speed or task.speed`, both existing fields already tracked by
the framework.

`TaskProgressColumn.render_speed` is also enhanced (backward compatibly) with
the same slow-task inversion and a `style` parameter, so unbounded tasks using
`show_speed=True` benefit too.

**Related:**
- https://github.com/Textualize/rich/discussions/926 — original feature request
- https://github.com/Textualize/rich/pull/3332 — first implementation, merged then reverted
- https://github.com/Textualize/rich/pull/3533 — separate open attempt at the same feature

### DIFF
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -752,7 +752,7 @@ class TaskProgressColumn(TextColumn):
         """
         if speed is None:
             return Text("", style=style)
-        if speed < 1:
+        if 0 < speed < 1:
             return Text(f"{1 / speed:.1f} s/{unit}", style=style)
         scale, suffix = filesize.pick_unit_and_suffix(
             int(speed),
--- a/rich/progress.py
+++ b/rich/progress.py
@@ -734,24 +734,33 @@ class TaskProgressColumn(TextColumn):
         )
 
     @classmethod
-    def render_speed(cls, speed: Optional[float]) -> Text:
-        """Render the speed in iterations per second.
+    def render_speed(
+        cls,
+        speed: Optional[float],
+        style: StyleType = "progress.percentage",
+        unit: str = "it",
+    ) -> Text:
+        """Render the speed in iterations per second, or seconds per iteration for slow tasks.
 
         Args:
-            task (Task): A Task object.
+            speed (Optional[float]): Speed in iterations per second, or None if unknown.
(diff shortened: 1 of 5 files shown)


rich/progress.py
```diff
@@ -734,24 +734,33 @@ def __init__(
...
+            unit (str, optional): Unit label for iterations. Defaults to "it".
 
         Returns:
             Text: Text object containing the task speed.
         """
         if speed is None:
-            return Text("", style="progress.percentage")
-        unit, suffix = filesize.pick_unit_and_suffix(
+            return Text("", style=style)
+        if speed < 1:
```
Sanjays2402: `speed < 1` includes 0.0, and that is reachable today: `Progress.advance(task, 0)` appends a sample unconditionally (unlike `update`, which only appends when `update_completed > 0`), so `task.speed` comes back as exactly 0.0, not None. `finished_speed or task.speed` keeps it 0.0, so the existing TaskProgressColumn path that used to render "0.0 it/s" now raises ZeroDivisionError on `1 / speed`. `if 0 < speed < 1` would keep the inversion and leave zero on the old branch.

PierreLapolla: Good catch thanks! I was indeed able to reproduce the ZeroDivisionError. Fixed it with your suggested `if 0 < speed < 1` and added a test to cover this behavior.

